Issue #27, opened 2021-06-15
Adding in Hyperspy/Pyxem Support


CSSFrancis:
Some of the developers over at pyxem were talking about adding in support for dynamic type simulations/ 4-D STEM https://github.com/pyxem/diffsims/issues/170.   abTEM came up as a good alternative which we could reference and work alongside rather than trying to support dynamic type simulations.

I was wondering if you would be interested in adding support for [hyperspy](https://github.com/hyperspy/hyperspy)/[pyxem](https://github.com/pyxem) to the package?

There are a couple of ways that this could be accomplished (non-exclusively):

- Adding support for the hyperspy file format as a parameter to the write function https://github.com/jacobjma/abTEM/blob/01e726c9accc7ce25c13e07e120ec45a0e65c645/abtem/measure.py#L696 
- Adding a `to_hyperspy` to the `Measurement` class which converts some `Measurement` object to some `BaseSignal` class.  (This would require a pyxem/hyperspy dependency which might be more of a consideration than just writing the files) 

I think that this would be an easy way to increase the visualization/ analysis tools available some simulation without too much added overhead. I would be willing to create a small pull request to make these changes if you are interested in adding in this support.

@din14970 Based on the conversation we had earlier you might have some additional thought.

jacobjma:
Yes, we would be happy to add support for hyperspy. 

I think adding a `to_hyperspy` method is the better option, however, hyperspy should only be imported inside that method, so it is an optional dependency. Adding additional options for export would also be welcome. 

Please do make a pull request.

jacobjma:
One gotcha is that calibrations in abTEM are allowed to be `None` (I should maybe change that). This might help:
```
calibrations = measurement.calibrations 
array = measurement.array

hyperspy_calibrations = []
for calibration, n in zip(calibrations, array.shape):
    if calibration is None:
        hyperspy_calibrations.append({'size':n, 'scale': calibration.sampling, 'units':calibration.units, 'name':calibration.name, 'offset':calibration.offset})
    else:
        hyperspy_calibrations.append({'size':n})
```

din14970:
Anything that can be considered a "measurement" would translate nicely to Hyperspy: projected potentials, simulated STEM images, simulated 4D-STEM data, simulated DP's. In this case you could also consider shaving off parts of code that are not core business of image simulation and leave that up to the analysis packages: making line profiles, integrating within a disk, ... just an idea.

[...]

PR #29, opened 2021-06-17
Adding Hyperspy support

CSSFrancis:
This pull request helps to add in support for Hyperspy as discussed by #27.

This allows for the creation of a `hyperspy.BaseSignal` object using the `to_hyperspy` function. 

For a basic example I have simulated two icosahedral clusters and then created a 4-D STEM simulation. 

image

```python 
import ase
from ase.io.cif import read_cif
from abtem import *
from ase.cell import Cell
from abtem.structures import orthogonalize_cell
from ase.cluster.icosahedron import Icosahedron

cluster = Icosahedron("Pd", noshells=3)
cluster2 = Icosahedron("Pd", noshells=3)
cluster2.translate([9,9,9])# moving one of the clusters
cluster.extend(cluster2) # Combining the two clusters
cluster.center(vacuum=20)
cluster = orthogonalize_cell(cluster)
fig, ax1 = plt.subplots(1, 1, figsize=(12,4))


potential = Potential(cluster, 
                      gpts=1024,
                      projection='finite', 
                      slice_thickness=1, 
                      parametrization='kirkland').build(pbar=False)

probe = Probe(energy=200e3,
              semiangle_cutoff=5,
              rolloff=0, defocus=0, Cs=0, focal_spread=0)
gridscan = GridScan(start=[10, 10], end=[50,50], sampling=4)
show_atoms(cluster, ax=ax1, title='Top view')
gridscan.add_to_mpl_plot(ax1)
detector = PixelatedDetector(max_angle=25) # About 10 nm^-1
measurement = probe.scan(gridscan, detector, potential)
signal = measurement.to_hyperspy("diffraction")

print(signal.axes_manager)

signal.plot(vmax= 0.005)
````

image

image

image

### DIFF
--- a/abtem/measure.py
+++ b/abtem/measure.py
@@ -700,17 +700,29 @@ class Measurement(AbstractMeasurement):
         signal_type: str
             The signal type alias for some signal type
         """
-        from hyperspy.singals import BaseSignal, Signal1D, Signal2D
-        if self.calibrations is None:
-            sig = BaseSignal(self.array)
+        from hyperspy._signals.signal2d import Signal2D
+        from hyperspy._signals.signal1d import Signal1D
+        signal_shape = np.shape(self.array)
+        axes = []
+        for i, size in zip(self.calibrations, signal_shape):
+            if i is None:
+                axes.append({"offset": 0,
+                             "scale": 1,
+                             "units": "",
+                             "name": "",
+                             "size": size})
+            else:
+                axes.append({"offset": i.offset,
+                            "scale": i.sampling,
+                            "units": i.units,
+                            "name": i.name,
+                            "size": size})
+            print(axes)
+        if len(signal_shape) == 3:
+            # This could change depending on the type of measurement
+            sig = Signal1D(self.array, axes=axes)
         else:
-            signal_shape = np.shape(self.array)
-            axes = [{"offset": i.offset,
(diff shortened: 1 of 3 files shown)


CSSFrancis:
@jacobjma 
For the most part I think that this is somewhat ready to go.  The only thing that I don't know is properly handled is which Signal type is created.  (`Signal1D` or `Signal2D`).  Right now the signal type is just inferred from the number of dimensions, but it could be made more explicit

I would suggest saving the parameters for the `Probe` object with every `Measurement` object (maybe it is and I just didn't look close enough).  For saving in `Hyperspy` we can save all of this information as metadata just to keep track of it.

CSSFrancis:
@din14970

jacobjma:
> For the most part I think that this is somewhat ready to go. The only thing that I don't know is properly handled is which Signal type is created. (Signal1D or Signal2D). Right now the signal type is just inferred from the number of dimensions, but it could be made more explicit

The current `Measurement` object is essentially completely generic and does not really contain enough information to correctly infer the hyperspy signal type. This will require some work which is underway with other changes in the next version in the `abtem-dask` branch. 

I will merge this, and we will have to revisit once those changes are implemented.

I will also look at saving metadata with measurements.

PR #45, opened 2021-07-16
Adding a notebook example for STEM probe propagation simulations

GeriTopore:
Hello,

Recently I have been working on producing some simulations of STEM probe propagations through crystalline specimen. To make sure that my simulation results make sense, I also tried to reproduce the images and plots presented in this paper [[1]](https://www.sciencedirect.com/science/article/abs/pii/S0304399103000925?via%3Dihub) by Voyles at al, shown below:

Untitled-1

I thought this would be a nice example to showcase the capabilites of abTEM so I wrote it neatly in a notebook and added it in the "examples/interactive" folder alongside the .cif file that contains the silicon unit cell along <110>.

The notebook uses the newly implemented `to_hyperspy()` method of the `Measurement` class to visualise the simulation results, although with the way that the method is currently implemented, the results array (which is essentially an image stack of the exit wave intensity after each multislice step) is recognised as a `Signal1D` object rather than `Signal2D`, so I had to switch between them using the hyperspy `as_signal2D()` method.

Let me know if this would be a nice addition to the documentation.

P.S: I am pretty new to the whole GitHub and open-source community environment, so any feedback on the code or the contribution in general is hugely appreciated. 

[[1] Voyles PM, Grazul JL, Muller DA. Imaging individual atoms inside crystals with ADF-STEM. Ultramicroscopy. 2003 Sep;96(3-4):251-73. doi: 10.1016/S0304-3991(03)00092-5. PMID: 12871793.](https://www.sciencedirect.com/science/article/abs/pii/S0304399103000925?via%3Dihub)


TomaSusi:
Thank you! This seems like an excellent addition, we'll review it in detail once Jacob is back from his summer holidays.

jacobjma:
Hi,

Really cool contribution, thanks! It works great.

The issue regarding signal type should be fixed with the upcoming version using dask. So you may like to revisit once that version is merged.

PR #46, opened 2021-07-19
Add to_hyperspy() method to potentials

GeriTopore:
This pull request helps to add in the ability to visualise a `PotentialArray` object as a Hyperspy `Signal2D`. Somewhat inspired by one of the previous pull requests https://github.com/jacobjma/abTEM/pull/29, I thought it would be useful to have the ability to visualise the potential as soon as it is built. 

Here is an example using a SrTiO3 crystal to make a potential and visualising it as a `Signal2D` object, using the navigation axis to move along the potential thickness:

```python
import hyperspy.api as hs
from ase.io import read
from abtem.potentials import Potential
from abtem import show_atoms

atoms = read('srtio3_100.cif')
atoms *= (2, 2, 10)

fig1, (ax1, ax2) = plt.subplots(1,2, figsize=(14,6))
show_atoms(atoms, title='Top View', plane='xy', ax=ax1)
show_atoms(atoms, title='Side View', plane='xz', ax=ax2);
```
srtio3

```python
potential = Potential(atoms, gpts=256, parametrization='kirkland').build() #PotentialArray object
potential_as_hspy = potential.to_hyperspy()
potential_as_hspy.plot()
potential_as_hspy.write("potential.hspy", format="hspy") # Save in .hspy format
```
nav
sig

I also added in a test case in `test/test_potentials.py`, but because I am still pretty new to contributing to open source projects I am not sure if it is the correct way to do it, although it did pass when I ran in with `pytest`.

Let me know if this would be a useful addition, but also if the method can be improved.


TomaSusi:
Looks all reasonable to me and the test seems appropriate, we'll review when Jacob is back from holidays.

jacobjma:
Thanks. It is merged, however, we may need to redo some of it with the upcoming dask implementation.

PR #53, opened 2021-09-12
MTF Bug Fix

matthewhelmi:
Spatial_frequencies is returning 1D arrays, which was not the case in pyQSTEM. 

The MTF was effectively doing a 1D convolution on the image wave, causing an anisotropic distribution of the spectral properties of the noise.

This has been fixed by taking the output from spatial_frequencies and creating a mesh grid to do a proper 2D convolution with the image wave.

Also the mtf was being applied to the image wave as a product with sqrt(mtf) in the Fourier domain. I am not sure of the reasoning for the square root so this has been removed.

Future note: The current default_mtf does not involve the Nyquist frequency. I think this should be addressed in the future. At least there is the possibility of providing a user defined mtf at the moment.

@schiotz


schiotz:
@jacobjma Thanks for merging this so fast.  I am making an EasyBuild configuration file for easier installation on Niflheim etc ().  If you are making a new release soon-ish, I will wait for it, otherwise I can just include this fix in the EasyBuild recipe as a patch.

PR #54, opened 2021-09-13
Make FFTW multithreading optional (default: on)

schiotz:
Currentlty, FFTW is running multithreaded in up to 12 threads.  This seems very sensible for stand-alone computers (the majority of use cases, I presume), but on a supercomputer cluster it leads to over-subscription of the cpu cores and possibly to jobs being killed by the system.  I would like to make multithreading optional.

It could probably easily be controlled from within the script, but since the use of multithreading depends more on the installation than on the script I suggest that it is controlled by an environment variable, ``$ABTEM_FFTW_THREADS``.  If set by the installation, it specifies the number of threads.

I am not sure what is the optimal default behaviour when  ``$ABTEM_FFTW_THREADS``: One thread, the current 12 threads, or the number of actual cpu cores in the machine.  So I have left the default at the current behaviour, 12 threads.

### DIFF
--- a/abtem/device.py
+++ b/abtem/device.py
@@ -3,13 +3,14 @@ from typing import Callable
 import numpy as np
 import psutil
 import pyfftw
+import os
 from numpy.core.numeric import normalize_axis_tuple
 
 from abtem.cpu_kernels import abs2, complex_exponential, interpolate_radial_functions, sum_run_length_encoded
 from abtem.interpolate import interpolate_bilinear_cpu
 
 FFTW_EFFORT = 'FFTW_MEASURE'
-FFTW_THREADS = 12
+FFTW_THREADS = int(os.getenv('ABTEM_FFTW_THREADS', '12'))
 FFTW_TIMELIMIT = 600
 
 try:  # This should be the only place import cupy, to make it a non-essential dependency


TomaSusi:
Thanks! Sounds reasonable to me, but is an environmental variable really the only/best way to do this for a modern Python code..? Honestly asking :)

schiotz:
@TomaSusi It is *certainly* not the only way!  My logic behind using an environment variable is that it is installation dependent what you want to do, more than script dependent.  But it is admittedly a bit clunky.  What are good alternatives?  Suggestions are most welcome :-)

I think it can be overruled in a script:
```
import abtem.device
abtem.device.FFTW_THREADS=8
```
but I'll test if it actually works and report back.

jacobjma:
I should first say that we are working on implementing Dask as the native way of using multiple processes in abTEM, and multiprocessing will eventually be the main parallelization model, even on single machines. This version will also use MKL FFT instead of FFTW. The abtem-dask branch already works ok (it is quite a bit faster on clusters), but it still needs some work to be stable. 

I would be fine with having an _optional_ environment variable, it probably won't survive the big merge with abtem-dask.

I have actually been looking into how to create persistent configurations in python, but I don't think I found a great solution. I don't think our Windows users are very comfortable with environment variables.

schiotz:
Nice!

You should be aware that MKL may not work well on AMD cpus.  At least the Intel compilers do not.  So maybe keeping FFTW as an option would be a good idea - not that I really know what I am talking about, just what I have heard.

Regarding this patch: The environment variable is optional.  But this patch is not important for me, if you have a more long term solution then feel free to just close it.  I can just include a patch with the EasyBuild configuration setting the number of FFTW threads to 1 (or include this patch and set the environment variable to 1).

jacobjma:
Alright, then I think we will skip this patch.

Regarding MKL vs FFTW: I have not quite decided to scratch FFTW, but the reports I have got are that MKL and FFTW are quite similar in speed, also on AMD, but MKL is a little less fiddly.

PR #55, opened 2021-09-13
Bug fix: Use == for string comparison, never "is".

schiotz:
I noticed warnings about this in a notebook (and in the test suite).  Never use the ``is`` operator to compare strings, as it tests for two objects being the same object, not having the same value.  Two strings with the same value will usually be different objects, if they are created in different places.  The code may work, but then it is being saved by "String Interning" where the python byte compiler keeps a table of the most common literal strings in the code, and maps them all to the same object.  This is not reliable, and depends on the Python version and environment.

See https://stackabuse.com/guide-to-string-interning-in-python/

### DIFF
--- a/abtem/measure.py
+++ b/abtem/measure.py
@@ -737,7 +737,7 @@ class Measurement(AbstractMeasurement):
         kwargs:
             Any of the additional parameters for saving a hyperspy dataset
         """
-        if format is "hdf5":
+        if format == "hdf5":
             with h5py.File(path, mode) as f:
                 f.create_dataset('array', data=self.array)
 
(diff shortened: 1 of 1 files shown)


schiotz:
Additional comment:  ``if x is None`` and ``if x is True`` are OK, as ``None``,  ``True`` and ``False`` are singleton objects.  However, ``if x is True`` will only be true if ``x`` is the boolean value True, not if it is any other object with truth value True.   Which is why ``if x is None`` makes sense for testing optional variables, as it will reliably detect if you have passed another object even if it has truth value False.

jacobjma:
Yes, thank you. That was an error, I will just go ahead merge this.

PR #64, opened 2021-11-24
reimplementing cartesian2polar() and polar2cartesian()

jilekz:
I tried to use cartesian2polar() function for aberration coefficient conversion and I noticed that the function is implemented only up to third order and not according to my expectations, so I tried to reimplement it. 

I am not sure why former implementation was done as it was. For example, is there a reason why there should be
```
polar['C12'] = - np.sqrt(cartesian['C12a'] ** 2 + cartesian['C12b'] ** 2)
polar['phi12'] = - np.arctan2(cartesian['C12b'], cartesian['C12a']) / 2.
polar['C21'] = np.sqrt(cartesian['C21a'] ** 2 + cartesian['C21b'] ** 2)
polar['phi21'] = np.arctan2(cartesian['C21a'], cartesian['C21b'])
```
rather than
```
polar['C12'] =  np.sqrt(cartesian['C12a'] ** 2 + cartesian['C12b'] ** 2) `
polar['phi12'] =  np.arctan2(cartesian['C12b'], cartesian['C12a']) / 2.
polar['C21'] = np.sqrt(cartesian['C21a'] ** 2 + cartesian['C21b'] ** 2)
polar['phi21'] = np.arctan2(cartesian['C21b'], cartesian['C21a'])
```
Note the interchanged coefficient C21b with coefficient C21a and removed minus signs.

### DIFF
--- a/abtem/transfer.py
+++ b/abtem/transfer.py
@@ -555,22 +555,24 @@ def polar2cartesian(polar):
     """
 
     polar = defaultdict(lambda: 0, polar)
-
+    
+    max_order = 5
     cartesian = dict()
-    cartesian['C10'] = polar['C10']
-    cartesian['C12a'] = - polar['C12'] * np.cos(2 * polar['phi12'])
-    cartesian['C12b'] = polar['C12'] * np.sin(2 * polar['phi12'])
-    cartesian['C21a'] = polar['C21'] * np.sin(polar['phi21'])
-    cartesian['C21b'] = polar['C21'] * np.cos(polar['phi21'])
-    cartesian['C23a'] = - polar['C23'] * np.sin(3 * polar['phi23'])
-    cartesian['C23b'] = polar['C23'] * np.cos(3 * polar['phi23'])
-    cartesian['C30'] = polar['C30']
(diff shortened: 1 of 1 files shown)


jacobjma:
Hi, I am really sorry that it took so long to look at this. Thank you for the contribution, I will merge this.

I think the conversion is to some extent a question of definition. Unfortunately, there exist several different definitions of the CTF parameters (see e.g. p. 26 of Kirkland's "Advanced computing in electron microscopy"). The current implementation of the conversion functions follow a particular convention, however, the proposed change is the most logical way to do the conversion.

It is still a problem that users might expect this function to convert their particular convention to proper abTEM parameters. I am still not sure what to do about that, other than provide documentation of this fact.

Lastly, I note that the proposed change passes a basic sanity check:
```
ctf_dict = polar2cartesian({})

for key in ctf_dict:
    ctf_dict[key] = np.random.rand()

for v1, v2 in zip(ctf_dict.values(), polar2cartesian(cartesian2polar(ctf_dict)).values()):
    assert np.isclose(v1, v2)
```

PR #67, opened 2022-01-18
Import from collections.abc rather than collections

jan-janssen:
Required for https://github.com/conda-forge/abtem-feedstock/pull/9


jacobjma:
Thank you!

PR #70, opened 2022-02-03
Fix using correct python version in github CI

ericpre:
- the python version used in the test matrix was incorrect: only the current version was used
- add python 3.10 to the test matrix


jacobjma:
Thank you!

PR #75, opened 2022-02-24
Improve `to_hyperspy` methods

ericpre:
Improve `to_hyperspy` methods:
- use a more generic approach
- fix order axes, etc.
- add `to_hyperspy` method to `PotentialArray` class


jacobjma:
Thanks. FYI I also made some small changes to the `to_hyperspy` methods in a local commit that will be merged with this.

PR #79, opened 2022-04-18
Fixed extent in 3D wavefunction example

Quantumstud:
Hello,

If I am not wrong, in line 7 of the example of 3d_wavefunction.ipnyb, the extent in the horizontal direction should correspond to the x-direction of the wave3d.extent. Hence the dimension index should be wave3d.extent[0].

Thank you.

Regards,

Tara Mishra


jacobjma:
You are right, and I am very happy to have PRs like this. Thank you.

PR #87, opened 2022-08-25
probe modification

smribet:
This PR has some examples of probe modification with `AbstractAperture` and functions to return the complex probe in real and reciprocal space. Was this what you had in mind for the beam modification?  If this framework makes sense to you, I will keep adding to this module.

```
import abtem
import numpy as np

from abtem.waves.aperture import Spokes
from abtem.waves.aperture import Vortex
from abtem.waves.aperture import Bullseye

probe = abtem.Probe(
    sampling = 0.05, 
    extent = 30, 
    energy = 300e3, 
    aperture = Vortex(1,20),
    #aperture = Spokes (6, np.pi/12, 30),
    #aperture = Bullseye (8, np.pi/10, 4, 4, 30),
    defocus = 100,
)


fig, (ax1, ax2) = plt.subplots(1, 2, figsize=(12,6))

domain_coloring_kwargs_0 = {
            "abs_scaling": lambda x: x/(1.5*np.max(x)),
            "saturation_adjustment":3}

probe.build().as_complex_image().show(
    domain_coloring_kwargs = domain_coloring_kwargs_0, 
    cbar = True, 
    ax = ax1);

domain_coloring_kwargs_1 = {
            "abs_scaling": lambda x: x**0.1,
            "saturation_adjustment":4}

probe.build().as_complex_diffraction().show(
    ax=ax2, 
    domain_coloring_kwargs = domain_coloring_kwargs_1, 
    cbar = True);
```

probe_mod_example


jacobjma:
Yes, this is exactly what I have in mind, please keep adding to this module.

You might also want to create an example notebook for the gallery with some explanations and references. All examples will be added to the test suite, ensuring that the notebook is recreated (including images), this will provide some insurance that changes to abTEM won't break your code. 

Btw, clearly the complex plotting needs some work, I just started using the `cplot` package, but it seems we need to work on the color bars.

PR #92, opened 2022-09-30
Ptychographic Iterative Engine Generalizations

gvarnavi:
Hey @jacobjma and @TomaSusi!

I was playing around with ptychography this week, and found the `abtem.reconstruct.epie` function to be very straight-forward to generalize to common use-case extensions to the PIE framework. Figured I'd share, in case these are useful to others.

I've added an [example notebook](https://github.com/gvarnavi/abTEM/blob/62f70696e7e0abfa033539abaad5d577755b732a/examples/reconstruction/multislice_mixed-state_simultaneous_epie.ipynb) with a walkthrough of most of the functionality, but essentially the PR replaces `abtem.reconstruct.epie` with a new top-level function `abtem.reconstruct.ptychographic_reconstruction` that does the following:

- Backwards compatible* with e-PIE when called with default arguments:
```python
ptychographic_reconstruction(measurement, probe_guess)
```
I've modified the two example notebooks under `examples/reconstruction` to reflect this.
- Simultaneous electrostatic and magnetic-vector potential reconstruction when called with two sets of diffraction patterns (e.g. flipping the sample to reverse the magnetic vector potential contribution sign):
```python
ptychographic_reconstruction((measurement_forward,measurement_reverse), probe_guess)
```
- Mixed-state PIE reconstruction when called with `num_probes` (or less commonly `num_objects`):
```python
ptychographic_reconstruction(measurement, probe_guess, num_probes = 3)
```
_Note:_ The current implementation performs a rank `num_probes` SVD decomposition on `probe_guess`, but equivalently user could pass in a tuple of probe arrays.

- Multi-slice PIE reconstruction when called with `num_slices` and `slice_thicknesses`:
```python
ptychographic_reconstruction(measurement, probe_guess, num_slices = 3, slice_thicknesses = t/3)
```
- Allows user to define their own PIE functions and pass them in as a dictionary, e.g.: 
``` python
ptychographic_reconstruction(measurement, probe_guess, 
    functions_dictionary={
        'exit_wave_func': _epie_exit_wave_func,
        'amplitude_modification_func': _epie_amplitude_modification_func,
        'update_func': _pie_update_func,
        'update_center_of_mass_func': _update_single_probe_center_of_mass_func
})
```
-  Allows user to restart reconstruction, by passing in `probe_guess` and  `object_guess` arrays:
```python
ptychographic_reconstruction(measurement, probe_output.array, object_guesses=object_output.array, energy=probe_guess.energy)
```
_Note:_ This requires the user to also specify the probe energy.

Looking forward to hearing what you think!

Cheers,
George

\*Almost, I actually added one small modification to the default call - using the regularized e-PIE algorithm, but added an additional parameter to fully-reproduce the previous behavior if desired.


TomaSusi:
This looks absolutely amazing, @gvarnavi! we’re a little swamped at the moment finishing up our huge dask branch update, but we will definitely get back to you and make sure this will continue to play nicely once we merge.

gvarnavi:
Sounds good @TomaSusi - let me know how I can help with making this compatible with the dask branch once it's merged!
I also just added some simple probe-position correction for the e-PIE algorithm to test against experimental ptychography data.

gvarnavi:
Hey @jacobjma @TomaSusi!

I just pushed some new commits which completely restructured the implementation I mentioned above.   

This was ultimately motivated to support reconstruction of experimental datasets which typically have a fixed angular sampling and maximum scattering angle. In order to support this, I had to switch to a padded object, which gets tiled by a smaller (typically square) region-of-interest probe. 

As the number of features started increasing, it become unwieldy to keep track of the logic in a single top-level function - so instead I opted for an Abstract Base Class type structure, which hopefully is cleaner. 

The current implementation defines an `AbstractPtychographicOperator` class which defines some common static methods as well as common properties, and requires `preprocess()`, `reconstruct()`, and `_prepare_measurement_outputs()` abstract methods from its subclasses:
1. `RegularizedPtychographicOperator`
2. `MixedStatePtychographicOperator`
3. `MultislicePtychographicOperator`
4. `SimultaneousPtychographicOperator`

Each subclass additionally defines (at-least) 4 static methods:
1. `_overlap_projection`
2. `_fourier_projection`
3. `_update_function`
4. `_position_correction`

_Note:_ The use of static methods is to allow the user the option to mix-and-match and even pass their own functions if desired (although this might have a limited use-case).

Finally, I've added a [tutorial notebook](https://github.com/gvarnavi/abTEM/blob/7a99e239b7d21a7c0309bf8fa172f2022c982155/examples/reconstruction/ptychographic_iterative_engine_tutorial.ipynb) walking through the usage of all these new capabilities, as-well as one experimental reconstruction.

Hope the restructuring will play nicely with the dask branch, and looking forward to hearing your thoughts on this!

jacobjma:
Hi @gvarnavi 

I finally tried it; this is amazing! If you are happy with it, I think it can be merged as is.

Very clean implementation. A couple of things you could do to make merging it into the Dask branch easier. We now generally use the [black](https://black.readthedocs.io/en/stable/) code style for consistency. We use the [NumPy docstring](https://numpydoc.readthedocs.io/en/latest/format.html) format; this would also allow us to make an autogenerated online [API reference](https://abtem.github.io/doc/reference/api/api.html) for the module. 

The example is great; I think it could be added to the [advanced tutorials section](https://abtem.github.io/doc/user_guide/tutorials/tutorials.html) in the new docs. Although, in that case, it should probably be shortened a bit, the more specialized parts, such as magnetic reconstructions, could be standalone examples.

Would it make sense to, at some point, parallelize it using Dask? Multislice ePIE is pretty heavy without a GPU. It seems that it could be done using an overlapping subdivision of the dataset (see, for example, https://doi.org/10.1364/OE.22.032082). This could be implemented using [`dask.array.map_overlap`](https://docs.dask.org/en/stable/array-overlap.html), with some communication between workers after each iteration.

Regarding the experimental example; are parameters such as `rotation_angle` calibrated using some other code or is it a rough estimate?

gvarnavi:
Hey @jacobjma,

Glad to hear you think the implementation is clean! I just went through and added Numpy-like docstrings for all functions as-well as formatted code using black, so happy for this to be merged as is when you see fit, yes.

>The example is great; I think it could be added to the [advanced tutorials section](https://abtem.github.io/doc/user_guide/tutorials/tutorials.html) in the new docs. Although, in that case, it should probably be shortened a bit, the more specialized parts, such as magnetic reconstructions, could be standalone examples.

Agreed on the example being too long (and perhaps ordered somewhat confusingly). Went ahead and re-ordered this in a series of notebooks, to demonstrate one algorithm at a time (r-PIE, sim-PIE, ms-PIE, mix-PIE), all on the STO [110] slab which is presumably a more realistic example. There's then one notebook using the weak-phase object image to demonstrate the main concepts/dimensions common to all algorithms, and finally a short notebook on an experimental dataset. Let me know if you think this is now too granular?

> Regarding the experimental example; are parameters such as rotation_angle calibrated using some other code or is it a rough estimate?

The parameters are common experimental microscope parameters related to the detector geometry. E.g. `rotation_angle` refers to the relative rotation between the CBED patterns and the grid-scan due to the helical path electrons take off the optic axis. Similarly, `counts_scaling_function` is a detector-specific value to convert the pixelated detector's output to actual electron counts.

>Would it make sense to, at some point, parallelize it using Dask? Multislice ePIE is pretty heavy without a GPU. It seems that it could be done using an overlapping subdivision of the dataset (see, for example, https://doi.org/10.1364/OE.22.032082). This could be implemented using [dask.array.map_overlap](https://docs.dask.org/en/stable/array-overlap.html), with some communication between workers after each iteration.

This would be great, yes! Happy to have a first pass at this (likely the earliest i could look at this would be in a couple of weeks though). I must admit, I don't have much experience with dask however, so happy to take suggestions in the right direction! E.g. which of the new branch modules do you think is the most straightforward example of dask usage I could look at? Or even boiler-plate code if you have any.

jacobjma:
Sorry about my sluggish responses, most of my coding time is currently tied to finally getting `abtem-dask` out the door.

> I just went through and added Numpy-like does wells for all functions as-well.

This is great! I just merged it. I just tested it, but I will go through it in more detail later and possibly give some detailed suggestions for improvement. I think, at this point, the ePIE module is largely your creation. 

> There's then one notebook using the weak-phase object image to demonstrate the main concepts/dimensions common to all algorithms, and finally a short notebook on an experimental dataset. Let me know if you think this is now too granular?

I don't think it is too granular, but once we merge with `abtem-dask`, we should put in some work to make an introductory example look great online. It should also reference the more specific examples, which could be in the online gallery (and thus would need a thumbnail image similar to [these examples](https://abtem.github.io/doc/getting_started/basics.html)). 

> The parameters are common experimental microscope parameters related to the detector geometry.

How do you automatically (or systematically) calculate those parameters, or just actually get them?

> This would be great, yes! Happy to have a first pass at this (likely the earliest i could look at this would be in a couple of weeks though). I must admit, I don't have much experience with dask however, so happy to take suggestions in the right direction! E.g. which of the new branch modules do you think is the most straightforward example of dask usage I could look at? Or even boiler-plate code if you have any.

Great, I am getting comfortable with Dask, so I can be helpful with any questions. I am not really sure that there is a great example of implementing this inside abTEM. 

The best example is probably how the [`gaussian_source_size` method](https://github.com/abTEM/abTEM/blob/19829ff53fd03085534786ebdbf7c6eec7ae9dd3/abtem/measurements.py#L2082) is implemented, this takes a 4D-STEM dataset and mixes the intensity of adjacent probe positions. The data can be chunked along (or equivalently parallelized across) the scan axes, but it is assumed that there are no chunks inside each diffraction pattern.  

At a very high level, the ePIE does something similar, i.e. combine the data from adjacent diffraction patterns to create something new. However, the parallel implementation will be a lot more complex, to start with there will have to be communication between chunks after each iteration.

TomaSusi:
Hi @gvarnavi, I've just gone through all the examples, awesome! I would polish these a bit when we migrate them to the dask branch; for example, repeating the same first cell each time is not ideal, rather one should explain what is done differently in each notebook.

One issue I noticed: you don't seem to include the `data_mos2_sample.mat` needed for the last example, could you add that?

PR #93, opened 2022-10-07
Clean up of the notebook update_quick_hrtem_and_saed.ipynb

schiotz:
Clean up of example notebook

* The original notebook first built a gold nanoparticle, then a graphene sheet, and proceeded with the latter.
* Frozen phonons were commented out, but subsequent cells assumed it was used.

Changes:

* Use the gold nanoparticle.
* Define a variable ``phonons`` that decides whether frozen phonons are used or not, and use it in plotting cells later in the notebook.
* Set phonon amplitude from Debye temperature.

@jacobjma You may want to veto the use of the Gold nanoparticle, then I can easily go back to graphene.


jacobjma:
Thanks, this is better. Sorry about the slow merge.

PR #95, opened 2022-12-01
PIE Notebooks cleanup and probe-normalization bug fix

gvarnavi:
This PR fixes a bug with the probe array normalization to handle finite-dose measurements.  
The notebooks have been updated to reflect this by using finite-dose measurements (On the plus side, the simultaneous reconstruction benefit to SNR is now very apparent!).   

Finally, added support for a pure-phase-object constraint in the initial reconstruction steps, which often helps with convergence and reconstruction quality (e.g. in the experimental example).


jacobjma:
Thanks, this looks great.

Issue #83, opened 2022-06-03
phase flipping and phase shift in CTF


EthanJUCB:
Hi Jacob and Toma,

I'd like to compare the CTF corrected experimental images with the abTEM simulated images which has the same phase flipped CTF applied. It seems like there is no function in the current CTF command. In addition, I wonder if it is possible to add phase shift to CTF in order to simulate the phase plate image.  Many thanks.

Best
Xi

jacobjma:
It would have been somewhat difficult to do before, however, I just added an optional phase shift argument to the `CTF`. 

I think this should suffice?
```python
ctf1 = CTF(energy=200e3, defocus=100, phase_shift=0)
ctf2 = CTF(energy=200e3, defocus=100, phase_shift=np.pi)

fig, (ax1, ax2) = plt.subplots(1,2)
profiles = ctf1.show(reciprocal_units=True, ax=ax1)
profiles = ctf2.show(reciprocal_units=True, ax=ax2)
```
![phase_shift](https://user-images.githubusercontent.com/17334286/172367688-8e21b710-68d7-47fb-a58c-cc20c4dccbf6.png)

EthanJUCB:
Hi Jacob,

The phase shift option looks great. I am going to give it a try.

Looking forward to the phase flipping option. 

Thanks very much.

jacobjma:
Hi,

I thought that phase flipping was just shifting the phase by pi, which is done above. What is phase flipping if not that?

EthanJUCB:
Hi Jacob,

Sorry for the really late reply. No, phase flipping refers to the flip of the negative part of CTF plots to positive. It would be great to have that in abTEM since it is the basic CTF correction.

TomaSusi:
Could you provide an example in the published literature, or some good web resource?On 17.02.2023, at 04:18, Xi ***@***.***> wrote:﻿
Hi Jacob,
Sorry for the really late reply. No, phase flipping refers to the flip of the negative part of CTF plots to positive. It would be great to have that in abTEM since it is the basic CTF correction.

EthanJUCB:
Hi Toma, 

Please see the paper from Downing and Gleaser, 
https://www.sciencedirect.com/science/article/pii/S0304399108000429?via%3Dihub
and this website http://www.bio.brandeis.edu/~shaikh/lab/ctf

It would be great to have the basic CTF corrections, such as phase flipping and Wiener filter, in abTEM.

TomaSusi:
[...]

PR #97, opened 2023-03-30
Flipping the CTF phase

TomaSusi:
Added the keyword flip_phase for CTF to flip the sign of the negative parts of the CTF if set to True. This seems to produce the desired effect when applied to HRTEM images (example modified from our walkthrough shown below):

download

Work in progress: please comment, and if this implementation looks ok, I will try implement the Wigner filter.


TomaSusi:
All right, I think this is ready for review, and possibly merging if the discussion in the issue is resolved (https://github.com/abTEM/abTEM/issues/83).

TomaSusi:
I think this can be merged now @jacobjma.

PR #103, opened 2023-05-23
Fix numpy error

CSSFrancis:
This fixes the error with numpy> 1.20.0 by changing np.int --> int and np.float to float and xp.int to xp.int64.


CSSFrancis:
Failures due to conda env no building due to out of memory error.  That's a new one for me. @jacobjma do you want to try rerunning them to see if it was just a fluke?

jacobjma:
@CSSFrancis thank you! I will merge it even with the failed checks for now, and hopefully figure it out tomorrow. It only seems to happen on 3.7.

PR #106, opened 2023-06-26
fixing binder

alex-rakowski:
Binder recipe doesn't build currently. This seems to fix it, not sure if you want to point at your repository or the abTEM version.
```
git+https://github.com/abTEM/abTEM@master#egg=abTEM
```


jacobjma:
Thanks!

Issue #108, opened 2023-07-14
[abtem-dask] Type error: .multislice() using crystal_potential with seeds=None


EugenKozyrau:
Using the .multislice method with a crystal_potential that has seeds=None produces a TypeError after ~1s of computing:

This happens on the most recent abtem-dask release.

Is there a workaround?

Edit: removal of outdated comments in code

Example code:

```
import abtem
import ase

graphene_nonorth = ase.build.graphene(vacuum=2)
graphene, transform = abtem.orthogonalize_cell(
    graphene_nonorth, max_repetitions=10, return_transform=True
    )

potential = abtem.Potential(  #Lobato parametrization IAM potential
    graphene,                     
    sampling=0.05,            #spacing of the xy-samples of the potential
    parametrization="lobato",
    slice_thickness=1,        #the spacing of the slices in direction of electron beam 
    projection="finite",      #how those slices are calculated
    plane="xy"
    )

#potential.build().compute()

probe = abtem.Probe(
    sampling=0.05,
        #extent=20,
    energy=60e3,            
    semiangle_cutoff=30,  
    defocus=2,
    C10=50,
    Cs=-50e-6 * 1e10,
    )

grid_scan = abtem.GridScan(
    potential=potential, fractional=True, start=(0, 0), end=(1., 1.), sampling = 0.123,  
    endpoint=False                                                   
    )

crystal_potential = abtem.CrystalPotential(potential_unit=potential, repetitions=(5, 5, 1), seeds=None)
probe.grid.match(crystal_potential)

# using multislice with a crystal potential that has seeds=None produces a TypeError after ~1s of computing:

pixelated_detector = abtem.PixelatedDetector(max_angle=200)
cbed = probe.multislice(crystal_potential, scan=grid_scan, detectors=pixelated_detector).compute()

TomaSusi:
Why are you using `seeds=None`..?

EugenKozyrau:
That's the default anyway.

14.07.2023 12:34:13 Toma Susi ***@***.***>:

TomaSusi:
So it gives an error without specifying the keyword, or only when you do specify it (even though it should be the default)? Maybe the type hinting is incorrect then, @jacobjma could have a look.

EugenKozyrau:
Yes, it gives the same error in both cases.
I have worked around it for now, by tiling the unit cell instead of the potential. (computation takes ~8s)

EugenKozyrau:
[...]

PR #109, opened 2023-07-14
seeds=None - TypeError fix  [iam.py]

EugenKozyrau:
Fixes a TypeError that occurs when calling CrystalPotential with the default seeds parameter.

Issue link:
https://github.com/abTEM/abTEM/issues/108


TomaSusi:
The changes suggested for lines 1391 & 1392:

```
seeds = self.seeds[start:stop]
array.itemset(i, (potential_unit, self.seeds[start:stop]))
```

don't really make sense to me. Having just defined it, are you not using `seeds` as the last argument on the second line?

EugenKozyrau:
Oh you're right, i could have just removed the slicing on line 1395 instead:
```
                if self.seeds is not None:
                    seeds = self.seeds[start:stop]
                else:
                    seeds = None

                array.itemset(i, (potential_unit, self.seeds))
```

PR #110, opened 2023-07-15
Attempt 2: TypeError fix [iam.py]

EugenKozyrau:
Fixes a TypeError that occurs when calling CrystalPotential, with the default seeds parameter.

In this updated commit, it is done by simply removing the slicing in line 1395.

This is attempt 2, because my commits on the first pull request got messy. Sorry!

Issue link:
https://github.com/abTEM/abTEM/issues/108


jacobjma:
It is merged, thank you.

Issue #86, opened 2022-08-10
beam modification


smribet:
Hi Jacob, 

I've been working with Colin Ophus on 4D-STEM simulations in abTEM with a modified probe.  We spoke to @TomaSusi about our work at M&M and would be excited to contribute phase and amplitude plates to abTEM. We would be interested in adding some of the more common probes (vortex, volta, zernike, bullseye, etc.) as well as a completely tunable option. What do you think?

The way we’ve been modifying the probe so far is by changing the `CTF` before it is passed into the `Probe,` but this doesn’t seem like the best solution.  One possibility would be to expand aperture.py for a class of plates that could be added to the `Probe` similar to the `CTF`. Where would you suggest adding this feature? 

Thanks, 
Stephanie

TomaSusi:
Thanks for opening the issue! @jacobjma is on holiday this week but we'll be sure to discuss this when he's back :)

smribet:
Enjoy the vacation! Look forward to discussing in the future

jacobjma:
Hi, I love that you want to contribute to abTEM! I think you should add your contribution to the dask branch, which should make this very easy. I will give you some detailed instructions tomorrow 👍.

jacobjma:
As I said I think it is better to contribute your code to the `abtem-dask` branch, otherwise your implementation would have to be completely rewritten in the merge.

It took a little longer to reply;  as I needed to make a few changes to make this as easy as possible. Now, in the `abtem-dask` branch you can add a new aperture type object, which is completely self-contained, but fully works within *abTEM*. Hence, you don't even need to touch the source. However, your custom aperture object has to fulfill some requirements to make it work.

Your aperture object should inherit from `AbstractAperture`, which requires the argument `energy` and `semiangle_cutoff`. This ensures that all apertures can calculate "their" Nyquist sampling.

You need to expose a property for every additional argument, e.g. below we have a `num_spokes` argument which also is a property. You need to have a metadata property, which will add metadata to the wave function, this can just be left as an empty dictionary. 

Finally, you need to implement the main method `evaluate_with_alpha_and_phi` which; based on a 2d grid of radial angles (`alpha`) and azimuthal angles(`phi`) (both in radians), will calculate the array which will be multiplied with the Fourier space wave function.

This sounds a little complicated, but I think the example below shows that it is pretty easy. We create a new aperture object called `MyAperture`, with an unusual shape as shown in the image below.

```
from typing import Union

import numpy as np

[...]

PR #114, opened 2023-08-15
Zernike phase plate

smribet:
This PR has the Zernike phase plate we had discussed awhile ago in issue #86.

[zernike_example.ipynb.zip](https://github.com/abTEM/abTEM/files/12346552/zernike_example.ipynb.zip)

### DIFF
--- a/abtem/transfer.py
+++ b/abtem/transfer.py
@@ -525,6 +525,7 @@ class Vortex(BaseAperture):
 
         return array
 
+
 class Zernike(BaseAperture):
     """
     Zernike aperture.
@@ -572,8 +573,8 @@ class Zernike(BaseAperture):
     def center_hole_cutoff(self) -> float:
         """Cutoff semiangle of aperture hole."""
         return self._center_hole_cutoff
-    
-    @property  
+
+    @property
     def phase_shift(self) -> float:
         """Phase shift of Zernike film."""
         return self._phase_shift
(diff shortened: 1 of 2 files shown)


jacobjma:
Looks good 👍 . Just a couple of things: 

There should not be applied an ifftshift in the `Waves.diffraction_patterns` method. 

I do understand why you do that. However, by default, *ab*TEM builts the probe shifted to the center of the `extent`; hence, the disk in complex reciprocal space will have the checkered pattern. You can avoid that by building the probe at `(0,0)`:
```python
probe = abtem.Probe(...).build(scan=(0,0))
```

We use float32 and complex64; hence I would change "float" to "float32". It will (hopefully) be converted down the line, but better to be consistent. Note: I have been thinking about supporting other precisions as a user configuration, even float16 might be interesting.

We use the [Black code formatting style](https://black.readthedocs.io/en/stable/). This just helps with keeping everything consistent, hence when someone inevitably applies autoformatting to the file, they do not get a diff from your contribution. This is not too important though.

A future improvement could be to enable multiple values for each parameter. This will enable easier parallelization over that parameter, but in this case, the interactive plotting enabled is probably more interesting.

smribet:
Hi Jacob, 

Thanks for all the feedback! It's great to learn more about abTEM and all the new features. Does this look better to you? Please let me know if there are any further changes to make. 

~Stephanie

jacobjma:
This is perfect. Thank you.

PR #209, opened 2025-05-15
Attempt at documenting SMatrix traige

argerlt:
Okay, I would NOT consider this something worth merging any time soon, but here is some documentation of the last 3 days of work I spent trying to figure out the SMatrix errors.

I made a file in test called `simplest_example_of_SMatrix_error` that reproduces the main error I've been seeing, and shows a change that partially addresses it.

There is also a commented out band-aid solution in `SMatrix.reduce` that seems to fix some problems while causing others. I don't think it's the right solution at all, but it helps highlight the problem with the unaccounted for axes.

Hope this helps someone, but I think anything beyond this is above my understanding.

### DIFF
--- a/abtem/prism/s_matrix.py
+++ b/abtem/prism/s_matrix.py
@@ -6,6 +6,7 @@ import inspect
 import operator
 import warnings
 from abc import abstractmethod
+from copy import deepcopy
 from functools import partial, reduce
 
 import dask.array as da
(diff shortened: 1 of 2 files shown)


argerlt:
@jacobjma ,@TomaSusi

TomaSusi:
Thanks a million, we'll try to get these fixed as soon as we can. In the meanwhile, the testsuite seems to pass apart from smatrix related tests.

TomaSusi:
The issues you encountered should now be fixed in https://github.com/abTEM/abTEM/tree/dev.

Issue #180, opened 2024-11-07
Request for Side View Visualization of Beam Intensity Depth Profile


Nick-tyh2110:
Hi,

I'm currently running STEM simulations using abTEM and I'm wondering if it's possible to generate a side-view visualization of the beam intensity depth profile for a specific probe location.

As shown in the attached image, which originates from the research paper ["Subatomic Channeling and Helicon-Type Beams in SrTiO3"] (https://journals.aps.org/prl/abstract/10.1103/PhysRevLett.122.075501), this visualization would provide a clearer understanding of the beam's interaction with the sample at that particular point.
![Screenshot 2024-11-07 163941](https://github.com/user-attachments/assets/d7b26998-c7c5-495b-a56c-740af0e5e7e7)

Please let me know if this functionality is currently available or if it's something that could be implemented in future versions of abTEM.

Thank you for your time and consideration.

TomaSusi:
Hi Nick! This isn't directly implemented, but we had an example that reproduced the plot in the legacy version of abTEM:
https://github.com/abTEM/abTEM-legacy/blob/master/examples/interactive/probe_propagation_Si.ipynb

It shouldn't be too hard to adapt this to the current version (changing some function and keyword names; not sure Hyperspy is needed), and we could put an example up if you're able to do that.

Nick-tyh2110:
@TomaSusi Thank you so much for the assistance!

Nick-tyh2110:
Currently I am working on adapting it to current version. 
May I know where can i find the following in V1.0.5 of abTEM?
`abtem.measure import Measurement, Calibration` of legacy abTEM

TomaSusi:
In the new version, we use a so-called flat namespace, so you can import everything directly from abtem, ie. `from abtem import Measurement`.

Nick-tyh2110:
I am stucked and i would need some help from you guys. 

Since I couldn't get the 'Si_110.cif' in the example, i willl just replicate it with the stem quickstart structure.

Here is what i have done so far:

# Functions:
def run_multislice(potential, probe):
    
    probe_copy = copy.deepcopy(probe) #creates a deep copy of probe
    n_slices = potential.num_slices #Number of slices in the potential
    gpts = potential.gpts #Grid points in each slice, a 2D array
    waves = np.empty((n_slices, gpts[0], gpts[1]), dtype='float64') # An empty 3D array to store the intensity of the probe after interacting with each slice.

    propagator = FresnelPropagator()

[...]

PR #210, opened 2025-05-16
3d plots for atoms and potentials

argerlt:
I was repeatedly running into some issues with `ase.build.surface` and `abtem.orthogonalize_cell` when trying to look down unusual axes (GaAs along the 14,2,1 was where the problem started for me, but that's ancillary to the topic of this PR)

Ultimately, the problem was a user error, but diagnosing it was devilishly difficult with the default 2d projection plots. Additionally, there were some missing features in `abtem.show_atoms`, such as the inability to set `plane` to off-plane viewing directions.

Since I needed 2D and 3D plots for my own work, I tried my hand at a generic `abtem version`.  some examples can be seen below:
```
ruby = ase.io.read("ruby.cif")* [9,11,4]
abtem.show_atoms(ruby, merge=False, plane = "xy")
abtem.show_atoms(ruby, merge=False, plane = (10,15))
abtem.show_atoms3d(ruby)
```

image

This PR definitely still needs some TLC, but I was far enough long it seemed worth sharing in case others have feedback.

I also want to add a 3d isosurface plot for potentials, and (maybe?) a pyvista plotting tool for both, as the graphics are considerably nicer looking and it's possible to quickly plot true ionospheres instead of scatterplot markers.

Let me know what you think, open to suggestions

### DIFF
--- a/abtem/__init__.py
+++ b/abtem/__init__.py
@@ -26,7 +26,7 @@ from abtem.potentials.iam import CrystalPotential, Potential, PotentialArray
 from abtem.prism.s_matrix import SMatrix, SMatrixArray
 from abtem.scan import CustomScan, GridScan, LineScan
 from abtem.transfer import CTF, Aperture, SpatialEnvelope, TemporalEnvelope
-from abtem.visualize.visualizations import show_atoms
+from abtem.visualize.visualizations import show_atoms, show_atoms3d
 from abtem.waves import PlaneWave, Probe, Waves
 
 __all__ = [
@@ -65,6 +65,7 @@ __all__ = [
     "TemporalEnvelope",
     "SpatialEnvelope",
     "show_atoms",
+    "show_atoms3d",
     "Waves",
     "Probe",
     "PlaneWave",
--- a/abtem/visualize/__init__.py
+++ b/abtem/visualize/__init__.py
@@ -1,4 +1,5 @@
 from abtem.visualize.visualizations import (
     Visualization,
     show_atoms,
+    show_atoms3d,
 )
--- a/abtem/visualize/visualizations.py
+++ b/abtem/visualize/visualizations.py
@@ -15,6 +15,8 @@ from matplotlib.axes import Axes
 from matplotlib.collections import PatchCollection
 from matplotlib.lines import Line2D
 from matplotlib.patches import Circle
+from mpl_toolkits.mplot3d.axes3d import Axes3D
+from scipy.spatial.transform import Rotation as R
 
 from abtem.atoms import pad_atoms, plane_to_axes
 from abtem.core import config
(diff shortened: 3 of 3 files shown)


TomaSusi:
Thank you for your efforts! I'm actually not 100% sure if this is something we should support in abTEM, since we usually don't allow arbitrary directions for the simulation... though that's no longer true for Bloch waves. Maybe you could explain more why you need this and why this should be include in abTEM?

Also, did you notice the visualization of 3D structures we already support in our advanced tutorial?
https://abtem.github.io/doc/user_guide/tutorials/advanced_atomic_models.html#visualizing-structures-in-3d

As long as the changes you propose don't alter existing behavior, the code is documented, and any tests that might be needed are added, we can certainly consider adding this if you think it is helpful for common workflows.

----

As for isosurfaces, I would typically suggest saving the potential as a `.cube` file and visualizing that in an external program (as is done in GPAW, see: https://gpaw.readthedocs.io/tutorialsexercises/wavefunctions/plotting/plot_wave_functions.html). However, maybe the fact that we have one direction that is inequivalent would require special functionality. Do you think we could do this without adding new dependencies?

Please also note a somewhat related issue on visualizing the probe intensity profile from the side, which could be relevant in this context:
https://github.com/abTEM/abTEM/issues/180

argerlt:
>Thank you for your efforts! I'm actually not 100% sure if this is something we should support in abTEM, since we usually don't allow arbitrary directions for the simulation... though that's no longer true for Bloch waves. Maybe you could explain more why you need this and why this should be include in abTEM?

Short answer, it's been helpful for both comprehension and error diagnostics.

A few longer  answers:
1) Small picture: I'm using abTEM to brute-force query which axes are best for precessed DPC-STEM (ie, what axes and with what procession angles do you see minimal variation in direct beam intensity due to factors OTHER than the defects and fields I care about).  Turns out, best axes are weird high-order zone axes (ie, not really on any axes),  and ASE occasionally makes errors when orthogonalizing them, and then again errors can occur with FrozenPhonons. Diagnosing my problems with only 2D plots restricted to three orthogonal views was rough, but 3d plots made the problem instantly clear. My solution ended up being making GaAs supercells externally and feeding them to ASE, so a fair argument is ASE and abTEM weren't "wrong", i was just using them poorly, but diagnosing all that was stilI way easier wit a tool that worked directly on abTEM objects. I figure as abTEM gets more popular, more weird niche use cases and associated errors like this might appear, and since I solved it for me, i could also just solve it generally for everyone.

2) Big picture: For me, it was a helpful learning tool as well. I understand logically that the abtem multislice is solving integrals over whole slices and that a 3d plot of atoms and potentials is possibly misleading as to how the algorithm actually works, but it helped me a lot with both diagnosing issues and understanding how the `Probe` would propagate.  For these weird essentially-off-axis plots, it also helped with determining a reasonable multislice thickness. Finally, when I tried using GPAW, I made some silly user errors which were next to impossible to see when the potential was integrated over slices, but showed up very obviously in a 3d plot.

3) You exactly hit the nail with Bloch wave as well. this is what I've been using very recently and it works great, but again, way easier to understand when you can plot a diffraction pattern next to the view it was taken from, without having to export to a different tool or use a different plotting function

> Also, did you notice the visualization of 3D structures we already support in our advanced tutorial?
> https://abtem.github.io/doc/user_guide/tutorials/advanced_atomic_models.html#visualizing-structures-in-3d

I had, but I had some frustrations with this as well. It's moving the atoms not the plot, which is an identical operation if you have your rotation conventions right, but I'm jumping between kikuchipy and abTEM which are passive/active respectively, and it was playing havoc with how I was brute-force searching the entire S2 sphere. Also, since the problem for me was "something is silently altering atoms incorrectly", any step that further altered the atoms themselves via rotation was obfuscating the problem more. Finally, from a simplicity standpoint, it was nice to just have the rotation capability in the same workhorse function used for viewing in the walkthrough. 

> As long as the changes you propose don't alter existing behavior, the code is documented, and any tests that might be needed are added, we can certainly consider adding this if you think it is helpful for common workflows.

Thanks. To that point, I think pyVista was overambitious and too easy to break. It makes REALLY nice plots, but matplotlib can make reasonable ones without any changes to the dependencies. 

> As for isosurfaces, I would typically suggest saving the potential as a .cube file and visualizing that in an external program (as is done in GPAW, see: https://gpaw.readthedocs.io/tutorialsexercises/wavefunctions/plotting/plot_wave_functions.html). 

I didn't know about this! very cool, I will have to look into it more...

> However, maybe the fact that we have one direction that is inequivalent would require special functionality. Do you think we could do this without adding new dependencies?

I think so? i can try anyway, if it doesn't work I'm fine letting this idea go. My kneejerk reaction is " why require an outside tool be used when an internal function could do it", but I might not be giving the difficulty of the plot enough credit.... Also though, I cannot use GPAW for what I specifically want, as it's being deployed onto a hundreds of systems, including several windows machines.

>Please also note a somewhat related issue on visualizing the probe intensity profile from the side, which could be relevant in this context:
>https://github.com/abTEM/abTEM/issues/180

Ooooh, that is also a neat idea.... Yeah, I guess again to your point, these are things you COULD do with other software, but if it's not too hard and doesn't break things, it would be simpler and more appealing to just do them inside abTEM.

argerlt:
OH! I totally forgot, but to add to this:

As someone new to this whole technique and related concepts, I had a hard time understanding how [orthogonalize_cell](https://abtem.github.io/doc/reference/api/_autosummary/abtem.atoms.orthogonalize_cell.html#orthogonalize-cell) was altering my data, specifically in reference to the `max_repitition` parameter.

Again, not really abTEM's fault, hindsight being 20/20 all the info was there and I just needed to read the documentation more, but the problem would have been considerably faster to diagnose with a 3d plot or an off-axis 2D projection.

Also also, the original `abtem.show_atoms` function references the ability to make off-axis 2d projection plots, but can't actually do so. one of the little fixes I should have added was this reintroduces that missing capability.

argerlt:
@TomaSusi , changed this to a draft until it's ready for consideration as a PR, but feel free to add/change whatever you want if there's something I'm doing that doesn't align with what you want with abTEM as a whole.

PR #254, opened 2026-03-16
CBED simulation using Bloch Waves

MathijsDoel:
This PR adds the functionality to generate a CBED pattern for a sample using the `BlochWaves` class, by passing in a `Probe` instance and the desired crystal thickness.

```python
bw_cbed = bloch_waves.calculate_CBED_patterns(
    probe, 
    thickness, 
    pbar=True
)
```
The function returns a stack of `DiffractionPatterns()` for each thickness value.

Attaching a notebook which compares the Bloch waves CBED result to the multislice CBED results for a sample of Si6N8 at three different energies (150 keV, 50 keV, 15 keV).

Multislice_vs_BlochWaves

[BW_CBED_showcase.ipynb](https://github.com/user-attachments/files/26026570/BW_CBED_showcase.ipynb)


TomaSusi:
@MathijsDoel: I understood from @gvarnavi that you saw some errors when testing out the new energy ensemble code for this PR, could you post them here so I can look into it and fix them?

MathijsDoel:
For a normal probe with a single scalar energy the ctf.nyquist_sampling gives the error:
`TypeError: float() argument must be a string or a real number, not 'EnergyEnsemble'`.

As well as the multislice function not working if the probe has an EnergyEnsemble energy instead of a single value:
```
probes = abtem.Probe(
    semiangle_cutoff=semiangle_cutoff,
    energy=[energy_high,energy_mid,energy_low],
    device=device
).match_grid(repeated_potential)

multislice_all = probes.multislice(repeated_potential, lazy=False)
```

This gives the following error:
`
EnergyUndefinedError: Energy is not defined
`

While the probe itself seems to work since I can do probes.show(explode=True) and see the different probes

TomaSusi:
Oh, very sorry to have missed this in my testing! These should now be fixed in https://github.com/abTEM/abTEM/pull/257, and additional tests were added to catch these in the future.

TomaSusi:
Please note that I made a new "dev" branch that PRs should direct to, with "main" in the future only used for releases merged from dev.

PR #266, opened 2026-04-22
Adding Soft Edges to the Bullseye Probe

mitis1:
Description:
This PR allows users to apply soft edges to the bullseye probe.

Key Changes:
Added parameters to specify the number of pixels allotted to edge softness and corner softness.
A setting of 0 for these parameters will revert the probe to binary (without soft edges).

This notebook demonstrates the modified bullseye probe functionality showing how soft edges affect the probe shape.
[test-bullseye-probe.ipynb](https://github.com/user-attachments/files/26953121/test-bullseye-probe.ipynb)


TomaSusi:
Thank you! A few questions:

- If `soft_edges = False`, does the code ignore if any values for `edge_softness_px` or `corner_radius_px` are given? That would seem to me to be the correct behavior in that case, though a warning could be raised.
- Conversely, if `soft_edges = True` and no values are given, the default values of 0 would effectively result in a hard aperture. This seems something to avoid, or at least throw a warning.
- Is it really the best idea to use pixels directly to define the softness? The number of pixels will depend on the sampling of the diffraction patterns, which in turn is determined by the extent of the wavefunctions/potential (https://abtem.github.io/doc/user_guide/appendix/antialiasing.html#reciprocal-space-sampling). This could lead to a situation where a user first specifies and visualizes a `BullseyeAperture`, but then later changes the potential thus altering the effective softening when the aperture is used with a potential (https://abtem.github.io/doc/user_guide/tutorials/structured_illumination.html#d-stem-with-bullseye-probe). Alternative would be to use the physical unit (1/Å) as radii, though this is perhaps a bit less counterintuitive, so perhaps simply nothing this in the docstring would be sufficient.
- Some more code comments on the various subfunctions would be helpful.

mitis1:
Thank you for the comments! I added comments to the sub functions. I removed the soft edges variable so the users can only change the edge softness and corner radius functions to set if the aperture has soft edges. I also changed their units to be mrads. Attached is the new notebook showing how the function works. 

[test-bullseye-probe-updated.ipynb](https://github.com/user-attachments/files/27253308/test-bullseye-probe-updated.ipynb)

TomaSusi:
Thanks! Looking at your notebook, the values you're inputting for the softening seem very small – is `3e-6` really consistent in the same units with the 15 mrad `semiangle_cutoff`?

mitis1:
I am so sorry I had a typo that messed up the values. I have updated the code and attached the new notebook. 

[test-bullseye-probe-updated.ipynb](https://github.com/user-attachments/files/27372057/test-bullseye-probe-updated.ipynb)

TomaSusi:
Looks to be almost ready: please rename the attribute functions on lines 498 and 503 to omit `_px`.

Have you tested the new aperture with a probe, and checked that a simple multislice simulation produces reasonable results?

For future reference: it is not good abTEM practice to manually calculate and plot arrays (especially NOT using internal methods) – the `show()` method works just fine for your BullseyeAperture. If you do want to access the array, you can use the `to_diffraction_patterns()` method; the naming is a bit odd perhaps, but as the aperture is defined in reciprocal space, it kinda makes sense.

mitis1:
I updated the function names and tried testing the bullseye function on a Si lattice. I attached the test notebook below. 

[test-bullseye-with-si-lattice.ipynb](https://github.com/user-attachments/files/27500063/test-bullseye-with-si-lattice.ipynb)

TomaSusi:
Thank you! While testing this, I actually discovered a slicing bug that affected some ASE-default structures where atoms are exactly at the cell boundary (now fixed in PR https://github.com/abTEM/abTEM/pull/273).

That caused some of the scan positions to appear different due to some atoms being dropped from the potentials. I've updated the notebook with a better workflow using CrystalPotential for the repetitions, and only scanning over the unit cell.
[test-bullseye-with-si-lattice_fixed.ipynb](https://github.com/user-attachments/files/27521479/test-bullseye-with-si-lattice_fixed.ipynb)

I've merged this now, thank you very much for contributing!

PR #274, opened 2026-05-10
Feature: automatically commensurate sampling and slicing

TomaSusi:
## Summary

- Adds `sampling='auto'` to `abtem.Potential`: automatically finds the smallest grid near 0.05 Å/pixel that is exactly commensurate with the crystal lattice, so every symmetry-equivalent atom lands on the same fractional grid index and receives an identical discretised potential value.
- Adds `slice_thickness='auto'`: automatically finds the coarsest slicing near 1 Å whose boundaries align with the crystal planes along z.
- Fixes a negative-zero floating-point bug where atoms produced at `x ≈ −ε` by cell-transform matrix arithmetic were mapped to `cell_length − ε` instead of 0, causing wrong FFT-peak placement and spurious commensurability failures.

## Technical approach

The commensurability algorithm works on **inter-atomic spacings** (GCD-based) rather than absolute positions, making it **translation-invariant**: a crystal that has been centered (`atoms.center()`) or arbitrarily shifted produces exactly the same grid as the unshifted primitive cell. For non-periodic (finite) structures such as rotated nanoparticles, the commensurability search is bypassed and the target sampling is used directly.

Non-orthogonal cells (hexagonal systems) are handled via the existing `orthogonalize_cell` path; the algorithm operates on the orthogonalised atom positions.

## Files changed

- `abtem/slicing.py` — new `commensurate_gpts()` and `commensurate_slice_thickness()` functions; negative-zero snap in `commensurate_gpts`.
- `abtem/potentials/iam.py` — `sampling='auto'` / `slice_thickness='auto'` logic in `Potential.__init__`; negative-zero snap in `_prepare_atoms`.

## Test plan

Three standalone verification scripts (100+ configurations total):

- [x] `test_commensurate.py` — 20/20: Si, Al, SrTiO3, MoS2, Au, GaN, TiO2, Fe, ZnO; [001] and [110] orientations; infinite and finite projection
- [x] `test_commensurate_extended.py` — 47/47: centering and arbitrary translation, slab geometry with vacuum (pbc=[T,T,F]), supercells 1×1/2×2/4×4, finite projection, hexagonal gpts invariance
- [x] `test_commensurate_lattices.py` — 33/33: rock-salt (MgO, NaCl), fluorite (CaF2), zinc-blende (GaAs, InP), diamond (C), tetragonal (BaTiO3, VO2), perovskite (SrVO3), orthorhombic (TiO2 brookite); parametrisations kirkland/peng/lobato all give identical grids


TomaSusi:
Pushed three follow-up fixes for issues found during review of #347 (which rebases on this branch):

1. **Root-level verification scripts weren't wired into pytest.** `test_commensurate.py`, `test_commensurate_extended.py`, and `test_commensurate_lattices.py` lived at the repo root and were never collected by pytest — in fact their presence there broke `pytest` collection from the repo root entirely. Their assertions (built-potential value equality across ~19 lattice types incl. non-orthogonal/hexagonal cells, translation/centering invariance, slab+vacuum geometry, parametrization consistency) are ported into `test/test_commensurate_potential.py` and the root scripts are removed.

2. **Trailing thin-slice + fixed-grid z-quantization bugs in `commensurate_slice_thickness`.** The z-boundary near `cell_z` wasn't merged back with the boundary near 0, so a real but small final slice could survive as its own near-degenerate slice instead of merging into the previous one (fix: mirror the existing mid-loop merge guard across the periodic wrap). Separately, unique z-planes were deduplicated by rounding to a fixed tolerance-wide grid anchored at 0, which could split two atoms much closer together than `tolerance` into different bins (or merge genuinely distinct planes landing in the same bin); fixed by switching to the sorted-neighbor-distance clustering already used in `commensurate_gpts`. Both covered by new regression tests.

3. **`AtomsEnsemble` was searching for commensurate planes in an arbitrary single MD frame.** `Potential.__init__` used `frozen_phonons.atoms`, which for `AtomsEnsemble` is just the first trajectory frame — a thermally-displaced snapshot, not a reference lattice. Each configuration in an MD-based ensemble is independent and generally non-commensurate, so `'auto'` sampling/slice_thickness now skip the commensurate search for `AtomsEnsemble` and fall back to the same plain extent/spacing logic already used for non-periodic structures.

All changes validated against the full `test_commensurate_potential.py`, `test_potentials.py`, and `test_atomic_potential.py` suites (no regressions).

TomaSusi:
Follow-up correction to the AtomsEnsemble fix above: it originally skipped commensurate grid/slice-thickness search for any AtomsEnsemble, but a single-configuration AtomsEnsemble is unambiguous -- that one frame is the configuration (this is a common pattern; e.g. test/strategies/potentials.py's md_frozen_phonons wraps even a single static structure in an AtomsEnsemble for API uniformity). Narrowed the check to only skip commensurability search when num_configs > 1, so a single-config ensemble now gets the same commensurate grid/slicing as passing the bare Atoms directly. Added a regression test asserting the two produce identical gpts/slice_thickness.

PR #283, opened 2026-06-03
Feature: radially varying annular detector sensitivity

TomaSusi:
## Summary

Real annular detectors are not uniformly sensitive between their inner and outer integration limits — they have an angle-dependent gain (a gradual edge roll-off, or a measured response curve). See e.g.:
https://www.sciencedirect.com/science/article/abs/pii/S030439910900117X?via%3Dihub

This PR adds a **`RadialSensitivity`** that weights the diffraction pattern by `w(α)` as a function of the radial scattering angle [mrad] before integration/binning, on all radial detectors.

## API

New `abtem.RadialSensitivity`, accepting **either** form:

- a **callable** `w(α)` mapping angle [mrad] → weight, e.g. `lambda a: a / 100`
- a measured **`(angles, values)`** curve, linearly interpolated onto the grid

A weight is a fraction in `[0, 1]` (1 = full sensitivity). A `sensitivity=` argument is added to `AnnularDetector`, `FlexibleAnnularDetector` and `SegmentedDetector`:

```python
import abtem, numpy as np

# HAADF with a linear edge roll-off
det = abtem.AnnularDetector(
    inner=40, outer=100,
    sensitivity=lambda a: np.clip((100 - a) / 60, 0, 1),
)

# measured response curve on a flexible detector
det = abtem.FlexibleAnnularDetector(
    step_size=1.0, inner=40, outer=120,
    sensitivity=([40, 80, 120], [1.0, 0.9, 0.7]),
)
```

## Implementation

- `sensitivity` lives on the shared `_AbstractRadialDetector` base, so all three radial detectors support it. Default `None` preserves existing behavior exactly.
- Weighting is **per-pixel**: `AnnularDetector` weights before integrating to the image; the polar detectors weight before binning, so integrating the resulting `PolarMeasurements` reproduces the sensitivity-weighted annular signal. New `_radial_weight_map` helper in `measurements.py`, threaded through `_annular_detector_mask`, `integrate_radial`, `_integrate_fourier_space`, `polar_binning` and `_radial_binning` for both eager and lazy (dask) paths.

## Validation

Sensitivities are validated as soon as the integration range is known (at construction when `outer` is set, otherwise at detect time):

- measured-curve **angles** must lie within `[inner, outer]` (no silent extrapolation);
- **weights/values** must be within `[0, 1]` — values `> 1` are unphysical and rejected; callables are sampled across the range and checked for finite, non-negative, `≤ 1` weights;
- `inner`/`outer` must be non-negative, regardless of sensitivity.

All raise descriptive `ValueError`s.

## Detector regions & visualization

- **Singular vs plural, made explicit.** `AnnularDetector.get_detector_region` (singular) returns the per-pixel detector efficiency map [%] — binary `0/100` for a uniform detector, sensitivity-weighted otherwise. `FlexibleAnnularDetector`/`SegmentedDetector` expose `get_detector_regions` (plural) — the polar bins, filled with per-bin efficiency [%] when a sensitivity is set so the profile shows. Calling the wrong one raises a guiding `NotImplementedError` (no bare `AttributeError`).
- **`show()`** renders the continuous per-pixel efficiency gradient with an auto-scaled `%` colorbar when a sensitivity is set (previously the discrete colormap collapsed an annular detector to a single flat color), and **crops the view to 1.2× the outer angle** for a tighter default instead of the full sampling-limited angle.

## Testing

- `test_radial_sensitivity` (lazy + eager): uniform sensitivity reproduces the unweighted detector bit-for-bit, matches a hand-computed `Σ w(α)·I(α)` ground truth, callable ≡ lookup form, Flexible↔Annular consistency, copy/partition round-trip, the singular/plural guards both directions, the efficiency-map labels/scaling, the `show` gradient, and all validation errors (range, negative, `>1`, non-increasing).
- `test_detector_show_crops_to_outer_angle`: the shown view is cropped to ~1.2× outer rather than the much larger sampling angle.
- Existing suites green: `test_detect.py` (15), `test_scan.py` + `test_simulate.py` (103 passed, 78 skipped).


elli-un:
Hi Toma, any new developments?

TomaSusi:
> Hi Toma, any new developments?

Yes – this should be fully functional already.

Please have a try! If you clone the repo, using ‘gh pr checkout 283’ gets these changes which you can then use with ‘pip install -e .’

elli-un:
Cool, will do. Thank you!

TomaSusi:
@elli-un how's it looking, is the functionality working as you'd hope for?

pzeiger:
Merge-order note re #378: both branches insert code in the `_annular_detector_mask` / `_polar_detector_bins` region of measurements.py, so whichever merges second gets a textual conflict (mechanical to resolve). More substantively, after #378 the `_polar_detector_bins` results are cached and shared between all callers — immutable tuples of read-only arrays — so the new `_polar_detector_bins` call this branch adds in `get_detector_region` must stay read-only on the result (it currently is: the `regions >= 0` usage builds a fresh boolean array, so no change needed, just don't add in-place edits later).

PR #290, opened 2026-06-14
Add Gram-Schmidt fallback to orthogonalize_cell for early-exit bug handling

chenguisen:
## Bug Description

`orthogonalize_cell` has an early-exit bug: when `tuple(np.diag(atoms.cell)) == tuple(box)`, the function returns the cell unchanged without any orthogonalization. This affects zone axes where the optimal integer repetition `n` preserves the diagonal (e.g., FCC Al (112), where v2x/v1x = -1.0), causing `RuntimeError: atoms must have an orthogonal cell` downstream in the multislice routine.

## Fix: Pure Gram-Schmidt Fallback

Instead of fixing the `diag==box` condition, this PR **routes the early-exit case to a pure Gram-Schmidt fallback**:

```
orthogonalize_cell:
  1. Already orthogonal? -> return
  2. diag(cell) != box? -> integer-n search + strain (standard abTEM path)
  3. diag(cell) == box? -> skip integer search (would early-exit)
  4. Cell still non-orthogonal? -> pure Gram-Schmidt (catches early-exit case)
  5. Return orthogonal cell
```

Key design decisions:
- No modification to `Potential.get_transformed_atoms()` -- the original bug condition remains visible
- Integer-n search preserved for zone axes where `diag != box` (most cases, e.g. Al (211))
- GS fallback only for zone axes where `diag == box` triggers the early exit
- The GS fallback also catches any other failure mode of the integer-n + strain path

## Test Results

| Zone Axis | Before | After | Path |
|-----------|--------|-------|------|
| Al (112) | RuntimeError | 12 atoms, off < 1e-15 | diag==box -> GS fallback |
| Al (121) | RuntimeError | 12 atoms, off < 1e-15 | diag==box -> GS fallback |
| Al (211) | 60 atoms | unchanged | diag!=box -> integer search |

All tests run on FCC Al (mp-134) with 3-layer slabs using `abtem.orthogonalize_cell()`.

## Reproducibility

All scripts, data, and analysis are at [chenguisen/ase_build](https://github.com/chenguisen/ase_build/).

### DIFF
--- a/abtem/atoms.py
+++ b/abtem/atoms.py
@@ -874,42 +874,58 @@ def orthogonalize_cell(
     if plane != "xy":
         atoms = rotate_atoms_to_plane(atoms, plane)
 
+
     if box is None:
         box = tuple(best_orthogonal_cell(atoms.cell, max_repetitions=max_repetitions))
 
     assert box is not None
 
-    if tuple(np.diag(atoms.cell)) == tuple(box):
+    if is_cell_orthogonal(atoms.cell):
         if return_transform:
             return atoms, (np.zeros(3), np.ones(3), np.zeros(3))
         else:
             return atoms
 
-    if np.any(atoms.cell.lengths() < tolerance):
-        raise RuntimeError("Cell vectors must have non-zero length.")
+    # --- Try integer-n search + strain ---
+    # This is the standard abTEM orthogonalization. It works for most zone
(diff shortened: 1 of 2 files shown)


chenguisen:
Superseded by #291 (cleaner single-commit implementation)

PR #298, opened 2026-06-23
adding exact propagator as default

gvarnavi:
This PR changes the default propagation order in `FourierMultislice` from the paraxial approximation to `order="exact"`.

The exact propagator now correctly handles spatial frequencies beyond the free-space cutoff (`k > k₀`) by treating them as evanescent waves rather than propagating modes. The implementation:

* Uses the exact free-space propagation kernel.
* Separates propagating (`k ≤ k₀`) and evanescent (`k > k₀`) components.
* Applies exponential decay to evanescent components through the imaginary part of the longitudinal wavevector.
* Uses numerically stable evaluation of `√(1-x) - 1` near `x = 0`.

This improves physical correctness for simulations containing high spatial frequencies while remaining fully backward compatible for propagating modes.

[higher-order-ms-pr.ipynb](https://github.com/user-attachments/files/29241258/higher-order-ms-pr.ipynb)


abtem/multislice.py
```diff
@@ -49,31 +50,70 @@ def _fresnel_propagator_array(
...
+    propagating = x <= 1.0
+    evanescent = x > 1.0
 
-    # Propagator corrected in Fourier-space, only valid for order=2
-    # Eq. (4) from Microscopy and Microanalysis (2020), 26, 1147-1157
-    if order == 2:
-        f = f * complex_exponential(
-            (-np.pi * thickness * wavelength**3) / 4.0 * (kx**4 + ky**4)
+    if order == "exact":
+        phase = xp.empty_like(x, dtype=xp.complex64)
```
TomaSusi: One should not manually set the dtype, but rather respect the abTEM config that controls float32/float64 (and corresponding complex64, complex128) precision.

This is how this is canonically done with a utility helper (with the `complex = True` argument relevant for this case)

```
from abtem.core.utils import get_dtype
dtype = get_dtype(complex=True)
```

TomaSusi (commented):
One dtype issue flagged.

PR #318, opened 2026-07-05
Adding new C-PRISM algorithm

cophus:
# C-PRISM: reduce every probe from the full plane-wave expansion

Adds `upsample=True` to `SMatrix`. It runs multislice on the coarse beam set at the given `interpolation`, as PRISM does, then reduces every probe from the complete plane-wave expansion of the aperture instead of the coarse one. The interpolation factor then sets only how many multislice runs the build costs. It no longer sets the aperture sampling or the real-space cropping window the probes are reconstructed on.

```python
s_matrix = abtem.SMatrix(
    potential=potential,
    energy=300e3,
    semiangle_cutoff=20,
    interpolation=4,
    upsample=True,
)

scan = abtem.GridScan.commensurate(s_matrix, gpts=128)
measurements = s_matrix.scan(scan, detectors=detectors).compute()
```

`upsample=True` is the whole interface. Everything else is derived from the specimen and the aperture. With `upsample=False` the class is unchanged, so the two run on identical inputs.

**Validation notebook (Colab, A100):** https://drive.google.com/file/d/1I_UYNLjUddg5R3tyqY88lO-tqA7sGBLJ/view?usp=drive_link

It runs multislice, PRISM at f = 1, 2, 4 and C-PRISM at f = 2, 4 on a Ge cell with one Au column: single-probe CBED comparisons at four positions, then a 128 x 128 probe 4D-STEM scan of the whole cell. Error is measured against multislice as percent of probe intensity per radial bin, downsampled onto a common grid by block summation and subtracted pixelwise before binning, since binning first lets errors of opposite sign cancel. PRISM at f = 1 measures exactly zero by that metric, which validates the comparison chain.

## How it works

1. **Build.** Multislice on the coarse beams, plus a ring of beams just outside the aperture that anchors the interpolation at the aperture edge.
2. **Reference.** Back-propagate every beam to the middle of the specimen.
3. **Remove the trivial phases.** Divide each beam by its tilt ramp and by the vacuum defocus phase to the reference depth. In vacuum this leaves exactly unity.
4. **Interpolate** onto every plane wave of the aperture, with a band-limited (trigonometric) interpolant.
5. **Compress** into modes, each one real-space image plus one vector of plane-wave amplitudes.
6. **Reduce.** Per probe, one kernel per mode by inverse FFT, then multiply and sum over modes. This is the only per-probe work.
7. **Route.** Each detector reads the branch its band belongs to.

Two parts do the work.

**Mid-plane referencing.** Tilting the illumination shears the specimen, so once the ramp and the vacuum chirp are removed, the only remaining tilt dependence is a linear phase whose slope is the lateral drift `lambda (t - z) q` an electron accumulates between scattering at depth `z` and leaving at the exit surface. Referenced to the exit surface, that drift spectrum is one-sided on `[0, t theta]`, while a band-limited interpolant resolves a spectrum centred on zero. Half the Nyquist budget was going to an empty half-interval. Back-propagating the beams to `t/2` centres the support and doubles the angle the same beams reach, from `L / (2 f t)` to `L / (f t)`. The propagator does not commute with the tilt ramp, so it cannot be undone per beam once the modes are formed. It does commute with the reduction, so it is left in place and only wave-function and real-space outputs are propagated back.

**Exact retention of the built beams' row space.** The compression keeps that row space whole and applies `tolerance` only to what is left over. One branch of the reduction is then the PRISM algorithm exactly, at any tolerance, so detectors above the interpolation's aliasing angle read a branch with PRISM's accuracy from the same build. The switch angle is snapped down to a detector boundary so no band straddles it. A pixelated detector spans both sides by construction, and its pattern is stitched on the simulation grid: the interpolated branch zero-padded, the plane-wave branch scattered onto every f-th pixel. That scatter is exact because a field periodic with period `L/f` has its spectrum supported on those points.

The mode count comes out close to the coarse beam count, so a probe is reduced from the full expansion for roughly the number of terms PRISM needs for the coarse one. Per-probe cost relative to PRISM at the same `f` is `(f W / N_a)^2`, the square of the window ratio, and that is the whole penalty. The algorithm tolerates a larger `f` at the same accuracy, so the comparison that matters is at matched error.

## Contents

| File | |
|---|---|
| `abtem/prism/s_matrix.py` | the compression, `CompressedSMatrixArray`, the reductions and the detector routing |
| `abtem/scan.py` | `GridScan.commensurate` |
| `abtem/__init__.py` | export `CompressedSMatrixArray` |
| `abtem/prism/utils.py` | azimuthal convention fix, below |
| `abtem/potentials/iam.py` | exit-plane mapping fix, below |
| `abtem/prism/_partitioned_s_matrix.py` | same convention in an unreachable module, plus a note saying so |
| `test/test_prism_upsample.py` | 78 cases over both devices |
| `test/test_prism.py`, `test/test_potentials.py` | regression tests for the two fixes |

### Commensurate scans

When the probe positions land on a whole number of pixels, the window offset splits as `j - c = a s + b` and the probe index enters only through the whole-step part, so strided views and one matrix product replace the per-probe gather. Measured 20 to 50 times faster on the GPU at identical output. `GridScan.commensurate(matrix_or_potential, gpts=...)` snaps a requested scan onto that grid. An ordinary `GridScan` still works and warns if its step misses it.

### Two fixes that are not C-PRISM

**Azimuthal convention** in `abtem/prism/utils.py`: `prism_coefficients` computed `arctan2(kx, ky)` where `polar_spatial_frequencies` and the rest of the package use `arctan2(ky, kx)`. This reflects any azimuthally dependent aberration (astigmatism, coma) in a PRISM reduction with a `CTF`. Pinned by `test_upsample_aberrated_ctf_matches_probe`, which compares a real-space aberrated probe intensity and is sensitive to the convention.

**Exit planes on slicing** in `PotentialArray`: slicing copied the parent's `exit_planes` unmapped, so a sliced potential could carry an exit plane outside its own slices and the multislice algorithm silently returned an unpropagated wave function. Pinned by `test_potential_array_slicing_maps_exit_planes`.

## Limits

- The usable interpolation factor falls with thickness as `f < L / (alpha t)`, because the drift the interpolation resolves grows with thickness. Mid-plane referencing buys a factor of two in that bound and no more. `blend_angle='auto'` warns and clamps when the aliasing angle falls inside the bright-field disk.
- The reduction window has to hold the probe at the worst position in the scan. The automatic estimate uses thickness and aperture only, so it does not know about aberrations. A scan at large defocus needs `window_gpts` set explicitly.
- The compressed representation is a per-pixel product of a mode image and a probe-centred kernel, so multislice does not pass through it and a specimen cannot be split into slabs that are compressed separately and recombined. Tested, and it fails as the algebra predicts.
- Peak memory is during the compression, where the built scattering matrix and the modes are both resident, a little over twice what survives. `max_rank` trades accuracy for both the resident size and the reduction time, in proportion.

## Testing

`test/test_prism_upsample.py` is new: 78 cases parametrized over `[gpu, "cpu"]`. It pins the exactness claims directly rather than only end to end, including that the plane-wave branch reproduces an independently built PRISM scattering matrix at every tolerance, that the lattice reduction equals the general one, that stitched patterns match PRISM at the lattice pixels and are exactly zero off them, and that the whole chain honours `config['precision']`.

A100-SXM4-40GB, CUDA 12.9, cupy 14.0.1, numpy 2.0.2, excluding `test_gpaw.py`:

```
prism: 281 passed, 0 failed, 0 skipped (393 s); 136 gpu cases, 136 passed, 0 skipped
full : 1563 passed, 0 failed, 34 skipped (1016 s); 622 gpu cases, 610 passed, 12 skipped
```

The 12 skipped GPU cases are `test_hyperspy`, which skips itself when hyperspy is not installed.


cophus:
# C-PRISM: upsampled, compressed scattering-matrix reduction via `SMatrix(upsample=True)`

**Revision note.** Following review feedback, C-PRISM is no longer a separate pipeline:
the `CPRISM`/`CPRISMArray` classes are gone, and the algorithm is now an option on
`SMatrix` itself. There is one builder class and one set of entry points
(`SMatrix.build` / `SMatrix.scan` / `SMatrix.reduce`); C-PRISM is enabled by two
keywords. With `upsample=False` (the default) every code path and result is unchanged
from `dev`, which the test suite pins.

```python
# PRISM (unchanged)
measurement = abtem.SMatrix(
    potential=potential, energy=150e3, semiangle_cutoff=20, interpolation=(2, 4),
).scan(scan=scan, detectors=haadf)

# C-PRISM: same class, same scan — two extra keywords
measurement = abtem.SMatrix(
    potential=potential, energy=150e3, semiangle_cutoff=20, interpolation=(2, 4),
    upsample=True, tolerance=1e-3,
).scan(scan=scan, detectors=haadf)
```

## What it does

PRISM trades accuracy for speed through the interpolation factor `f`: the plane-wave
expansion is subsampled by `f`, and every probe is cropped to a `1/f` window, so the
error grows quickly with `f` (the crop discards the probe's scattering tails, and the
aperture is sampled ever more coarsely).

With `upsample=True`, the scattering matrix is still *built* from the coarse expansion
(the multislice cost still drops by ~`f²`), but it is then interpolated back to the
**full plane-wave expansion of the aperture** and compressed by an exact truncated
SVD. Probes are reduced from the full expansion on the full grid: no window crop, no
coarsened aperture sampling. The interpolation factor only controls how many
multislice runs are used to build the scattering matrix — accuracy degrades much more
slowly with `f`, and probe positions commensurate with the grid stay commensurate at
any `f`.

`SMatrix.build` returns a `CompressedSMatrixArray` (the truncated factorization
`T ≈ U Σ Vᴴ`) instead of an `SMatrixArray`; both expose the same
`scan`/`reduce`/`dummy_probes` interface. At `interpolation=(1, 1)` the expansion is
already complete, so a plain `SMatrixArray` is returned and the result is *identical*
to PRISM (tested to machine precision).

## New `SMatrix` keywords

| keyword | default | meaning |
|---|---|---|
| `upsample` | `False` | Enable the C-PRISM build: interpolate the coarse expansion to the full aperture and compress it. |
| `tolerance` | `1e-3` | Relative singular-value threshold of the adaptive truncation; the retained rank is certified against the exact spectrum of the interpolated operator. |
| `max_rank` | `None` | Optional cap on the number of retained modes. |
| `window_gpts` | `None` | Optional crop window for the compressed reduction. Decoupled from `interpolation` (unlike PRISM); default is the full grid. |
| `position_quantization` | `None` | Optionally quantize fractional probe positions, limiting the number of reduction kernels for scans incommensurate with the grid. |
| `max_batch_expansion` | `"auto"` | Stream the full-window reduction in batches of plane waves, bounding its memory (see *Memory* below). |

`max_rank`, `position_quantization` and `max_batch_expansion` raise a `ValueError`
without `upsample=True`; `tolerance` and `window_gpts` are documented as ignored. The
compression stage reports progress under `diagnostics.task_progress`.

## API examples

One-shot scan, including frozen phonons (configurations are built, compressed and
reduced one at a time and averaged, eagerly or lazily):

```python
frozen_phonons = abtem.FrozenPhonons(atoms, num_configs=8, sigmas=0.1, seed=0)
potential = abtem.Potential(frozen_phonons, gpts=(384, 768), slice_thickness=2)

measurement = abtem.SMatrix(
    potential=potential, energy=150e3, semiangle_cutoff=20,
    interpolation=(2, 4), upsample=True,
).scan(scan=scan, detectors=haadf, lazy=True).compute()
```

Build once, reduce many — the compressed scattering matrix is reusable across scans,
detectors and CTFs:

```python
s_matrix = abtem.SMatrix(
    potential=potential, energy=150e3, semiangle_cutoff=20,
    interpolation=(2, 4), upsample=True, tolerance=1e-3,
)
s_matrix_array = s_matrix.build()  # CompressedSMatrixArray, always in memory

haadf_image = s_matrix_array.scan(scan=scan, detectors=haadf)
aberrated = s_matrix_array.scan(
    scan=scan, detectors=haadf,
    ctf=abtem.CTF(defocus=50, Cs=1e5, semiangle_cutoff=20),
)
cbed = s_matrix_array.reduce(scan=(21.5, 8.3), detectors=abtem.WavesDetector())
```

Choosing the tolerance from the retained spectrum:

```python
plt.semilogy(s_matrix_array.singular_values / s_matrix_array.singular_values[0])
s_matrix_array.rank  # modes kept at the current tolerance
```

Bounding the reduction memory for large fields of view (see *Memory* below); the
option can also be set per reduction:

```python
s_matrix = abtem.SMatrix(
    potential=potential, energy=150e3, semiangle_cutoff=20,
    interpolation=(2, 4), upsample=True, max_batch_expansion=256,
)
# or: s_matrix_array.reduce(scan=scan, detectors=haadf, max_batch_expansion=256)
```

## How the compressed build works

1. **Coarse build on a disk.** The coarse expansion spans a disk around the aperture
   (padded by a support margin) instead of the full bounding rectangle — the corners
   are undersampled at coarse `f` and only add multislice runs (−26% plane waves at
   `(2, 4)`, −45% at `(4, 8)`).
2. **Phase removal.** The plane-wave tilt phase *and* the propagation (defocus) phase
   `exp(−iπλt|k|²)` accumulated through the specimen thickness are factored out of
   each beam, exposing the smooth variation of the scattering matrix with the wave
   vector. Both are restored on the dense expansion afterwards.
3. **Interpolate, then compress.** The phase-removed matrix is factored once as
   `T = LQ` with orthonormal `Q`, obtained from the Hermitian eigendecomposition of
   the small Gram matrix `TTᴴ` (accumulated in double precision) rather than a tall
   QR — 5–14× faster and one full matrix copy less of peak memory, since `Q` is
   contracted on the fly and never materialized. A band-limited (trigonometric)
   interpolant maps the small beam-side factor to the dense expansion, and the SVD of
   the interpolated operator follows exactly from the small projected matrix — the
   dense matrix is never formed, and the tolerance is certified on the operator the
   scan actually applies. A band-limited interpolant is essential: local (spline)
   interpolation aliases its error into ghost probes displaced by `extent/f`, which
   annular detectors integrate as signal.
4. **Reduce by expansion.** By default the compressed matrix is expanded back to the
   interpolation-(1, 1) scattering matrix and reduced with the existing
   `SMatrixArray` machinery (one high-arithmetic-intensity GEMM); with `window_gpts`
   set, a windowed per-position mode contraction is used instead.

`scan`/`reduce` reuse the existing per-configuration orchestration (frozen phonons,
CTF ensembles, eager and lazy evaluation); the compression requires the scattering
matrix in memory, so upsampled reductions run one potential configuration per task
(`disable_s_matrix_chunks`).

## Accuracy

Benchmark: SrTiO₃/LaTiO₃ (110) interface, 55 Å thick, 150 kV, 20 mrad, gpts (384, 768);
single-probe CBED error on an atomic column, in absolute intensity units
(`½·Σ|ΔI|` per unit incident intensity, masked ≥4 px from the detector edge), vs. a
multislice reference:

| interpolation | PRISM | `upsample=True` |
|---|---|---|
| (1, 1) | 1.1e−06 | 1.1e−06 (identical) |
| (1, 2) | 7.8e−03 | **3.6e−03** |
| (2, 4) | 3.1e−02 | **1.6e−02** |
| (4, 8) | 8.0e−02 | **6.1e−02** |

On scanned images the gap is larger, because PRISM's window crop error is coherent
across the scan while the C-PRISM error is not: at `(2, 4)` the BF image error is
~30× lower (9e−4 vs 2.8e−2) and the HAADF error ~5× lower (9e−5 vs 4.4e−4) in the
same absolute units.

## Cost

Same benchmark, full Nyquist scan (7808 positions), single-threaded-FFT CPU timings:

| interpolation | PRISM | `upsample=True` (retained modes) |
|---|---|---|
| multislice reference | 1002 s | — |
| (1, 1) | 1567 waves, 216 s | identical path, 211 s |
| (1, 2) | 795 waves, 108 s | 337 modes, 167 s |
| (2, 4) | 211 waves, 27 s | 246 modes, 70 s |
| (4, 8) | 61 waves, 8.5 s | 86 modes, 42 s |

The multislice build cost falls with `f` exactly as PRISM's; the compression itself
is now sub-second at this scale (the Gram-based factorization replaces a tall QR
that took 5–23 s here and ~9 min on a 10⁶-pixel scattering matrix); the remaining
fixed cost is the full-aperture reduction (≈ a PRISM-(1,1) reduction, so the total
decreases monotonically with `f`). At a matched time budget the two algorithms trade
differently per detector: C-PRISM's time-vs-error front is below PRISM's for
BF/LAADF/CBED work, while PRISM remains competitive for HAADF-only mapping on CPU,
where its crop error hides inside the BF disk. On GPU the reduction GEMM is the same
kernel PRISM uses.

**Memory.** The default full-window reduction materializes the expanded
interpolation-(1, 1) scattering matrix, i.e. the memory footprint of PRISM at
`interpolation=1`. For large fields of view, `max_batch_expansion` streams the
expansion instead: plane waves are expanded in batches and contracted with the
reduction coefficients on the fly, bounding the peak at one batch of plane waves
plus one batch of reduced wave functions (the automatic reduction batches are
enlarged so the repeated-expansion overhead stays at `rank / batch`). Measured on a
case whose expansion is 3.5 GB: peak memory +3.6 GB → **+0.2 GB** at unchanged wall
time, with the streamed and expanded reductions agreeing to float precision.

## PRISM fixes found while validating

Found while pinning C-PRISM against PRISM/`Probe`, all in `dev`'s code paths:

- The CTF expansion-coefficient normalization used the complex square instead of the
  absolute square, scaling reduced probes of any aberrated CTF by an arbitrary factor
  (2.7× intensity for a typical aberrated probe, different per CTF-ensemble member).
  PRISM with an aberrated CTF now matches the equivalent `Probe` to float precision.
- `SMatrixArray.scan` forwarded a `rechunk` argument that `reduce` does not accept,
  raising `TypeError` on every direct call.
- The azimuthal angle in `prism_coefficients` and the partitioned scattering matrix
  was `arctan2(kx, ky)`, mirroring φ-dependent aberrations relative to the abTEM
  convention `arctan2(ky, kx)` (latent — these paths currently only see symmetric
  CTFs).

Regression tests pin the coefficient conventions.

## Tests

`test/test_prism_upsample.py` (24 tests): vacuum probe identity at several factors
and rank-1 vacuum compression; multislice agreement at `(1, 1)` including off-grid
positions; identity with PRISM at `(1, 1)` (returns a plain `SMatrixArray`); lower
error than PRISM at matched interpolation with an absolute-error bound; windowed
reduction and window normalization; streamed-expansion agreement with the expanded
reduction (grid scans, off-grid positions, aberrated CTFs, eager and lazy) and its
option validation; the singular-value spectrum; frozen phonons (eager + lazy) and
lazy≡eager; multiple detector types; CTF ensembles; aberrated-CTF probe match (pins
the azimuth convention); interpolation-independent `downsampled_gpts`; defocus-phase
properties; option validation; and PRISM copy-kwarg round-trip. The API refactor and
the Gram-based factorization were each verified against the previous revision
(bit-identical scans / 3e-7 relative drift). Full suite: 836 passed.

---

Updated notebooks:
[will be added]

cophus:
# C-PRISM: upsampled, compressed scattering-matrix reduction via `SMatrix(upsample=True)`

**Revision note.** Following review feedback, C-PRISM is no longer a separate pipeline:
the `CPRISM`/`CPRISMArray` classes are gone, and the algorithm is now an option on
`SMatrix` itself. There is one builder class and one set of entry points
(`SMatrix.build` / `SMatrix.scan` / `SMatrix.reduce`); C-PRISM is enabled by two
keywords. With `upsample=False` (the default) every code path and result is unchanged
from `dev`, which the test suite pins.

```python
# PRISM (unchanged)
measurement = abtem.SMatrix(
    potential=potential, energy=150e3, semiangle_cutoff=20, interpolation=(2, 4),
).scan(scan=scan, detectors=haadf)

# C-PRISM: same class, same scan — two extra keywords
measurement = abtem.SMatrix(
    potential=potential, energy=150e3, semiangle_cutoff=20, interpolation=(2, 4),
    upsample=True, tolerance=1e-3,
).scan(scan=scan, detectors=haadf)
```

## What it does

PRISM trades accuracy for speed through the interpolation factor `f`: the plane-wave
expansion is subsampled by `f`, and every probe is cropped to a `1/f` window, so the
error grows quickly with `f` (the crop discards the probe's scattering tails, and the
aperture is sampled ever more coarsely).

With `upsample=True`, the scattering matrix is still *built* from the coarse expansion
(the multislice cost still drops by ~`f²`), but it is then interpolated back to the
**full plane-wave expansion of the aperture** and compressed by an exact truncated
SVD. Probes are reduced from the full expansion on the full grid: no window crop, no
coarsened aperture sampling. The interpolation factor only controls how many
multislice runs are used to build the scattering matrix — accuracy degrades much more
slowly with `f`, and probe positions commensurate with the grid stay commensurate at
any `f`.

`SMatrix.build` returns a `CompressedSMatrixArray` (the truncated factorization
`T ≈ U Σ Vᴴ`) instead of an `SMatrixArray`; both expose the same
`scan`/`reduce`/`dummy_probes` interface. At `interpolation=(1, 1)` the expansion is
already complete, so a plain `SMatrixArray` is returned and the result is *identical*
to PRISM (tested to machine precision).

## New `SMatrix` keywords

| keyword | default | meaning |
|---|---|---|
| `upsample` | `False` | Enable the C-PRISM build: interpolate the coarse expansion to the full aperture and compress it. |
| `tolerance` | `1e-3` | Relative singular-value threshold of the adaptive truncation; the retained rank is certified against the exact spectrum of the interpolated operator. |
| `max_rank` | `None` | Optional cap on the number of retained modes. |
| `window_gpts` | `None` | Optional crop window for the compressed reduction. Decoupled from `interpolation` (unlike PRISM); default is the full grid. |
| `position_quantization` | `None` | Optionally quantize fractional probe positions, limiting the number of reduction kernels for scans incommensurate with the grid. |
| `max_batch_expansion` | `"auto"` | Stream the full-window reduction in batches of plane waves, bounding its memory (see *Memory* below). |

`max_rank`, `position_quantization` and `max_batch_expansion` raise a `ValueError`
without `upsample=True`; `tolerance` and `window_gpts` are documented as ignored. The
compression stage reports progress under `diagnostics.task_progress`.

## API examples

One-shot scan, including frozen phonons (configurations are built, compressed and
reduced one at a time and averaged, eagerly or lazily):

```python
frozen_phonons = abtem.FrozenPhonons(atoms, num_configs=8, sigmas=0.1, seed=0)
potential = abtem.Potential(frozen_phonons, gpts=(384, 768), slice_thickness=2)

measurement = abtem.SMatrix(
    potential=potential, energy=150e3, semiangle_cutoff=20,
    interpolation=(2, 4), upsample=True,
).scan(scan=scan, detectors=haadf, lazy=True).compute()
```

Build once, reduce many — the compressed scattering matrix is reusable across scans,
detectors and CTFs:

```python
s_matrix = abtem.SMatrix(
    potential=potential, energy=150e3, semiangle_cutoff=20,
    interpolation=(2, 4), upsample=True, tolerance=1e-3,
)
s_matrix_array = s_matrix.build()  # CompressedSMatrixArray, always in memory

haadf_image = s_matrix_array.scan(scan=scan, detectors=haadf)
aberrated = s_matrix_array.scan(
    scan=scan, detectors=haadf,
    ctf=abtem.CTF(defocus=50, Cs=1e5, semiangle_cutoff=20),
)
cbed = s_matrix_array.reduce(scan=(21.5, 8.3), detectors=abtem.WavesDetector())
```

Choosing the tolerance from the retained spectrum:

```python
plt.semilogy(s_matrix_array.singular_values / s_matrix_array.singular_values[0])
s_matrix_array.rank  # modes kept at the current tolerance
```

Bounding the reduction memory for large fields of view (see *Memory* below); the
option can also be set per reduction:

```python
s_matrix = abtem.SMatrix(
    potential=potential, energy=150e3, semiangle_cutoff=20,
    interpolation=(2, 4), upsample=True, max_batch_expansion=256,
)
# or: s_matrix_array.reduce(scan=scan, detectors=haadf, max_batch_expansion=256)
```

## How the compressed build works

1. **Coarse build on a disk.** The coarse expansion spans a disk around the aperture
   (padded by a support margin) instead of the full bounding rectangle — the corners
   are undersampled at coarse `f` and only add multislice runs (−26% plane waves at
   `(2, 4)`, −45% at `(4, 8)`).
2. **Phase removal.** The plane-wave tilt phase *and* the propagation (defocus) phase
   `exp(−iπλt|k|²)` accumulated through the specimen thickness are factored out of
   each beam, exposing the smooth variation of the scattering matrix with the wave
   vector. Both are restored on the dense expansion afterwards.
3. **Interpolate, then compress.** The phase-removed matrix is factored once as
   `T = LQ` with orthonormal `Q`, obtained from the Hermitian eigendecomposition of
   the small Gram matrix `TTᴴ` (accumulated in double precision) rather than a tall
   QR — 5–14× faster and one full matrix copy less of peak memory, since `Q` is
   contracted on the fly and never materialized. A band-limited (trigonometric)
   interpolant maps the small beam-side factor to the dense expansion, and the SVD of
   the interpolated operator follows exactly from the small projected matrix — the
   dense matrix is never formed, and the tolerance is certified on the operator the
   scan actually applies. A band-limited interpolant is essential: local (spline)
   interpolation aliases its error into ghost probes displaced by `extent/f`, which
   annular detectors integrate as signal.
4. **Reduce by expansion.** By default the compressed matrix is expanded back to the
   interpolation-(1, 1) scattering matrix and reduced with the existing
   `SMatrixArray` machinery (one high-arithmetic-intensity GEMM); with `window_gpts`
   set, a windowed per-position mode contraction is used instead.

`scan`/`reduce` reuse the existing per-configuration orchestration (frozen phonons,
CTF ensembles, eager and lazy evaluation); the compression requires the scattering
matrix in memory, so upsampled reductions run one potential configuration per task
(`disable_s_matrix_chunks`).

## Accuracy

Benchmark: SrTiO₃/LaTiO₃ (110) interface, 55 Å thick, 150 kV, 20 mrad, gpts (384, 768);
single-probe CBED error on an atomic column, in absolute intensity units
(`½·Σ|ΔI|` per unit incident intensity, masked ≥4 px from the detector edge), vs. a
multislice reference:

| interpolation | PRISM | `upsample=True` |
|---|---|---|
| (1, 1) | 1.1e−06 | 1.1e−06 (identical) |
| (1, 2) | 7.8e−03 | **3.6e−03** |
| (2, 4) | 3.1e−02 | **1.6e−02** |
| (4, 8) | 8.0e−02 | **6.1e−02** |

On scanned images the gap is larger, because PRISM's window crop error is coherent
across the scan while the C-PRISM error is not: at `(2, 4)` the BF image error is
~30× lower (9e−4 vs 2.8e−2) and the HAADF error ~5× lower (9e−5 vs 4.4e−4) in the
same absolute units.

## Cost

Same benchmark, full Nyquist scan (7808 positions), single-threaded-FFT CPU timings:

| interpolation | PRISM | `upsample=True` (retained modes) |
|---|---|---|
| multislice reference | 1002 s | — |
| (1, 1) | 1567 waves, 216 s | identical path, 211 s |
| (1, 2) | 795 waves, 108 s | 337 modes, 167 s |
| (2, 4) | 211 waves, 27 s | 246 modes, 70 s |
| (4, 8) | 61 waves, 8.5 s | 86 modes, 42 s |

The multislice build cost falls with `f` exactly as PRISM's; the compression itself
is now sub-second at this scale (the Gram-based factorization replaces a tall QR
that took 5–23 s here and ~9 min on a 10⁶-pixel scattering matrix); the remaining
fixed cost is the full-aperture reduction (≈ a PRISM-(1,1) reduction, so the total
decreases monotonically with `f`). At a matched time budget the two algorithms trade
differently per detector: C-PRISM's time-vs-error front is below PRISM's for
BF/LAADF/CBED work, while PRISM remains competitive for HAADF-only mapping on CPU,
where its crop error hides inside the BF disk. On GPU the reduction GEMM is the same
kernel PRISM uses.

**Memory.** The default full-window reduction materializes the expanded
interpolation-(1, 1) scattering matrix, i.e. the memory footprint of PRISM at
`interpolation=1`. For large fields of view, `max_batch_expansion` streams the
expansion instead: plane waves are expanded in batches and contracted with the
reduction coefficients on the fly, bounding the peak at one batch of plane waves
plus one batch of reduced wave functions (the automatic reduction batches are
enlarged so the repeated-expansion overhead stays at `rank / batch`). Measured on a
case whose expansion is 3.5 GB: peak memory +3.6 GB → **+0.2 GB** at unchanged wall
time, with the streamed and expanded reductions agreeing to float precision.

## PRISM fixes found while validating

Found while pinning C-PRISM against PRISM/`Probe`, all in `dev`'s code paths:

- The CTF expansion-coefficient normalization used the complex square instead of the
  absolute square, scaling reduced probes of any aberrated CTF by an arbitrary factor
  (2.7× intensity for a typical aberrated probe, different per CTF-ensemble member).
  PRISM with an aberrated CTF now matches the equivalent `Probe` to float precision.
- `SMatrixArray.scan` forwarded a `rechunk` argument that `reduce` does not accept,
  raising `TypeError` on every direct call.
- The azimuthal angle in `prism_coefficients` and the partitioned scattering matrix
  was `arctan2(kx, ky)`, mirroring φ-dependent aberrations relative to the abTEM
  convention `arctan2(ky, kx)` (latent — these paths currently only see symmetric
  CTFs).

Regression tests pin the coefficient conventions.

## Tests

`test/test_prism_upsample.py` (24 tests): vacuum probe identity at several factors
and rank-1 vacuum compression; multislice agreement at `(1, 1)` including off-grid
positions; identity with PRISM at `(1, 1)` (returns a plain `SMatrixArray`); lower
error than PRISM at matched interpolation with an absolute-error bound; windowed
reduction and window normalization; streamed-expansion agreement with the expanded
reduction (grid scans, off-grid positions, aberrated CTFs, eager and lazy) and its
option validation; the singular-value spectrum; frozen phonons (eager + lazy) and
lazy≡eager; multiple detector types; CTF ensembles; aberrated-CTF probe match (pins
the azimuth convention); interpolation-independent `downsampled_gpts`; defocus-phase
properties; option validation; and PRISM copy-kwarg round-trip. The API refactor and
the Gram-based factorization were each verified against the previous revision
(bit-identical scans / 3e-7 relative drift). Full suite: 836 passed.

---

Updated notebooks:
[will be added]

TomaSusi:
Thanks! The PR description is stale after you folded the functionality into the existing API.

My Claude-assisted review revealed these issues:

- Missing GPU test parametrization (only real actionable finding). Every test in the new test/test_prism_upsample.py, and the new test_prism_aberrated_ctf_matches_probe in test/test_prism.py, runs CPU-only — none use the @pytest.mark.parametrize("device", [gpu, "cpu"]) pattern that the rest of test_prism.py already follows (e.g. test_prism_matches_probe). The compression path (xp.linalg.eigh, xp.linalg.svd, the ThreadPoolExecutor-based _reduce_to_waves) has array-module-dependent branches (e.g. threading is explicitly disabled if xp is np else 1) that get zero exercise on GPU. Given the PR body itself notes "We still need to test this module on GPU," this should be tracked before merge/release, not silently left CPU-only.

Minor / non-blocking:

- The new compression code (_compress, _expanded_slab, CompressedSMatrixArray.wave_vectors, etc.) hardcodes np.float32/np.complex128 rather than using get_dtype(), so it won't honor abtem.config['precision'] = 'float64'. This mirrors a few pre-existing hardcodes already in this file, so it's not a fresh pattern, but it's worth a follow-up since it's new code in a file that otherwise uses get_dtype() correctly elsewhere.
- window_gpts silently no-ops when passed to SMatrix(upsample=False) (set to None), while max_rank/position_quantization/max_batch_expansion raise ValueError in the same situation. This is deliberate (a comment explains it's needed so _copy_kwargs round-trips the PRISM-derived cropping window), but it's an inconsistent user-facing surface — a user who forgets upsample=True gets silent no-op instead of an error.
- The azimuthal-convention fix in _partitioned_s_matrix.py (multi-GPU partitioned path) has no dedicated test, before or after this PR — the equivalent fix in the main path is pinned by a real regression test, this one isn't.

cophus:
Updates, with help from Claude:

Thanks. All four points are addressed, `dev` is merged in, and the description is rewritten. The GPU run found one real bug, described under point 1.

**1. GPU test parametrization.** `test_prism_upsample.py` now parametrizes over `[gpu, "cpu"]` with the `from utils import gpu` marker the rest of `test_prism.py` uses. The device is threaded through the potentials, probes and scattering matrices rather than left at the config default. `test_prism_aberrated_ctf_matches_probe` is parametrized too.

A100-SXM4-40GB, CUDA 12.9, cupy 14.0.1, numpy 2.0.2:

```
prism: 281 passed, 0 failed, 0 skipped (393 s); 136 gpu cases, 136 passed, 0 skipped
full : 1563 passed, 0 failed, 34 skipped (1016 s); 622 gpu cases, 610 passed, 12 skipped
```

The 12 skipped GPU cases are `test_hyperspy`, which skips itself without hyperspy installed. Every C-PRISM and PRISM GPU case ran. This covers `_reduce_to_waves` dispatching to the batched implementation when `xp is not np` and the `ThreadPoolExecutor` path that only runs when it is, `_reduce_memory_budget` and the GPU row-block factor, the device-to-host round trips in the lattice reduction and the stitched pattern assembly, and `eigh` on the Gram matrix.

The bug it found: `_blend_weight` built the `blend_angle='aperture'` weight with `np.asarray(probe.aperture._evaluate_kernel(probe))`. The dummy probe inherits `device=self.device`, so on the GPU that argument is a cupy array and numpy refuses to convert one implicitly. The path raised `TypeError` on every GPU call. The weight now stays in its own array module.

**2. Hardcoded dtypes.** Fixed. One of them was not cosmetic: `_expanded_s_matrix_array` cast the wave vectors to `np.float64` unconditionally, and `complex_exponential` preserves its input dtype, so the expand path built `complex128` coefficients against a `complex64` scattering matrix. That is twice the coefficient memory and an fp64 GEMM on the GPU. The rest now take their real dtype from `get_dtype()`, in both branches of `SMatrix.wave_vectors` so the file keeps one convention. A new test asserts the whole chain at both precisions on both devices.

Two uses of double precision are deliberate and stay, now with the reason in a comment: the Gram matrix, whose eigenvalues are the squares of the singular values and would resolve the spectrum only to ~1e-3 relative in single precision, and the pixel positions, whose fractional part is grouped to 1e-4 of a pixel.

**3. `window_gpts` silent no-op.** Fixed at the other end. `SMatrix._copy_kwargs` passes back the value that was given instead of the derived window, so the PRISM round trip works without the no-op and `window_gpts` joins the others in raising when `upsample=False`.

**4. Partitioned azimuthal fix.** Nothing in the package imports `_partitioned_s_matrix`, and `remove_tilt` and `interpolate_full` both call `CTF.evaluate`, which does not exist anywhere in abTEM. Calling it gives `AttributeError: 'CTF' object has no attribute 'evaluate'`, on `dev` as well. The module cannot run, so the convention is unreachable and a test would mean repairing the module first. I put a note at the top of the file recording that, and pointing at `polar_spatial_frequencies` as the convention it follows. The live path stays pinned by `test_upsample_aberrated_ctf_matches_probe`. That module is a candidate for deletion or repair in its own PR. I can drop my one-line change to it if you want the diff kept to C-PRISM.

**Merge.** The only conflict was `s_matrix.py`, between the C-PRISM additions and the `_eager_build_s_matrix_detect` split plus the new `transition_potential_scan`. Both sides are kept whole.

One change to the two chunking tests. They asserted the chunked reduction is bit for bit the unchunked one. A later commit on this branch moved the mode axis to the front of the reduction operands, which changes the row count BLAS sees and so its summation order, and the last bits move on some builds (bit-identical on my machine, not on the A100). They now assert a 1e-6 relative bound and print the magnitude on failure. A chunking bug misplaces whole rows and lands far above that.

TomaSusi:
Thank you Colin!

PR #269, opened 2026-04-29
Major GPU update: memory-bounded potential chunking, GPU kernel improvements, and opt-in multi-GPU

TomaSusi:
## Summary

This PR removes the main GPU memory bottleneck in the multislice path and delivers a ~2–4× speedup for the `projection='finite'` case.  Previously `build()` placed the entire slice dimension into a single dask chunk, so the complete potential had to fit in VRAM at once — making large-grid simulations impossible regardless of how much VRAM the card had.  A suite of coordinated changes now keeps peak VRAM bounded at every stage.

---

### 1. Memory-bounded slice chunking (`iam.py`, `multislice.py`, `chunks.py`)

`multislice_and_detect` now iterates over the potential in chunks of contiguous slices rather than materialising the whole potential at once.  Each chunk is built eagerly, propagated through, then discarded before the next is built.

- New `generate_chunked_slices(chunk_size="auto")` method on `BasePotential`, `Potential`, `CrystalPotential`, and `PotentialArray`.
- `CrystalPotential` has a dedicated override that builds the unit cell once and tiles it into chunks, avoiding the 2× peak-memory spike the base-class would incur from re-accumulating slices.
- New `estimate_potential_chunk_size(gpts, device)` in `chunks.py`: queries CUDA free memory at _computation_ time (after the probe batch is already resident) and sizes each chunk to 35% of effective VRAM with a 5× per-slice cost factor to account for CuPy memory-pool fragmentation.
- New config key `potential.slice-chunk-size` (default `"auto"`); set to a fixed integer to override.
- On CPU the default is no chunking (single chunk = pre-existing behaviour); chunking can be enabled via the config key when RAM is genuinely limited.

### 2. Coordinated VRAM-aware scan batch sizing (`waves.py`, `chunks.py`)

- New `estimate_scan_batch_size(gpts, dtype, device)` in `chunks.py`: allocates up to 50% of effective VRAM for the probe batch at _graph-construction_ time (before the probe is resident), with a 6× per-probe overhead factor.
- Because `estimate_potential_chunk_size` runs at computation time (probe already resident), the two estimates naturally see the same VRAM and stay coordinated without explicit cross-referencing.
- Batch size is snapped to the nearest power-of-two for GPU-friendly memory alignment.

### 3. CuPy `RawKernel` for radial table interpolation (`core/_cuda.py`, `integrals.py`)

The Numba CUDA JIT kernel for finite-projection table interpolation is replaced by a C++ CuPy `RawKernel`:

[...]

PR #324, opened 2026-07-06
Add phonon-loss diffraction and momentum-resolved spectral measurements

TomaSusi:
## Summary

Adds a phonon-loss / thermal-diffuse-scattering (TDS) energy-loss workflow: an energy-resolved frozen-phonon ensemble, functions to turn its exit waves into energy-resolved diffraction patterns and momentum-resolved spectra S(q, E), two new spectral detectors, and detailed-balance thermal weighting to split the classically-computed TDS signal into quantum loss/gain sides.

### 1. Build an energy-resolved frozen-phonon ensemble

`EnergyResolvedAtomsEnsemble` wraps a 2D array of `Atoms` snapshots `(n_energies, n_configs)` — one group of frozen-phonon configurations per energy bin (e.g. one phonon-mode energy, or an energy-transfer magnitude from an external MD/lattice-dynamics calculation):

```python
import numpy as np
from abtem import EnergyResolvedAtomsEnsemble, Potential, PlaneWave

energies = [0.0, 0.02, 0.05, 0.10]  # eV — must start at 0 and increase
# one list of Atoms snapshots per energy in `energies`
snapshots_per_energy = [make_snapshots(e, n_configs=20) for e in energies]

ensemble = EnergyResolvedAtomsEnsemble(
    energy_resolved_snapshots=snapshots_per_energy,
    energies=energies,
    ensemble_mean=False,  # keep all configs — needed for phonon_loss_diffraction_patterns
)

# projection="finite" is recommended over the default "infinite" whenever the
# potential uses more than one slice — see the "Also in this PR" note below.
potential = Potential(ensemble, sampling=0.05, slice_thickness=1.0, projection="finite")
exit_waves = PlaneWave(energy=100e3).multislice(potential, lazy=False)
# exit_waves.ensemble_axes_metadata == [EnergyLossAxis(...), FrozenPhononsAxis(...)]
```

### 2. Compute coherent / incoherent / TDS diffraction patterns

`phonon_loss_diffraction_patterns` collapses the `FrozenPhononsAxis` into the standard multislice-TDS decomposition, per energy bin:

```python
from abtem import phonon_loss_diffraction_patterns
# or: exit_waves.phonon_loss_diffraction_patterns(...)

dp_tds = phonon_loss_diffraction_patterns(exit_waves)  # component="tds" is the default
dp_tds.array.shape  # (n_energies, ny, nx) — FrozenPhononsAxis gone, EnergyLossAxis kept

dp_all = phonon_loss_diffraction_patterns(exit_waves, component="all")
# stacks [coherent, incoherent, tds] along a new leading OrdinalAxis(label="component")
```

`component` selects which of `I_coherent = |FT(Σψ)|²/N²`, `I_incoherent = Σ|FT(ψ)|²/N`, or `I_tds = I_incoherent - I_coherent` to return (`"all"` returns all three).

### 3. Thermal weighting: unfold into quantum loss/gain sides

A single multislice run per energy *magnitude* gives the classical, loss/gain-symmetric TDS intensity. Splitting it into the physical loss (`+E`, electron loses energy / phonon created) and gain (`-E`, electron gains energy / phonon annihilated) sides is a quantum, temperature-dependent effect (Bose-Einstein detailed balance) — not something the frozen-phonon sampling itself distinguishes. Passing `temperature` does that split:

```python
dp_weighted = phonon_loss_diffraction_patterns(
    exit_waves, component="tds", temperature=300.0  # K
)
# EnergyLossAxis is now signed and mirrored:
# [-0.10, -0.05, -0.02, 0.0, 0.02, 0.05, 0.10]  (for energies=[0.0, 0.02, 0.05, 0.10])
```

For each nonzero `E`, with `n = 1 / (exp(E / kB·T) - 1)`:

```
I_loss(+E) = I_tds(E) * (n + 1) / (2n + 1)     # phonon creation
I_gain(-E) = I_tds(E) *  n      / (2n + 1)     # phonon annihilation
```

`I_loss(+E) + I_gain(-E) == I_tds(E)` — the split conserves the total spectral weight computed classically at each energy magnitude. The zero-energy bin passes through unweighted (no loss/gain asymmetry at zero energy transfer). Requires `component="tds"` and energies starting at 0 and strictly increasing; raises `ValueError` otherwise.

*(Adapted from P. Zeiger's original DrProbe-based workflow. His cross-mode zero-loss-peak synthesis — averaging each mode's coherent component into a shared ZLP estimate, needed because his sampling ran one phonon mode at a time — is intentionally not reproduced here: it doesn't apply to this ensemble model, where the elastic/ZLP is a separate, standard calculation.)*

### 4. Integrate over q: momentum-resolved spectrum S(q, E)

`momentum_resolved_spectrum` integrates a `DiffractionPatterns` with an `EnergyLossAxis` over q using either detector:

```python
from abtem import SpectralAnnularDetector, SpectralSlitDetector, momentum_resolved_spectrum

annular_detector = SpectralAnnularDetector(outer=2.0, q_min=0.0, q_max=40.0)
spectrum = momentum_resolved_spectrum(dp_weighted, annular_detector)
spectrum.array.shape  # (n_q, n_E)

# equivalent perpendicular acceptance via a slit instead of a disk sweep:
slit_detector = SpectralSlitDetector(width=4.0, q_min=0.0, q_max=40.0)
spectrum_slit = momentum_resolved_spectrum(dp_weighted, slit_detector)
```

`SpectralAnnularDetector` sweeps an offset disk (radius `outer`) along q; `SpectralSlitDetector` integrates rectangular strips via interpolation. Both share the same `q_min`/`q_max` convention, and both accept `q_sampling` to bin down the native pixel resolution. Both work identically whether the input is eager or lazy (`.compute()`d later) and on CPU or GPU.

### 5. Visualize

```python
spectrum.show()  # pcolormesh of S(q, E), q [mrad] vs energy loss

dp_weighted.show(explode=True)  # per-energy-bin diffraction patterns
```

## Also in this PR

**Slice-boundary hopping pitfall (documentation).** With the default `projection="infinite"`, `Potential` assigns each atom to exactly one slice via a hard z cutoff with no padding (`SliceIndexedAtoms`). If an `EnergyResolvedAtomsEnsemble`'s configurations include out-of-plane displacement, an atom near a slice boundary can flip its entire potential contribution between slices across near-identical configurations, producing spurious discontinuities in the resulting spectra — confirmed on a real graphene/MACE phonon-loss workflow; `projection="finite"` (padded `SlicedAtoms`) resolved it. Single-slice runs are unaffected since there's no boundary to cross. Documented on `EnergyResolvedAtomsEnsemble` and `phonon_loss_diffraction_patterns`.

**Bug fixes found while building this feature:**
- `SpectralSlitDetector`'s detector mask (`_slit_detector_mask`) tested pixel membership against an axis-aligned bounding box of the rotated slit corners, correct only when `angle` is a multiple of 90° — verified ~6x area overcount at 45°. Fixed by rotating the pixel grid into the slit's local frame instead. Not exposed via `momentum_resolved_spectrum` (which uses `interpolate_line` and always handled rotation correctly) — affected `SpectralSlitDetector.detect()`/`.get_detector_region()` used directly. Also tightened `corners=`-mode validation, which silently dropped a conflicting `angle`/`q_min`/`offset` instead of raising as documented.
- `MomentumResolvedSpectrum.show(explode=...)` mapped grid coordinates to ensemble axes positionally in ascending order rather than by the order `explode` was given, so a non-ascending sequence (e.g. `explode=[2, 0]`) silently plotted the wrong data under a correctly-labeled title (or raised `IndexError` for incompatible axis sizes). Fixed by mapping by axis identity. No test coverage existed for `explode` before this.
- Lazy (`.compute()`d later) exit_waves on GPU hit `TypeError: Unsupported type <class 'dask.array.core.Array'>` in several places: `_thermal_weight_tds`'s `concatenate`/`flip`, `component="all"`'s `stack`, and `momentum_resolved_spectrum`'s `moveaxis`/`stack`/`tensordot` — CuPy's own array functions (unlike NumPy, which dispatches to dask via `__array_function__`) don't accept a dask array. Routed through dask's own implementations whenever the array is still lazy, regardless of device (`_array_module_fn` helper). Separately found and fixed: a fragmented energy axis (one dask chunk per energy value, `EnergyResolvedAtomsEnsemble`'s default chunking) caused `interpolate_line` + q-binning to silently compute a wrong-sized result — consolidated into a single chunk at the source. Also found a genuine dask 2025.11.0 bug where `da.tensordot` with negative axis indices computes the wrong data while reporting the correct shape — worked around with positive indices.
- `MomentumResolvedSpectrum.show()` called `np.asarray()` directly on possibly GPU-resident data; CuPy deliberately blocks that implicit conversion. Fixed using the `.get()`-if-available pattern already used elsewhere in the codebase (`noise.py`).

**Shared with #257 (energy-ensemble support):** `EnergyLossAxis` (`abtem/core/axes.py`) is a class distinct from `EnergyAxis`, so this branch's phonon-loss energy axis can't collide with #257's energy-ensemble beam-energy axis regardless of merge order. Paul Zeiger separately fixed `Waves.apply_ctf`/`.wavelength` to resolve the per-member energy for a multi-energy ensemble (previously silently mutated `accelerator.energy` and applied one wavelength to every member), and made `CTF`/`Aperture` accept a raw list/array of energies.

**Performance:** `momentum_resolved_spectrum`'s `SpectralAnnularDetector` path no longer recomputes the O(gpts²) spatial-frequency grid on every q-step.

**Merged in `dev`** (most recently: PR #269's memory-bounded potential chunking / GPU kernel improvements / opt-in multi-GPU, linear-scaling PRISM-EELS, the new C-PRISM algorithm, and #333's eager-multislice-with-2-ensemble-axes fix, among ~70 commits). Three conflicts resolved by hand: `waves.py` (this branch's per-energy dask rechunking vs. dev's potential-reuse prebuild — kept both, rechunking first so the prebuild check sees the true final chunk count), `prism/s_matrix.py` (this branch's multi-energy `SMatrix` support vs. dev's new PRISM upsample/validation options — both kept), `test/test_prism.py` (trivial, kept dev's added regression test). Verified after merging: full suite green, plus direct smoke tests of multi-energy `SMatrix` combined with dev's `upsample` validation, and multi-energy `Probe.scan` combined with dev's potential-prebuild optimization.

**Then #257 (energy-ensemble support) merged into `dev` as a squash commit**, which — since squashing discards the shared commit history this branch and #257 both descended from — reintroduced conflicts in files this branch had already reconciled with #257's pre-squash history: `core/axes.py`, `detectors.py`, `transfer.py`, `waves.py` (all purely additive — this branch's own code vs. an empty/absent counterpart on the dev side, since dev's squashed version predates it) and `test/test_energy_ensemble.py` (add/add — took dev's more complete version, which included newer SMatrix reduce/scan energy-ensemble tests this branch's copy predated). One conflict needed a real decision rather than just picking a side: `detectors.py`'s `get_detector_region` had this branch computing `energy` via the type-aware `_gpts_and_sampling_from_obj` (handles both `Waves` and `DiffractionPatterns` inputs) vs. dev calling `_energy_from_waves(waves)` directly, which assumes a `.energy` attribute `DiffractionPatterns` doesn't have — kept this branch's version and verified `get_detector_region()` with a `DiffractionPatterns` input directly (which dev's side would have raised on). Full suite green after this merge too (1119 passed), plus direct smoke tests of multi-energy `apply_ctf`/`.wavelength` and `get_detector_region(DiffractionPatterns)`.

## Test plan
- [x] Full suite passes on MacOS/CPU and Linux/CPU+GPU
- [x] New `test/test_phonon_loss.py`: component consistency, invalid-component handling, thermal weighting (signed axis ordering, zero-bin passthrough, detailed-balance conservation, validation errors), lazy-exit-waves correctness (matches eager, both with and without thermal weighting)
- [x] New tests in `test/test_spectral_detectors.py`: rotated-slit mask area at angles 0–137°, `corners=` conflicting-parameter validation, `show(explode=...)` correctness for ascending/non-ascending orders and `explode=True` (verified against ground truth, not just "does it crash"), collapse-to-member-0 warning, `momentum_resolved_spectrum` with a lazy/fragmented-chunk input matches eager for both detector types, `show()` with GPU-resident (cupy-like) data
- [x] Manual checks: `_annular_detector_mask` with a precomputed `k2` bit-identical to computing it internally; eager-vs-lazy-vs-independent-reference for `EnergyResolvedAtomsEnsemble` (2×2 and 3×2 shapes) after #333; full phonon-loss → momentum-resolved-spectrum → `.show()` pipeline reproduced end-to-end with lazy dask input
- [x] After the final `dev` merge: full suite green; direct smoke tests confirming the hand-resolved conflicts (multi-energy `SMatrix` + upsample validation; multi-energy `Probe.scan` + potential prebuild) work together correctly


TomaSusi:
@pzeiger I'm planning to merge this soonish, did you have a chance to take a look? Just added the missing thermal weighting :)

pzeiger:
Hi Toma,

I will try to have a look today. I wanted to get the multi-GPU things figured out before, since one has to always take a whole node on the new cluster I have access to, so I want to minimize wasting GPU time.

Best,
Paul

TomaSusi:
I found and in about 30 min will fix one more issue, so today's perfect

TomaSusi:
Okay, should be all good now!

pzeiger:
Hi Toma,

can I push additional fixes straight to this branch or should I create a separate branch?

Best,
Paul

TomaSusi:
Feel free to push directly here!

TomaSusi:
After running a bunch of calculations on GPU for my upcoming conference presentations (IMC and Belgrade), and fixing a few minor issues, I'd say this is good to go. Going to merge this now, thank you @pzeiger – let's have a beer to celebrate getting this finally done :)

TomaSusi:
@pzeiger: While checking up on some of the curious features of the simulated graphene spectra we looked at yesterday, I came across something worth double-checking on the normalization side: a probable double application of quantum (Bose-Einstein) statistics between abTEM (which is directly reimplementing the original FRFPMS method) and hiPhive rattled structure generation.

This PR's description documents the intended convention explicitly:

> A single multislice run per energy *magnitude* gives the **classical**, loss/gain-symmetric TDS intensity. Splitting it into the physical loss (+E) and gain (-E) sides is a quantum, temperature-dependent effect (Bose-Einstein detailed balance) — **not something the frozen-phonon sampling itself distinguishes.**

with the detailed-balance split applied downstream, as a pure reweighting of the already-computed classical intensity:

```
I_loss(+E) = I_tds(E) * (n + 1) / (2n + 1)
I_gain(-E) = I_tds(E) *  n      / (2n + 1)
```

This matches your original FRFPMS method too — the δ-thermostat MD heats each frequency band to a fixed *classical* temperature, with no zero-point motion or Bose-Einstein occupation anywhere in the snapshot generation itself.

I've been generating the phonon-based snapshots with `quantum=True`, which doesn't just add a temperature dependence — it assigns each mode's displacement amplitude via the full Bose-Einstein formula *including the zero-point (+1) term*:

```python
np.sqrt(hbar * (2 * n_BE(temperature_eV, hbar * w_s) + 1) / (2 * w_s))
```

If a workflow generates its `EnergyResolvedAtomsEnsemble` snapshots with `quantum=True` and then calls `phonon_loss_diffraction_patterns(..., temperature=...)` on top, quantum statistics get applied twice: once via zero-point/Bose-Einstein-weighted amplitudes at the sampling stage, and again via the detailed-balance loss/gain split, which was designed assuming a classical input. Since Bose-Einstein occupation diverges most strongly at low energy, double-counting it would distort relative intensities most in the already-dominant low-energy region — plausibly significant for real spectra, not just a rounding-level effect.

Flagging this since it came up while I was separately debugging some candidate artifacts in a real graphene phonon-loss simulation built on my rattling + this PR's API, and you mentioned you had some doubts about the normalization. The plots I showed were 0 K so this isn't relevant for the features I am investigating, but is worth addressing.

Probably we should add a flag to abTEM to allow either MD or phonon-based snapshot generation paths to be used correctly, and do the normalization differently based on that?

🤖 Written with the help of [Claude Code](https://claude.com/claude-code)

PR #331, opened 2026-07-13
Opt-in multi-GPU support via dask-cuda

pzeiger:
Hi!

This PR introduces fixes to the multi-GPU support in abTEM. Claude has implemented all of the changes and I have kept a watchful eye. Specifically this PR adds opt-in support for distributing a GPU calculation across **all GPUs on a
node**, using [dask-cuda](https://docs.rapids.ai/api/dask-cuda/stable/) (one worker process per GPU). Single-GPU behavior is unchanged; multi-GPU is entirely opt-in.

## Usage

Automatic — abTEM starts a dask-cuda cluster spanning every visible GPU:
```python
abtem.config.set({"device": "gpu", "dask.multi-gpu": True})

#continue with your abTEM script
```

Bring your own cluster (for RMM pools, UCX, a subset of GPUs) — abTEM detects and uses an active dask-cuda client:
```python
from dask_cuda import LocalCUDACluster
from distributed import Client
client = Client(LocalCUDACluster())
abtem.config.set({"device": "gpu"})

#continue with your abTEM script
```

## What's in it
1. New config flag `dask.multi-gpu` (default false) in `abtem/core/abtem.yaml`. 
2. _compute (array.py) now, on the GPU path, respects an active distributed client instead of always forcing the synchronous scheduler, and auto-starts a dask-cuda cluster when dask.multi-gpu is enabled (and there is >1 GPU, no active client, and no explicit scheduler=). Removes the dead num_workers/threads_per_worker kwargs.
3. backend.ensure_cuda_cluster() — memoizes a LocalCUDACluster client and rebuilds it if a previous one was shut down; clear error if dask-cuda is missing.
4. backend.is_gpu_dask_client() — only single-threaded, GPU-pinned clients (as produced by LocalCUDACluster) are used for CuPy; a plain multi-threaded LocalCluster (e.g. one opened for the dashboard) is not used and abTEM safely falls back to synchronous execution.
5. abtem[all] extra aggregating the optional runtime extras. (GPU packages are intentionally excluded — cupy/dask-cuda wheels are CUDA/RAPIDS-version specific and must match the local toolkit.)

## Testing
1. `test/test_multigpu_logic.py` — mock/CPU-cluster logic tests that run on ordinary CI: client-suitability, ensure_cuda_cluster memoization/liveness/ missing-dependency, and every scheduler-selection branch of _compute.
2. test/test_multigpu.py — real dask-cuda integration tests, marked multigpu and skipped unless there are ≥2 GPUs and dask-cuda (requires_multigpu): cluster is GPU-appropriate, workers pin to distinct GPUs, a compute distributes over all workers (explicit-client and dask.multi-gpu paths), and the HAADF image is invariant to the number of GPUs.
3. Validated on a 4×A40 node (SrTiO₃, 3600 atoms, frozen-phonon HAADF): near-linear scaling — 1/2/4 GPUs = 754.7 / 378.5 / 189.8 s (1.99× / 3.98×, 99% efficiency) — and the resulting image is bitwise identical regardless of GPU count (max relative deviation 0.0).

## Dependencies / caveats
dask-cuda is optional and user-installed (like cupy), matched to the local CUDA/RAPIDS version. Note the current dask-cuda pins a specific dask; on newer dask it may need a --no-deps install until a matching dask-cuda releases.
Speed-up requires the computation to split into ≥ N independent chunks (frozen phonons, scan positions, tilt series, PRISM); a single plane-wave multislice won't benefit.
Docs: abTEM/doc#19

### DIFF
--- a/abtem/array.py
+++ b/abtem/array.py
@@ -45,7 +45,9 @@ from abtem.core.backend import (
     copy_to_device,
     cp,
     device_name_from_array_module,
+    ensure_cuda_cluster,
     get_array_module,
+    is_gpu_dask_client,
 )
 from abtem.core.chunks import Chunks, iterate_chunk_ranges, validate_chunks
 from abtem.core.ensemble import Ensemble, _wrap_with_array, unpack_blockwise_args
@@ -506,14 +508,29 @@ def _compute(
     if is_gpu:
         check_cupy_is_installed()
 
-        if "scheduler" not in kwargs:
+        from distributed import get_client
+
+        try:
+            client = get_client()
+        except ValueError:
+            client = None
+
+        # Start a dask-cuda cluster spanning all visible GPUs when multi-GPU is
+        # enabled, no client is already handling the computation, and the user has
+        # not explicitly requested a scheduler.
+        if (
+            client is None
+            and "scheduler" not in kwargs
+            and config.get("dask.multi-gpu", False)
+            and cp.cuda.runtime.getDeviceCount() > 1
+        ):
+            client = ensure_cuda_cluster()
+
+        # The threaded scheduler and multi-threaded workers cannot be used with CuPy;
+        # fall back to synchronous execution unless a suitable (dask-cuda) client is
+        # handling the computation.
+        if not is_gpu_dask_client(client) and "scheduler" not in kwargs:
             kwargs["scheduler"] = "synchronous"
-
-        if "num_workers" not in kwargs:
-            kwargs["num_workers"] = cp.cuda.runtime.getDeviceCount()
-
-        if "threads_per_worker" not in kwargs:
-            kwargs["threads_per_worker"] = cp.cuda.runtime.getDeviceCount()
 
     with _compute_context(
         progress_bar, profiler=profiler, resource_profiler=resource_profiler
--- a/abtem/core/abtem.yaml
+++ b/abtem/core/abtem.yaml
@@ -18,6 +18,9 @@ dask:
   chunk-size: 128 MB
   # The target chunk size to use for dask arrays on the gpu
   chunk-size-gpu: 512 MB
+  # Automatically distribute gpu computations over all visible gpus by starting a
+  # dask-cuda cluster. Requires the optional dask-cuda package.
+  multi-gpu: false
 cupy:
   # The size of the fft cache in MB used by cupy
   # https://docs.cupy.dev/en/stable/user_guide/fft.html#fft-plan-cache
--- a/abtem/core/backend.py
+++ b/abtem/core/backend.py
@@ -52,6 +52,80 @@ def check_cupy_is_installed():
         raise RuntimeError("CuPy is not installed, GPU calculations disabled")
 
 
+_cuda_cluster_client = None
+
+
+def ensure_cuda_cluster():
+    """
+    Start a dask-cuda cluster spanning all visible GPUs and return its client.
+
+    The cluster assigns one worker process to each GPU, allowing dask to distribute
+    computations across all of them. It is created once per process and reused on
+    subsequent calls. Requires the optional dask-cuda package.
+
+    Returns
+    -------
+    distributed.Client
+        The client connected to the dask-cuda cluster.
+    """
+    global _cuda_cluster_client
+
+    if _cuda_cluster_client is not None:
+        if getattr(_cuda_cluster_client, "status", None) == "running":
+            return _cuda_cluster_client
+        # The previous cluster was shut down; discard it and start a new one.
+        _cuda_cluster_client = None
+
+    try:
+        from dask_cuda import LocalCUDACluster  # type: ignore
+    except ImportError:
+        raise RuntimeError(
+            "The dask-cuda package is required to distribute computations across "
+            "multiple GPUs. Please install it (see "
+            "https://docs.rapids.ai/api/dask-cuda/stable/install/), or set the "
+            "configuration option 'dask.multi-gpu' to false."
+        )
+
+    from distributed import Client
+
+    _cuda_cluster_client = Client(LocalCUDACluster())
+
+    return _cuda_cluster_client
+
+
+def is_gpu_dask_client(client) -> bool:
+    """
+    Check whether a distributed client can safely execute CuPy computations.
+
+    Only a client whose workers are each single-threaded — as produced by
+    ``dask_cuda.LocalCUDACluster``, which additionally pins one GPU per worker — is
+    considered suitable. The threaded scheduler and multi-threaded workers share a
+    single CUDA context per process, which cannot be used with CuPy.
+
+    Parameters
+    ----------
+    client : distributed.Client or None
+        The client to check.
+
+    Returns
+    -------
+    bool
+        True if the client is running and all of its workers are single-threaded.
+    """
+    if client is None:
+        return False
+
+    if getattr(client, "status", None) != "running":
+        return False
+
+    try:
+        nthreads = client.nthreads()
+    except Exception:
+        return False
+
+    return len(nthreads) > 0 and all(n == 1 for n in nthreads.values())
+
+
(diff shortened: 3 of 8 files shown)


TomaSusi:
Nice, thanks a lot @pzeiger!

Have you tested if this works with my GPU memory improvements? 
https://github.com/abTEM/abTEM/pull/269

TomaSusi:
I asked Claude, and there is going to indeed be some changes that need to be made:

> Both PRs edit the exact same block in abtem/array.py's _compute():
> 
> [#331](https://github.com/abTEM/abTEM/pull/331) replaces the old unconditional scheduler="synchronous" / num_workers / threads_per_worker block with client-detection logic (get_client() → ensure_cuda_cluster() → is_gpu_dask_client()).
> [#269](https://github.com/abTEM/abTEM/issues/269) leaves that exact block untouched but renames the is_gpu check to a new _is_gpu_array_object() helper (for its own _device-attribute tracking), and — more importantly — adds a second, separate forced-synchronous block in write_to_directory(), using dask.config.set(scheduler="synchronous") rather than a kwarg.
> Git will flag _compute() as a conflicting hunk (same lines, different rewrites). Mechanically resolvable — attach [#331](https://github.com/abTEM/abTEM/pull/331)'s client/cluster logic to #269's renamed helper — but not a silent auto-merge.
> 
> abtem/core/abtem.yaml also gets adjacent-but-independent new keys from both PRs (dask.multi-gpu vs cupy.fft-cache-size/potential.slice-chunk-size) — trivial to reconcile. pyproject.toml touches different sections in each PR, no real conflict there.
> 
> The deeper problem: they fight over the same invariant
> [#269](https://github.com/abTEM/abTEM/issues/269)'s comment in write_to_directory is explicit about why it forces synchronous execution on GPU:
> 
> "Use synchronous scheduler on GPU to avoid multiple dask tasks running in parallel, each loading potential chunks + wave arrays into VRAM simultaneously."
> 
> Its entire memory model — abtem/core/chunks.py and estimate_scan_batch_size — sizes chunks against "effective free VRAM queried at compute time," under the assumption that only one task touches a given GPU at a time. [#331](https://github.com/abTEM/abTEM/pull/331)'s entire point is to relax exactly that assumption so multiple GPU tasks can run concurrently.
> 
> The good news: this isn't fundamentally incompatible. dask-cuda's LocalCUDACluster pins one single-threaded worker process per physical GPU, so each worker still has effectively-exclusive access to its own GPU — [#269](https://github.com/abTEM/abTEM/issues/269)'s per-worker VRAM budget query would still be accurate inside a dask-cuda worker. The concerning gap is #269's own comment about nested dask.compute() calls: it forces the scheduler via dask.config.set() (a global override) specifically because an inner array.compute() call, with no explicit scheduler, would otherwise fall back to dask's local threaded scheduler even while a distributed client is active — spawning multiple threads hammering one GPU concurrently within a single worker process. [#331](https://github.com/abTEM/abTEM/pull/331) only patches the top-level _compute() entry point; it doesn't touch write_to_directory's inner compute call or audit other nested-compute sites. If [#331](https://github.com/abTEM/abTEM/pull/331) lands as-is on top of [#269](https://github.com/abTEM/abTEM/issues/269), enabling dask.multi-gpu wouldn't reintroduce cross-GPU thrashing, but it could reopen exactly the single-GPU multi-thread VRAM-thrashing bug [#269](https://github.com/abTEM/abTEM/issues/269) fixed, for any nested compute call that isn't covered by [#331](https://github.com/abTEM/abTEM/pull/331)'s client check.
> 
> What reconciling actually needs
> Not a mechanical merge — a design decision: keep [#331](https://github.com/abTEM/abTEM/pull/331)'s outer scheduler selection (per-GPU-cluster distribution), but make the "avoid concurrent tasks per GPU" guarantee scoped to within a single worker/GPU rather than global. Concretely: audit every nested dask.compute() call (starting with write_to_directory) so it still can't fall back to a local threaded scheduler inside a dask-cuda worker, even when the outer graph is legitimately spread across multiple GPU workers.

Would it make sense to rebase 269 on 331 to implement the multi-GPU pathway in 269?

> I'd actually go the other direction — land [#269](https://github.com/abTEM/abTEM/issues/269) first, then rebase [#331](https://github.com/abTEM/abTEM/pull/331) onto the post-[#269](https://github.com/abTEM/abTEM/issues/269) dev, rather than rebasing #269 onto [#331](https://github.com/abTEM/abTEM/pull/331).
> 
> The reason: [#331](https://github.com/abTEM/abTEM/pull/331) is the one relaxing a constraint #269 establishes (single-task-per-GPU exclusivity, enforced not just in _compute() but also via the defensive dask.config.set(scheduler="synchronous") in write_to_directory that guards against nested computes). Whoever does the rebase inherits the job of walking through every place that constraint is enforced and deciding whether it should become "unless there's a valid per-GPU dask-cuda client" instead of a blanket force. It's more natural for [#331](https://github.com/abTEM/abTEM/pull/331)'s author to do that audit against #269's already-landed code than to have #269's memory-chunking logic try to retrofit onto a scheduler that's already been loosened out from under it.
> 
> Two things to flag regardless of direction:
> 
> Rebase order only avoids the git-mechanical conflict, not the semantic work. Someone still has to open write_to_directory (which [#331](https://github.com/abTEM/abTEM/pull/331) currently doesn't touch at all) and decide if per-GPU exclusivity still holds there under dask-cuda's one-process-per-GPU model. That's manual reasoning, not something a rebase resolves automatically.
> Stacking one open PR branch directly on another's branch balloons the diff — [#331](https://github.com/abTEM/abTEM/pull/331) rebased onto #269's branch would show all of [#269](https://github.com/abTEM/abTEM/issues/269)'s 4671 additions as part of [#331](https://github.com/abTEM/abTEM/pull/331)'s PR until #269 actually merges. Cleaner to merge [#269](https://github.com/abTEM/abTEM/issues/269) into dev first, then rebase [#331](https://github.com/abTEM/abTEM/pull/331) onto the updated dev, so each PR's diff stays scoped to its own changes.
> Given #269 has already been extensively GPU-validated and kept in sync with dev (per my notes from 2026-07-05), it's also the more merge-ready of the two — another reason to sequence it first.

TomaSusi:
The GPU memory and performance improvements are VERY significant, so I would agree with Claude that we should merge those in first, and then you'd need to check that the multi-GPU pathway works with them. Basically I think it should be easy enough: we distribute the trivially parallelizable workload over multiple GPUs, and the automatic chunking (now including the potential) within each GPU works as in #269.

pzeiger:
Hi Toma!

Thanks for the feedback. I will work with Claude to rebase this on your GPU changes, which are significant and very important for larger scale calculations. And I will also remove the "all" target ;). I did not know that uv has it by default.

Best,
Paul

TomaSusi:
Thank you! I actually reversed my position about the [all], it's the only way to achieve this with pip and we do wish to continue supporting it. So that can stay :)

pzeiger:
I have asked Claude to port the multi-GPU changes here to the branch [`potential_chunking`](https://github.com/abTEM/abTEM/tree/potential_chunking) which is the base of PR #269. The result can be found in a new branch [`multi-gpu-on-269`](https://github.com/abTEM/abTEM/tree/multi-gpu-on-269). I have verified that the computation still runs and speeds up as reported if it is based on the memory improvements delivered by PR #269.

@TomaSusi: what do you think is the best course of action? Should I merge everything into the `potential_chunking` branch and update PR #269 or will you merge those changes first and then we will update this PR?

TomaSusi:
Claude recommends that we merge multi-gpu-on-269 into potential_chunking and fold it into PR [#269](https://github.com/abTEM/abTEM/issues/269). Let me know if you can do that, or if I should take care of it :)

pzeiger:
Sure, I can do that.

pzeiger:
The content of this PR has been merged into the existing PR #269.

PR #334, opened 2026-07-22
Add BiP-PRISM beam partitioning to PRISM-EELS

PhilippPelz:
## Summary

Adds the **BiP-PRISM** (bi-partitioned PRISM) algorithm for core-loss STEM-EELS on top of the PRISM-EELS beam-basis driver, following *"The BiP-PRISM algorithm for fast and scalable core-loss STEM-EELS simulations"* (Algorithm 5).

The dual-scattering-matrix beam-basis driver already computes the exact map `I(ρ) = Σ_q |a_q(ρ)|²`, but builds the detector matrix `S₂` over the full reciprocal grid (`O(G²)`). BiP-PRISM instead builds **both** the probe-forming matrix `S₁` and the detector matrix `S₂` on a sparse set of hex-ring **parent beams** and reconstructs the full beam sets locally at each ionized atom by windowed natural-neighbor interpolation with magnitude preservation. Resident matrix construction and memory then scale with the parent count rather than the full beam / detector set, which is what makes full-resolution maps and momentum-resolved qEELS tractable on consumer GPUs.

Based on top of #289 (PRISM-EELS MVP); the diff here is only the BiP-PRISM addition.

## What's added

**New module `abtem/prism/_bipartite.py`:**
- `select_parent_beams` — hex-ring parent selection (DC + `n_radial` rings, snapped to the nearest beam, deduplicated; indices into the beam set so the full-parent limit is exact).
- `natural_neighbor_weights` — vectorised Delaunay-barycentric partition-of-unity weights (exact at the parents, nearest-fallback outside the hull).
- `windowed_reconstruct` — per-atom reconstruction on a crop window: de-tilt parents → NNW combine → magnitude-preserve → re-tilt.
- `focal_backprop_distance` / `pad_window` / `fresnel_margin` — S₁ focal back-propagation helpers.

**`prism_transition_potential_scan_beam_basis` (`abtem/inelastic/core_loss.py`) gains:**
- `partitions_s1` / `partitions_s2` / `n_angular` — build `S₁` / `S₂` on parent beams.
- `mag_preserve` — magnitude-preserving reconstruction that recovers the absolute intensity the convex complex average would otherwise lose to decoherence.
- `collection_angle` — restrict `S₂` to the detector-collection disk (the map regime): leaves the detector-integrated map unchanged while making the `S₂` partition accurate at low parent counts and cheaper to build.
- `focal_backprop` — S₁-only accuracy lever that reconstructs at the scattering-centroid plane before propagating to the ionization plane.

With every new argument left at its default, the driver reproduces the previous exact beam-basis result **bit-for-bit**.

**`SMatrix.transition_potential_scan`** exposes the feature via a `reduction` switch (`"real_space"` | `"beam_basis"` | `"bipartite"`) plus the partition / collection / focal kwargs; the eager, lazy and frozen-phonon paths are wired, and passing any partition kwarg auto-selects the beam-basis backend.

```python
measurement = s_matrix.transition_potential_scan(
    transition_potential,
    scan=scan,
    detectors=AnnularDetector(inner=0, outer=theta),
    double_channel=True,
    collection_angle=theta + 5,   # limit S2 to the detector disk
    partitions_s1=4,              # hex-ring parents for S1
    partitions_s2=4,              # hex-ring parents for S2
    focal_backprop="centroid",    # optional S1 accuracy lever
)
```

## Validation

New `test/test_bipprism.py` (23 tests) covers:
- the geometry / reconstruction helpers (full-parent identity, partition of unity, phase-convention roundtrip);
- **no-partition limit is bit-exact** vs the existing beam-basis path (already gated against `Probe.transition_potential_scan` multislice at `interpolation=1`);
- partitioned results **converge monotonically to the exact map** as the parent count grows (Alg. 5 full-parent limit);
- detector-disk `S₂` partitioning reaches **Pearson > 0.99 at `n_radial = 4`**, and limiting `S₂` to a disk that covers the detector leaves the map unchanged;
- **magnitude-preserving** reconstruction recovers absolute scale that the plain complex average loses (the loss grows with thickness);
- **focal back-propagation** reduces the S₁ reconstruction error (~3× at 6 slices, growing with thickness);
- public-API dispatch and eager/lazy equivalence.

The existing `test/test_ionization.py` suite is unchanged and continues to pass (no regressions).

## Notes

- Follow-ups not included here: explicit qEELS / 4D-cube output modes, the peel-once `S₂` + fold-weights-into-contraction optimisation (the MVP reconstructs the full columns on the crop window, which proves correctness but does not yet realise the full `O((B_{p,1}+B_{p,2})·G)` memory scaling), and a numerical cross-check against the reference implementation.
- Tested on CPU (Python 3.11, numpy 1.26, zarr 3.0.8). The GPU path follows the existing `xp`/backend abstractions but was not exercised here.

### DIFF
--- a/abtem/inelastic/core_loss.py
+++ b/abtem/inelastic/core_loss.py
@@ -2068,6 +2068,7 @@ def prism_transition_potential_scan_beam_basis(
             )
 
     # --- Main loop ---
+    warned_empty_mask = False
     for slice_index, transmission in enumerate(transmissions):
         s_waves = _step(s_waves, transmission)
 
@@ -2119,6 +2120,44 @@ def prism_transition_potential_scan_beam_basis(
             site_xy = np.array(
                 [atom.position[0], atom.position[1]], dtype=np.float32
             )
+
+            # Which scan positions fall inside this atom's PRISM window
+            # (only depends on site_xy, not on any of the S1/S2 reconstruction
+            # below) -- compute and check it FIRST so a scan too coarse for
+            # this atom (no position lands inside its window, so it would
+            # contribute exactly zero) skips the reconstruction work entirely
+            # instead of building empty-batch arrays that some backends (e.g.
+            # CuPy's FFT) reject.
+            mask = xp.ones(n_positions, dtype=bool)
+            if interpolation[0] > 1:
+                mask &= (
+                    xp.abs(positions[:, 0] - site_xy[0])
+                    % (extent[0] - prism_region[0])
+                ) <= prism_region[0]
+            if interpolation[1] > 1:
+                mask &= (
+                    xp.abs(positions[:, 1] - site_xy[1])
+                    % (extent[1] - prism_region[1])
+                ) <= prism_region[1]
+
+            if not bool(mask.any()):
+                if not warned_empty_mask:
+                    warnings.warn(
+                        "PRISM-EELS beam-basis: at least one ionization site "
+                        "has no scan position within its PRISM window "
+                        "(extent / interpolation, cropped to half that per "
+                        "axis) -- the scan is too coarse relative to "
+                        "`interpolation` for that site to contribute "
+                        "anything. Such sites are skipped (equivalent to a "
+                        "zero contribution); use a denser scan or lower "
+                        "`interpolation` if this is unintended.",
+                        stacklevel=2,
+                    )
+                    warned_empty_mask = True
+                continue
+
+            coeff_masked = coefficients[mask]  # (n_masked, n_k)
+
             site_pixel = site_xy / full_sampling_arr
             site_pixel_int = np.rint(site_pixel).astype(int)
             crop_corner = (
@@ -2167,20 +2206,6 @@ def prism_transition_potential_scan_beam_basis(
             else:
                 s1_crop = wrapped_crop_2d(s_waves.array, crop_corner, window_gpts)
             HS1 = H_crop[:, None] * s1_crop[None, :]  # (n_T, n_k, wh, ww)
-
-            mask = xp.ones(n_positions, dtype=bool)
-            if interpolation[0] > 1:
-                mask &= (
-                    xp.abs(positions[:, 0] - site_xy[0])
-                    % (extent[0] - prism_region[0])
-                ) <= prism_region[0]
-            if interpolation[1] > 1:
-                mask &= (
-                    xp.abs(positions[:, 1] - site_xy[1])
-                    % (extent[1] - prism_region[1])
-                ) <= prism_region[1]
-
-            coeff_masked = coefficients[mask]  # (n_masked, n_k)
 
             if double_channel:
                 if partitions_s2 is not None:
--- a/abtem/inelastic/core_loss.py
+++ b/abtem/inelastic/core_loss.py
@@ -1701,6 +1701,12 @@ def prism_transition_potential_scan_beam_basis(
     sites=None,
     double_channel: bool = True,
     inelastic_crop: float | tuple[float, float] | None = None,
+    partitions_s1: int | None = None,
+    partitions_s2: int | None = None,
+    n_angular: int = 6,
+    mag_preserve: bool = True,
+    collection_angle: float | None = None,
+    focal_backprop: str | float | None = None,
 ):
     """PRISM-EELS beam-basis reduction (Brown et al. Sec. IV B / Eq. dropped
     in supplementary; ``PRISM_double_channeling_nanoparticle.m``) — the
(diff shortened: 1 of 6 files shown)


TomaSusi:
Thank you! A couple of very minor comments:

1. In the new _bipartite.py _window_tilt helper, get_dtype() is imported and used for real_dtype, but dy/dx and π are still hardcoded to np.float32 right next to it (should respect config['precision'] throughout).
2. Two docstring lines (194 in pad_window and 220 in _window_tilt) exceed the repo’s 88-char ruff line-length limit in the same file (not enforced by CI here, but would fail lint if you add it later).

TomaSusi:
I've committed a benchmark script (`bench_bipprism_eels.py`) that compares the different available algorithms for core-loss simulations in abTEM: 

- original multislice approach (Dwyer 2008)
- original PRISM-EELS method (Brown 2019; implemented as prism_transition_potential_scan and dubbed "real-space" since for each atom it forms H(r) · S1(r) on a real-space crop window)
- beam-basis reduction implemented in https://github.com/abTEM/abTEM/pull/289
- new bipartite method implemented here

I ran the benchmarks on GPU on an RTX4090, with the results as follows (single-channel first, double-channeling second):

image
image

Double-channel thickness is the standout. At 1000 Å, bipartite hits ~141x speedup over multislice, real_space ~124x, beam_basis ~103x — and multislice itself needs 6780 seconds (1.88 hours) for that one data point (nz=256, 4096 positions), scaling as a clean O(N²) (the ratio compounds toward almost exactly 4x per doubling of nz — expected, since double-channel must propagate every ionization event to the exit, and atom count also grows linearly with thickness). This is genuinely the case for BiP-PRISM's double-channel utility.

The cell-size sweep gives you a clean failure hierarchy. In double-channel: real_space OOMs at sc=8, beam_basis at sc=12, bipartite survives to sc=16 — a direct, uncluttered demonstration that BiP-PRISM's memory scaling advantage is real and holds up further than either predecessor. At sc=16 only multislice completes at all (8864s — 2.5 hours), so there's correctly no speedup data point there since every PRISM variant is NaN.

A pleasant surprise: bipartite isn't just winning on memory anymore — it's starting to win on time too, at extremes. At nz=256 in the thickness sweep, bipartite (49.9s) edges out beam_basis (65.0s) and even real_space (54.7s) — the reconstruction overhead becomes proportionally smaller as the per-atom exit-propagation cost (shared by all variants) dominates at depth.

To me this all looks reasonable and good to be merged, but perhaps it would be nice to rerun the benchmarks on another system.

TomaSusi:
Ah sorry @PhilippPelz, I merged the underlying claude/prism-eels-mvp branch to dev, and deleted that branch without thinking how it would affect this. I rebased this branch on dev, but then obviously couldn't push it since it's your repo.

If you run these, that should rebase your changes on dev:

```
git fetch https://github.com/abTEM/abTEM.git claude/bip-prism-eels-rebased
git push origin FETCH_HEAD:bip-prism-eels --force
```

PhilippPelz:
No problem @TomaSusi — but heads up that deleting `claude/prism-eels-mvp` didn't just
change the base, it **auto-closed this PR**, and GitHub won't let it be revived:
it refuses to retarget the base of a closed PR, and refuses to reopen a PR whose base
branch has been deleted. So I've opened **#338** with the rebased work instead.

The rebase is done and verified. Since #289 was *squash*-merged, the old mvp commits
aren't ancestors of `dev`, so a plain `git rebase dev` would try to replay #289 on top
of itself; the correct form is `git rebase --onto origin/dev ba4264f9`, replaying only
the four BiP-PRISM commits. `git range-diff` against the pre-rebase branch shows
commits 1, 2 and 4 patch-identical, with commit 3 differing only in `import` context in
`test/test_ionization.py`. That matches your `claude/bip-prism-eels-rebased` branch.
On the rebased branch `test/test_ionization.py` gives 25 passed / 1 skipped on CPU, and
the `interpolation=1` validation still reproduces conventional multislice on both
channels.

If you'd rather keep the discussion here than move to #338, this PR *can* be recovered —
but only by someone with write access to `abTEM/abTEM` (I only have `pull`). The old tip
`0b064a43` still exists, so:

```
git push origin 0b064a43:refs/heads/claude/prism-eels-mvp   # resurrect the base
gh api -X PATCH repos/abTEM/abTEM/pulls/334 -f state=open   # reopen
gh api -X PATCH repos/abTEM/abTEM/pulls/334 -f base=dev     # retarget
git push origin --delete claude/prism-eels-mvp              # clean up
```

Order matters: reopen *before* retargeting, since both operations are blocked in the
other order. If you do that, close #338 and I'll move the benchmark comment back here.

Also: thanks for running those benchmarks. I'm cross-checking them on an RTX A6000
(48 GB) and will post the numbers on #338. One thing to expect — your cell-size failure
hierarchy is partly a 24 GB artifact. Your last *successful* points were sc=4
(`real_space`), sc=8 (`beam_basis`) and sc=12 (`bipartite`, ~10.5 GB peak), all of which
fit comfortably in 48 GB, so the ladder should shift outward and the result becomes a
measured memory ratio rather than an OOM ordering.

TomaSusi:
I've added you as a collaborator with write access, so you should be able to push now :) Sorry for the hassle!

PR #338, opened 2026-07-29
Add BiP-PRISM beam partitioning to PRISM-EELS

PhilippPelz:
Adds BiP-PRISM for core-loss STEM-EELS: build S1 and S2 on sparse hex-ring parent beams, then reconstruct their target columns using natural-neighbor interpolation with magnitude preservation. This supersedes #334; its review history and RTX 4090 benchmark discussion remain there.

The public entry point is `SMatrix.transition_potential_scan(..., reduction="bipartite")`, with `partitions_s1`, `partitions_s2` and an optional detector collection angle. The unpartitioned beam-basis path remains available.

```python
measurement = s_matrix.transition_potential_scan(
    transition_potential,
    scan=scan,
    detectors=AnnularDetector(inner=0, outer=theta),
    reduction="bipartite",
    double_channel=True,
    collection_angle=theta + 5,
    partitions_s1=4,
    partitions_s2=4,
)
```

Zero-defocus reconstruction is the default. Optional focal back-propagation remains available; its cropped propagation is approximate and is bypassed when all S1 parents are retained.

The implementation includes the reviewed precision, phase and normalization corrections: preserve configured S1 complex precision; remove/restore the adjoint S2 vacuum phase while retaining antialias support; normalize double-channel amplitudes to the PRISM cell; preserve detection-cell normalization for smaller single-channel crops; and center odd-sized transition crops correctly.

The performance update reuses backend geometry and reconstructs S1/S2 once per occupied slice when active crop coverage and a joint workspace budget permit it. Active focal propagation, insufficient coverage or memory retain window reconstruction. The default additional reconstruction-workspace estimate is capped at 2 GiB; this is not a total-memory cap. CPU requires greater crop coverage than one grid, while GPU also permits equal coverage. Double-channel contraction selects beam-first or probe-first ordering from the actual dimensions, avoiding an all-transition HS1 temporary.

Complete native GPU scans compared with the previous PR head `153ef7dd` on an RTX A6000:

| Grid / slices / scan | f | Before | After | Speedup |
|---|---:|---:|---:|---:|
| 128² / 4 / 16² | 1 | 1.778 s | 0.815 s | 2.18× |
| 128² / 4 / 16² | 4 | 0.574 s | 0.363 s | 1.58× |
| 256² / 4 / 16² | 4 | 3.979 s | 3.302 s | 1.21× |
| 128² / 8 / 16² | 4 | 1.178 s | 0.744 s | 1.58× |
| 128² / 4 / 4² | 4 | 0.593 s | 0.391 s | 1.52× |
| 128² / 4 / 32² | 1 | 1.902 s | 1.279 s | 1.49× |

These are medians of three alternating synchronized calls after warmup, with fresh S matrices and identical prepared inputs. Python 3.12.10, NumPy 1.26.4, CuPy 13.4.1; two synthetic transition channels, 25 mrad collection and 20 mrad annular detection. Single-precision image relative L2 differences were at most 4.09e-7. Two additional double-precision GPU comparisons differed by less than 3.6e-16. The small CPU f=1 case improved 3.31×; f=2 was approximately unchanged. This shared-workstation benchmark does not establish universal speedups.

Memory is a tradeoff: the 128² f=4 case's incremental CuPy allocation peak increased from 130 to 597 MiB. The f=1 case decreased from 1269 to 1038 MiB. These measurements exclude prepared inputs and do not measure all CUDA/library allocations.

Regression coverage includes both backends and precisions, wrapped anisotropic crops, individual/both partitioned legs, single/double channel, magnitude replacement, focal/full-parent fallback, S2 vacuum phases and zero-thickness support, noncontributing sites, mixed cache eligibility, cache lifetime, forced contraction orders and empty batches. Final local validation: **351 passed, 178 skipped** across the BiP-PRISM, reuse, contraction, ionization, core-loss-fix, PRISM and detector suites. Lint checks passed for the helpers/new tests; `git diff --check` passed. The regression run used `PYTHONUTF8=1`: without it this environment reproduces an order-dependent ASCII source-decoding failure on the unchanged PR head after ionization tests. No source-encoding workaround is included in the patch.

[Upstream full-suite CI](https://github.com/abTEM/abTEM/actions/runs/34379812771) passed on Python 3.11, 3.12, 3.13 and 3.14 at `daf361c8`.


PhilippPelz:
## Benchmark cross-check on an RTX A6000 (48 GB)

Re-ran @TomaSusi's `bench_bipprism_eels.py` from #334 on an RTX A6000 (48 GB, driver
580.159.04) against this rebased branch (`f608f30a`). Same parameters as the 4090 run
— `sc=4, interp=4, partitions_s1=4, partitions_s2=4, collection_angle=25.0 mrad`,
200 keV, 25 mrad aperture, 20 mrad detector — and the script itself is byte-identical
across the rebase, so the two runs are directly comparable. The `interpolation=1`
validation passed for both channels before timing (`real_space`, `beam_basis`, and
`bipartite` at the full-parent limit all reproduce conventional multislice).

**One deliberate deviation:** the multislice baseline is cost-capped at `nz<=64` and
`sc<=8`, because on this card the four expensive baseline points would have cost
~15 h on their own (every timed call is preceded by an untimed warm-up, so each is
paid twice). The PRISM variants are **uncapped**. So `nan` for multislice at nz=128/256
and sc=12/16 below is *our cap, not a failure* — multislice would have completed those
given the time. Total run: 3 h 44 m.

### Double-channel, thickness — bipartite wins on time at 1000 Å

| nz | thickness | multislice | real_space | beam_basis | bipartite |
|---|---|---|---|---|---|
| 16 | 62.5 Å | 38.732 s / 4609 MB | 3.464 s / 3351 MB | **1.373 s** / 1186 MB | 2.084 s / 832 MB |
| 64 | 250 Å | 600.808 s / 4609 MB | 16.141 s / 3357 MB | **11.500 s** / 1192 MB | 16.770 s / 838 MB |
| 128 | 500 Å | capped | 37.862 s / 3365 MB | **31.461 s** / 1200 MB | 32.117 s / 846 MB |
| 256 | 1000 Å | capped | 76.127 s / 3381 MB | 77.744 s / 1216 MB | **45.839 s / 862 MB** |

This **reproduces your crossover, with a wider margin**. You had bipartite at 49.9 s
edging beam_basis (65.0 s) and real_space (54.7 s); here bipartite is 1.70× faster than
beam_basis and 1.66× faster than real_space, while also using the least memory. The
mechanism is visible in the scaling: from 500→1000 Å, bipartite grows ×1.43 against
×2.0 (real_space) and ×2.5 (beam_basis).

Multislice's O(N²) behaviour reproduces exactly — ×3.7, ×3.88, ×3.92, ×3.96 per
doubling across the measured points, against ×2.0 for every reduction. Extrapolating
that trend from the last measured baseline (600.808 s at nz=64) gives ~9400 s at
1000 Å, implying **~205× for bipartite** and ~120× for the other two. Flagging clearly
that this is *inferred*, unlike your ~141× which was measured against a real 6780 s
baseline — the numbers don't conflict, they're a different card and a faster bipartite.

### Double-channel, cell size — bipartite is the only PRISM variant that runs at sc≥12

| sc | gpts | multislice | real_space | beam_basis | bipartite |
|---|---|---|---|---|---|
| 8 | 256² | 1094.190 s / 18435 MB | 103.475 s / 15366 MB | 28.565 s / 11813 MB | 31.402 s / **3580 MB** |
| 12 | 384² | capped | **OOM** | **OOM** | 258.467 s / 9735 MB |
| 16 | 512² | capped | **OOM** | **OOM** | 1520.645 s / 22760 MB |

At sc=8, bipartite uses **3.3× less memory** than exact beam_basis (3580 vs 11813 MB) at
comparable wall-clock. At sc=12 both predecessors fail — real_space asked for 14.7 GB on
top of 31.8 GB, beam_basis 8.1 GB on top of 38.5 GB, i.e. ~43 GiB each — while bipartite
completed in 9.7 GB, a **≥4.5× advantage** that is the difference between running and not
running. sc=16 at 22760 MB is a data point no variant reached on your 4090.

**The one result I did not expect:** `beam_basis`'s double-channel ceiling is **sc=8 on
both cards** — doubling the VRAM bought it nothing. Its requirement jumps from 11.8 GB at
sc=8 to >46 GB at sc=12, a ~4× step that leapfrogs the 24→48 GB gap entirely. So its
limit isn't "24 GB was too small", it's that the S₂ build outgrows VRAM generations.
`real_space` did gain exactly one rung (sc=4 → sc=8). That makes the case for
partitioning S₂ stronger than the 4090 data could show, since there the same wall reads
as a mere capacity shortfall.

### Measured double-channel speedups vs multislice

| sweep | point | real_space | beam_basis | bipartite |
|---|---|---|---|---|
| scan positions | 65536 | OOM | **97.2×** | 87.7× |
| thickness | 250 Å | 37.2× | **52.2×** | 35.8× |
| cell size | sc=8 | 10.6× | **38.3×** | 34.8× |

### Single-channel: bipartite trades time for memory

Single-channel is much less favourable to bipartite, and worth stating plainly: at sc=16
beam_basis is 36.508 s / 5995 MB against bipartite's 58.010 s / 3545 MB — 1.7× less
memory for 1.6× more time. Speedups plateau near 4.0× (beam_basis) and 2.7× (bipartite)
in the thickness sweep. The bipartite/beam_basis memory ratio is monotone in cell size
(1.21× → 0.82× → 0.65× → 0.59× at sc=4/8/12/16), i.e. bipartite starts *behind* at small
cells where the fixed parent count is pure overhead. This is consistent with the
mechanism — single-channel doesn't force the full S₂ build, so there's much less for the
S₂ partition to save.

Two other single-channel observations:
- `real_space` becomes **slower than multislice** at sc=8 (101.259 s vs 60.273 s) and
  pathological at sc=12 (973.486 s, 31797 MB). Neither point existed on the 4090 — it
  OOM'd first — so the real-space driver stops being a win at large cells on a big card.
- `real_space` OOM'd at 65536 scan positions in both channels (26.8 GB on top of 31.1 GB).

### Notes on methodology

- **Reproducibility**: sweep 2's nz=16 row and sweep 3's sc=4 row are the same physical
  configuration and agree to within 1–5% in both channels, so the variant differences
  above are well clear of run-to-run scatter.
- **Speedup ratios are not hardware-portable.** At 4096 positions double-channel I get
  27.8× for beam_basis where your figure reads ~43×, but at 65536 I get 97× where yours
  reads ~68×. The ordering and the memory scaling carry across; a single headline
  multiplier does not. Could you share the raw tables or `.npz` from your run? Reading
  values off a log-scale plot is the weakest link in this comparison and I'd rather
  compare exact numbers.
- Multislice peak memory scales with scan size here (up to 36857 MB at 65536 positions)
  whereas your figure shows it flat near 4.4 GB. That's most likely
  `dask.chunk-size-gpu` auto-scaling `max_batch` to available VRAM, which would mean the
  multislice memory curve isn't comparable between cards.

Figures and raw `.npz` for both channels are saved locally; happy to attach them here if
useful.

TomaSusi:
Thank you! Do these match what you got from your initial custom implementation, ie. are we getting similar performance in abTEM?

PhilippPelz:
On a small test (128x128) scatterem is ~20% faster. I have not profiled it deeply.

TomaSusi:
> On a small test (128x128) scatterem is ~20% faster. I have not profiled it deeply.

Thanks for checking! Might be worth asking Claude if there are obvious optimizations we could easily bring over, or whether this is solely due to torch.

Note also that @pzeiger is now intensively working on improving the abTEM side of things, and especially on multi-GPU and ROCm architectures, he's actively finding important improvements.

TomaSusi:
Reviewed the 9 commits added since the rebase (`f608f30a` → `daf361c8`). All correctness/precision fixes plus one performance rework — summary:

- **Precision**: `_window_tilt` no longer hardcodes `float32` for the phase ramp; now respects `config['precision']`.
- **Crop centering**: fixed off-center odd-sized transition-potential crops (`-inelastic_window_gpts[0] // 2` → `-(inelastic_window_gpts[0] // 2)`) — verified by hand this is a real fix, not cosmetic (floor-division on the negated value vs. negating the floor division differ for odd sizes).
- **S2 vacuum-phase handling**: de-phases parent columns before NNW interpolation, restores phase after, and zeroes reconstructed columns outside the antialias support for zero-thickness-adjacent slices. Covered by dedicated tests that independently recompute the expected phase/support.
- **Detection normalization**: single-channel embeds the TP crop into a cell-sized field before FFT; double-channel gets a `sqrt(N_cell/N_native)` correction. Both now converge `real_space`/`beam_basis`/`bipartite` to matching total intensity (verified across 4 interpolation settings × 4 crop sizes).
- **Performance rework**: `_reconstruct_slice` batches per-atom NNW reconstruction once per slice under a memory budget; `contract_eels` picks beam-first vs. probe-first contraction order via an arithmetic cost estimate — checked the formula against its own test cases, internally consistent.

Test coverage: two new files (~1000 lines) covering cache correctness, budget fallback, full-parent exactness for focal back-prop, GPU/CPU parity, and precision propagation. CI green on 3.11–3.14. No correctness issues found — looks ready to merge.

TomaSusi:
Shall we wait to merge this until the article is accepted @PhilippPelz, or you reckon this is it for now?

Issue #19, opened 2021-03-01
multislice hangs when used with multiprocessing on Unix


schiotz:
Hi Jacob,

We try to generate a lot of images in parallel, using the multiprocessing module.  If the PlaneWave.multislice() method is first called directly, and later using multiprocessing, the child processes hang.  It happens on Unix, where the default method for starting new processes is the 'fork' method.  Changing to the 'spawn' method (default on Windows) or the 'forkserver' method prevents the problem from occurring.

It could be some kind of shared global resource that is causing the problem.  With the fork method, the child processes inherits the used resource from the parent process, whereas with the spawn method a new Python process is created, and abtem is loaded again.

The attached script shows the problem.

CC: @matthewhelmi

schiotz:
Hmm, cannot attach a .py, compressing as .zip

[bugexample.zip](https://github.com/jacobjma/abTEM/files/6062955/bugexample.zip)

schiotz:
For viewing:

```
from ase.cluster.cubic import FaceCenteredCubic
import multiprocessing as mp
import numpy as np

# Imports from abTEM
from abtem.potentials import Potential
from abtem.waves import PlaneWave
from abtem.transfer import CTF
from abtem.noise import poisson_noise
from abtem.measure import Measurement, Calibration
from abtem.mtf import MTF



def get_image(atoms):
    
    """
    Generate an image and the corresponding image wave
    """
    
    ## ---- Image and microscope parameters ----
    dose = 1e4
    v0 = 300e3
    defocus = -50
    Cs = 5e4
    focal_spread = 30
    sampling = 0.05
    
    ## ---- Calculating the grid size ----
    size = np.max([int(atoms.cell[0,0]),int(atoms.cell[1,1])])
    num_pts = int(size/sampling*2)
    
    ## ---- Calculating the image wave ----
    wave = PlaneWave(sampling = sampling, energy=v0)
    potential = Potential(atoms, sampling=sampling, gpts=num_pts, projection = 'infinite',parametrization='kirkland')
    exit_wave = wave.multislice(potential, pbar = False)
    #ctf = CTF(energy = v0, defocus=defocus,Cs=Cs,focal_spread=focal_spread)
    #img_wave = exit_wave.apply_ctf(ctf)
    
    #measurement_raw = img_wave.intensity()
    #measurement_noise = poisson_noise(measurement_raw, dose)

    #return measurement_noise

def makeimages(maxcpu=5):
[...]

PR #346, opened 2026-08-21
Multi gpu hardening

pzeiger:
Hi!

While running a large HAADF simulation on Perlmutter with `dask.multi-gpu` enabled (introduced in PR #269), a 40 GB A100 OOM'ed after 24 seconds while three other GPUs sat idle. It turned out that saving a lazy result with `to_zarr()` never started the multi-GPU cluster. Instead the flag is silently ignored on that path and the whole scan runs serially on one GPU. A `compute()` before `to_zarr()` correctly used multiple GPUs and allowed the calculation to complete, but the user would expect that the `compute()` is not necessary in this case, since it is not necessary on single GPUs or CPUs. Claude and I traced this and a series of related problems, which this PR fixes, one commit per issue.

The overarching theme is making the default `to_zarr` path work, and making abTEM say what it is doing when it cannot:

- `to_zarr()` now distributes across GPUs exactly like `.compute()`
- the client's configuration now actually reaches the workers — previously `abtem.config.set` never did, so e.g. a float64 computation dispatched to workers silently ran in float32. This affects the existing compute path too, and is probably the most important fix in here
- more robust automatic VRAM handling: a more conservative batch estimate on FFT-unfriendly grids, a device-relative bound on the cuFFT plan cache (with a safe fallback for single plans bigger than the bound), and removal of a redundant copy of every wave batch in multislice
- warnings whenever abTEM falls back to something you did not ask for (multi-GPU requested but not used, a missing `__main__` guard, FFT-unfriendly grid sizes), plus config keys for an RMM pool and GPU subsets.

Everything is benchmarked in the description below: the workload that crashed in 24 seconds now completes in 13 minutes on 4 GPUs (or ~50 minutes on a single GPU), results are bit-identical between branches and consistent between execution modes, and there are no measured speed regressions.

Best,
Paul

---------------
# Claude's summary
Follow-up to #269's opt-in multi-GPU support. A production HAADF scan on a Perlmutter 4×A100 node crashed with `dask.multi-gpu: true`: saving the lazy result with `to_zarr()` silently ignored the flag, ran the entire scan on one GPU, and OOM'd it. This PR fixes that and everything else found while chasing it — one commit per issue, every claim benchmarked.

<details>
<summary><b>The original failure log</b> (Perlmutter, 4×A100 node — trimmed to the relevant frames)</summary>

The script follows the natural pattern — enable multi-GPU, build a scan, save the result:

```python
abtem.config.set({"device": "gpu", "precision": "float64", "dask.multi-gpu": True})

potential = abtem.Potential(atoms, sampling=0.05, slice_thickness=atoms.cell[2, 2] / 4)
probe = abtem.Probe(energy=60e3, semiangle_cutoff=31.5)
detector = abtem.AnnularDetector(inner=70, outer=220)
scan = abtem.GridScan(start=(0, 0), end=(1, 1), fractional=True, potential=potential)

measurement = probe.scan(potential, scan=scan, detectors=detector)
measurement.to_zarr(ostr)   # <-- never consults dask.multi-gpu
```

Despite four visible GPUs and the flag enabled, the whole scan ran serially on GPU 0 and exhausted it within seconds:

```text
Traceback (most recent call last):
  File ".../simscript.py", line 46, in main
    measurement.to_zarr(ostr)
  File "abtem/array.py", line 440, in to_zarr
    output = dask.compute(delayed_write, **kwargs)[0]
  ...
  File "abtem/multislice.py", line 732, in multislice_and_detect
    waves = waves_input.copy()
  ...
cupy.cuda.memory.OutOfMemoryError: Out of memory allocating 2,859,279,872 bytes
                                   (allocated so far: 41,337,662,976 bytes)

real    0m24.467s
```

Each ingredient of this crash maps to a commit below: the ignored flag, the oversized batches on an FFT-unfriendly grid, the redundant copy, the unbounded FFT plan cache — and the fact that nothing in the output hinted at any of it.

</details>

### Correctness

- **`to_zarr()` now distributes.** The multi-GPU cluster bootstrap lived only in `.compute()`; `to_zarr()` on a lazy result forced the synchronous scheduler instead. It now computes through the same scheduler resolution as `.compute()` and writes the result eagerly (`compute=False` still returns the delayed write, unchanged; its nested compute runs under the caller's scheduler — documented in the docstring — so on GPU prefer `compute=True`).
- **Client configuration now reaches the workers.** dask-cuda workers spawn fresh and see only the yaml defaults; `abtem.config.set` in the client never reached them. Because abTEM resolves configuration inside tasks (`get_dtype` reads `precision` at call time), any distributed run with non-default configuration silently used the defaults — our float64 scans were actually computed in float32. The client's merged configuration is now pushed on every dispatch to **any** distributed client (the multi-GPU cluster or a user-provided CPU cluster), carried by a named worker plugin so that workers joining or restarting later receive it too. This bug predates the PR and affects the existing compute path on `dev` too.

### Automatic VRAM handling

- **More conservative batch sizing on FFT-unfriendly grids.** The 6× per-probe overhead estimate — per probe, on the wavefunction grid; the scan positions are the FFT batch dimension and play no part in it — was calibrated on fast-radix grids; on the Bluestein-fallback 2623×2271 grid the real footprint measured ~13×, so the auto-sized batches exhausted the card. Such grids now use 12×.
- **The cuFFT plan cache is bounded: `auto` = 25 % of device memory, per device** (was unlimited). Bluestein-grid plans are large — measured 11.0 GB for one complex128 16-probe batch, 7.2× the batch bytes — so unbounded retention across the several batch shapes of a scan wastes tens of GB. For a single plan larger than the bound, CuPy does not evict or degrade gracefully — it raises a hard `RuntimeError` ("the plan memsize is too large") that aborts the computation (CUDA only; the ROCm path does not track plan memsize). Such plans now run uncached instead, at replanning cost, with a `UserWarning` that names the shape, the bound, and both remedies (raise `cupy.fft-cache-size`, or reduce `max_batch`).
- **A redundant copy of every incoming wave batch in multislice is removed** — one full batch of VRAM saved per task. Results are unchanged (tested).

### Diagnostics and configuration

- abTEM now warns whenever it falls back to something the user did not ask for: multi-GPU requested but declined (with the reason), a missing `if __name__ == "__main__"` guard (dask-cuda spawns workers; the stock error is cryptic), and grid sizes that force cuFFT's slow Bluestein fallback (with a suggested fast size).
- Public `backend.get_cuda_cluster_client()` accessor; new config keys `dask.multi-gpu-rmm-pool` and `dask.multi-gpu-devices` forwarded to the cluster. The fft-size helpers commit is shared verbatim with #347 — whichever PR merges first, the other rebases clean.

## Behavior changes to weigh in review

1. The plan-cache default: `auto` = 25 % of device memory (was unlimited; `-1` restores it).
2. The 12× overhead factor halves auto-sized batches on Bluestein grids only.
3. New `UserWarning`s (multi-GPU declined; slow FFT grid; oversized plan run uncached).
4. Client configuration now reaches workers (any distributed client, not only the multi-GPU cluster) — distributed results change to what the user actually configured.

## Benchmarks

**The production workload** (101,871-position HAADF scan, 2623×2271 grid, float64) on a 40 GB A100 node:

| | 1 GPU | 4 GPUs |
|---|---|---|
| dev | **OOM after 24 s** | no path: `to_zarr()` stays serial; only a manual `.compute()` before saving distributes — and then **silently in float32** |
| this branch | completes: 3089 s | completes: **799 s (3.86×)** |

The 3.86× matches the compute path's previously validated 3.90× scaling — and is itself the end-to-end verification of the configuration fix: before it, distributed runs appeared "12.9× faster than serial", which is impossible at equal precision. Checksums are bit-identical between branches on every benchmark case, serial and distributed runs agree, and repeated runs reproduce within ±0.2 %.

**Everyday single-GPU scans are unaffected.** A100, shipped defaults, medians (run-to-run variance is a few percent):

| case | dev | this branch |
|---|---|---|
| GPU scan 336², 64 positions | 28.7 ms | 30.4 ms |
| GPU scan 2625×2304, 16 positions | 121.6 ms | 120.3 ms |
| GPU scan 2623×2271 (Bluestein), 16 positions | 174.3 ms | 174.0 ms |
| GPU to_zarr 336², 64 positions | 37.5 ms | 36.6 ms |
| CPU scan 336², 16 positions | 198.9 ms | 207.7 ms |
| CPU scan 2625×2304, 4 positions | 3869 ms | 3864 ms |

The same comparison on a local dev machine (AMD Radeon 8050S iGPU via CuPy/ROCm hipFFT; CPU = FFTW, single thread), same-session dev baseline, shipped defaults:

| case | dev | this branch |
|---|---|---|
| GPU scan 336², 64 positions | 55.6 ms | 48.6 ms |
| GPU scan 2625×2304, 16 positions | 1133 ms | 1127 ms |
| GPU scan 2623×2271 (Bluestein), 16 positions | 1519 ms | 1473 ms |
| GPU to_zarr 336², 64 positions | 58.1 ms | 54.0 ms |
| CPU scan 336², 16 positions | 129.9 ms | 123.5 ms |
| CPU scan 2625×2304, 4 positions | 2496 ms | 2477 ms |

Parity or better on every case (up to −13 %); the small-grid gains come from the removed batch copy. Result checksums are bit-identical between branches on both machines.

The plan-cache default was tuned by A/B measurement: an earlier flat 2 GB bound cost 6–8 % on the large fast-grid case by evicting its plans; the device-relative `auto` default restores parity (the 120.3 ms above).

## Testing

- Each commit carries unit tests; `test_multigpu_logic.py` includes process-cluster tests asserting every worker's configuration is identical to the client's (including workers added by scaling up), re-propagation after a change, config push on the CPU compute path, and that a forced synchronous scheduler reaches nested computes.
- Full battery (`test_array`, `test_waves`, `test_scan`, `test_measure`, `test_potential_chunking`, `test_multigpu*`, `test_fft`, `test_copy`, `test_import`): passing on CPU and on a single-GPU ROCm box; the known `test_realspace_multislice[gpu]` failures there pre-exist on unmodified `dev`.

### DIFF
--- a/abtem/array.py
+++ b/abtem/array.py
@@ -430,7 +430,7 @@ class ComputableList(list):
         if is_gpu:
             kwargs = _resolve_gpu_scheduler(dict(kwargs))
 
-        with _compute_context(
+        with _nested_compute_guard(kwargs), _compute_context(
             progress_bar, profiler=False, resource_profiler=False
         ) as (_, profiler, resource_profiler):
             arrays = dask.compute([array for _, array in arrays_to_write], **kwargs)[0]
@@ -584,6 +584,27 @@ def _resolve_gpu_scheduler(kwargs: dict) -> dict:
     return kwargs
 
 
+@contextmanager
+def _nested_compute_guard(kwargs: dict):
+    """Extend a forced synchronous scheduler to nested computes.
+
+    Passing ``scheduler="synchronous"`` to ``dask.compute`` governs only that
+    call's graph. A task body may itself call ``.compute()`` -- e.g.
+    ``CrystalPotential.generate_slices`` pooling a lazy multi-configuration
+    ``potential_unit`` -- and a nested compute reads the process-wide default
+    scheduler: the threaded one, which must not execute CuPy tasks
+    concurrently. Mirror the forced choice into the dask configuration for the
+    duration of the outer compute. When no scheduler is forced (a distributed
+    client is in charge, or the caller chose one), the configuration is left
+    alone so client routing is unaffected.
+    """
+    if kwargs.get("scheduler") == "synchronous":
+        with dask.config.set(scheduler="synchronous"):
+            yield
+    else:
+        yield
+
+
 def _compute(
     array_objects: list[ArrayObjectType],
     progress_bar: Optional[bool] = None,
@@ -598,7 +619,7 @@ def _compute(
     if is_gpu:
         kwargs = _resolve_gpu_scheduler(kwargs)
 
-    with _compute_context(
+    with _nested_compute_guard(kwargs), _compute_context(
         progress_bar, profiler=profiler, resource_profiler=resource_profiler
     ) as (_, profiler, resource_profiler):
         arrays = dask.compute([wrapper.array for wrapper in array_objects], **kwargs)[0]
--- a/abtem/array.py
+++ b/abtem/array.py
@@ -552,18 +552,29 @@ def _resolve_gpu_scheduler(kwargs: dict) -> dict:
     except ValueError:
         client = None
 
+    multi_gpu = config.get("dask.multi-gpu", False)
+
     # Start a dask-cuda cluster spanning all visible GPUs when multi-GPU is
     # enabled, no client is already handling the computation, and the user has
     # not explicitly requested a scheduler.
-    if (
-        client is None
-        and "scheduler" not in kwargs
-        and config.get("dask.multi-gpu", False)
-        and cp.cuda.runtime.getDeviceCount() > 1
-    ):
-        client = ensure_cuda_cluster()
+    if client is None and "scheduler" not in kwargs and multi_gpu:
+        if cp.cuda.runtime.getDeviceCount() > 1:
+            client = ensure_cuda_cluster()
+        else:
+            warnings.warn(
+                "dask.multi-gpu is enabled but only one GPU is visible; "
+                "computing on a single GPU with the synchronous scheduler.",
+                UserWarning,
+            )
 
     if not is_gpu_dask_client(client) and "scheduler" not in kwargs:
+        if client is not None and multi_gpu:
+            warnings.warn(
+                "dask.multi-gpu is enabled but the active dask client is not a "
+                "single-threaded one-worker-per-GPU (dask-cuda) client; "
+                "falling back to the synchronous scheduler on a single GPU.",
+                UserWarning,
+            )
         kwargs["scheduler"] = "synchronous"
 
     return kwargs
--- a/abtem/array.py
+++ b/abtem/array.py
@@ -331,7 +331,6 @@ class ComputableList(list):
 
         if is_zip:
             # Use ZipStore for .zip files
-            @dask.delayed
             def write_to_zipstore(
                 computed_arrays,
                 url,
@@ -372,16 +371,10 @@ class ComputableList(list):
 
                 return url
 
-            delayed_arrays = [
-                (i, dask.delayed(array.compute)()) for i, array in arrays_to_write
-            ]
-            delayed_write = write_to_zipstore(
-                delayed_arrays, url, metadata_list, overwrite, compressors=compressors
-            )
+            write_func = write_to_zipstore
 
         else:
             # Use directory store for non-.zip files
-            @dask.delayed
             def write_to_directory(computed_arrays, url, metadata_list, overwrite):
                 import shutil
 
@@ -411,33 +404,42 @@ class ComputableList(list):
 
                 return url
 
+            write_func = write_to_directory
+
+        if not compute:
             delayed_arrays = [
                 (i, dask.delayed(array.compute)()) for i, array in arrays_to_write
             ]
-            delayed_write = write_to_directory(
+            return dask.delayed(write_func)(
                 delayed_arrays, url, metadata_list, overwrite
             )
 
-        if not compute:
-            return delayed_write
-
-        # On GPU, force the synchronous scheduler to avoid multiple dask tasks
-        # running in parallel, each loading potential chunks + wave arrays into
-        # VRAM simultaneously. We must use dask.config.set() (via the guard) rather
-        # than just passing scheduler= to dask.compute(), because the delayed graph
-        # contains inner array.compute() calls that would otherwise use dask's
-        # default threaded scheduler. The guard makes an exception for an active
-        # single-threaded dask-cuda client, which already runs one task per GPU.
+        # Compute the arrays first -- resolving the GPU execution context the
+        # same way as ArrayObject.compute, so the opt-in multi-GPU cluster
+        # (``dask.multi-gpu``) is started on this path too -- then write the
+        # gathered results eagerly. Writing through a delayed task that itself
+        # calls array.compute() would nest a compute inside a scheduler task:
+        # the synchronous scheduler then needs a global config override to
+        # reach the nested call, and under a distributed client the nested
+        # compute would run on the local scheduler inside a single worker,
+        # serializing the whole computation onto one GPU.
         is_gpu = config.get("device") == "gpu" or any(
             _is_gpu_array_object(obj) for obj in self
         )
         if is_gpu:
-            check_cupy_is_installed()
+            kwargs = _resolve_gpu_scheduler(dict(kwargs))
 
-        with _gpu_scheduler_guard(is_gpu), _compute_context(
+        with _compute_context(
             progress_bar, profiler=False, resource_profiler=False
         ) as (_, profiler, resource_profiler):
-            output = dask.compute(delayed_write, **kwargs)[0]
+            arrays = dask.compute([array for _, array in arrays_to_write], **kwargs)[0]
+
+        output = write_func(
+            [(i, array) for (i, _), array in zip(arrays_to_write, arrays)],
+            url,
+            metadata_list,
+            overwrite,
+        )
 
         profilers = tuple(p for p in (profiler, resource_profiler) if p is not None)
         if profilers:
@@ -518,20 +520,30 @@ def _is_gpu_array_object(obj) -> bool:
     return hasattr(obj, "device") and obj.device == "gpu"
 
 
-@contextmanager
-def _gpu_scheduler_guard(is_gpu: bool):
-    """Guard nested/implicit GPU computes against the threaded scheduler.
+def _resolve_gpu_scheduler(kwargs: dict) -> dict:
+    """Resolve the dask scheduler for a GPU computation.
 
-    On GPU, force the synchronous scheduler so that nested ``array.compute()``
-    calls cannot fall back to dask's default threaded scheduler and load multiple
-    potential/wave chunks into a single GPU's memory at once. When a single-
-    threaded dask-cuda client is active (one worker pinned per GPU), that per-GPU
-    exclusivity already holds inside each worker, so the work is left to the
-    client to distribute across the cluster instead of being forced synchronous.
+    Starts the opt-in multi-GPU cluster (``dask.multi-gpu``) when more than one
+    GPU is visible and no client or explicit scheduler is already in charge.
+    When no suitable single-threaded dask-cuda client ends up handling the
+    computation, the synchronous scheduler is selected instead: the threaded
+    scheduler and multi-threaded workers share a single CUDA context per
+    process, which cannot be used with CuPy, and concurrent tasks would load
+    multiple potential/wave chunks into a single GPU's memory at once.
+
+    Parameters
+    ----------
+    kwargs : dict
+        Keyword arguments destined for ``dask.compute``. A ``scheduler`` key
+        set by the caller is always respected.
+
+    Returns
+    -------
+    dict
+        The keyword arguments, with ``scheduler="synchronous"`` injected when
+        no suitable client is available.
     """
-    if not is_gpu:
-        yield
-        return
+    check_cupy_is_installed()
 
     from distributed import get_client
 
@@ -540,11 +552,21 @@ def _gpu_scheduler_guard(is_gpu: bool):
     except ValueError:
         client = None
 
-    if is_gpu_dask_client(client):
-        yield
-    else:
-        with dask.config.set(scheduler="synchronous"):
-            yield
+    # Start a dask-cuda cluster spanning all visible GPUs when multi-GPU is
+    # enabled, no client is already handling the computation, and the user has
+    # not explicitly requested a scheduler.
+    if (
+        client is None
+        and "scheduler" not in kwargs
+        and config.get("dask.multi-gpu", False)
+        and cp.cuda.runtime.getDeviceCount() > 1
+    ):
+        client = ensure_cuda_cluster()
+
+    if not is_gpu_dask_client(client) and "scheduler" not in kwargs:
+        kwargs["scheduler"] = "synchronous"
+
+    return kwargs
 
 
 def _compute(
@@ -559,31 +581,7 @@ def _compute(
     )
 
     if is_gpu:
-        check_cupy_is_installed()
-
-        from distributed import get_client
-
-        try:
-            client = get_client()
-        except ValueError:
-            client = None
-
-        # Start a dask-cuda cluster spanning all visible GPUs when multi-GPU is
-        # enabled, no client is already handling the computation, and the user has
-        # not explicitly requested a scheduler.
-        if (
-            client is None
-            and "scheduler" not in kwargs
-            and config.get("dask.multi-gpu", False)
-            and cp.cuda.runtime.getDeviceCount() > 1
-        ):
-            client = ensure_cuda_cluster()
-
-        # The threaded scheduler and multi-threaded workers cannot be used with CuPy;
-        # fall back to synchronous execution unless a suitable (dask-cuda) client is
-        # handling the computation.
-        if not is_gpu_dask_client(client) and "scheduler" not in kwargs:
-            kwargs["scheduler"] = "synchronous"
+        kwargs = _resolve_gpu_scheduler(kwargs)
 
     with _compute_context(
         progress_bar, profiler=profiler, resource_profiler=resource_profiler
--- a/abtem/array.py
+++ b/abtem/array.py
@@ -41,6 +41,7 @@ from abtem.core.axes import (
     axis_to_dict,
 )
 from abtem.core.backend import (
+    push_config_to_workers,
     check_cupy_is_installed,
     copy_to_device,
     cp,
@@ -576,6 +577,9 @@ def _resolve_gpu_scheduler(kwargs: dict) -> dict:
                 UserWarning,
             )
         kwargs["scheduler"] = "synchronous"
+
+    if is_gpu_dask_client(client):
+        push_config_to_workers(client)
 
     return kwargs
 
--- a/abtem/array.py
+++ b/abtem/array.py
@@ -41,7 +41,6 @@ from abtem.core.axes import (
     axis_to_dict,
 )
 from abtem.core.backend import (
-    push_config_to_workers,
     check_cupy_is_installed,
     copy_to_device,
     cp,
@@ -49,6 +48,7 @@ from abtem.core.backend import (
     ensure_cuda_cluster,
     get_array_module,
     is_gpu_dask_client,
+    push_config_to_workers,
 )
 from abtem.core.chunks import Chunks, iterate_chunk_ranges, validate_chunks
 from abtem.core.ensemble import Ensemble, _wrap_with_array, unpack_blockwise_args
@@ -243,6 +243,11 @@ class ComputableList(list):
             For directory stores, use .zarr extension or a directory path.
         compute : bool
             If true compute immediately; return dask.delayed.Delayed otherwise.
+            Note that the returned delayed write nests an ``array.compute()``
+            executed under whatever scheduler is active when the caller
+            finally computes it -- the multi-GPU bootstrap and the forced
+            synchronous GPU scheduler apply only to ``compute=True``. On GPU,
+            prefer ``compute=True`` or compute the result before saving.
         overwrite : bool
             If given array already exists, overwrite=False will cause an error, where
             overwrite=True will replace the existing data.
--- a/abtem/array.py
+++ b/abtem/array.py
@@ -427,6 +427,8 @@ class ComputableList(list):
         is_gpu = config.get("device") == "gpu" or any(
             _is_gpu_array_object(obj) for obj in self
         )
+        _push_config_to_active_client()
+
         if is_gpu:
             kwargs = _resolve_gpu_scheduler(dict(kwargs))
 
@@ -605,6 +607,24 @@ def _nested_compute_guard(kwargs: dict):
         yield
 
 
+def _push_config_to_active_client():
+    """Mirror the configuration onto an active distributed client's workers.
+
+    The silently-defaulted-configuration bug is not specific to the multi-GPU
+    cluster: any distributed client (a CPU LocalCluster, a SLURMCluster, ...)
+    executes tasks in worker processes that never saw the client's
+    ``abtem.config.set``. Deduplicated inside push_config_to_workers, so
+    calling this on every dispatch is cheap.
+    """
+    try:
+        from distributed import get_client
+
+        client = get_client()
+    except (ImportError, ValueError):
+        return
+    push_config_to_workers(client)
+
+
 def _compute(
     array_objects: list[ArrayObjectType],
     progress_bar: Optional[bool] = None,
@@ -615,6 +635,8 @@ def _compute(
     is_gpu = config.get("device") == "gpu" or any(
         _is_gpu_array_object(obj) for obj in array_objects
     )
+
+    _push_config_to_active_client()
 
     if is_gpu:
         kwargs = _resolve_gpu_scheduler(kwargs)
--- a/abtem/core/abtem.yaml
+++ b/abtem/core/abtem.yaml
@@ -40,6 +40,7 @@ cupy:
   # > 0   — bound cached workspace to this many bytes (e.g. "512 MB").
   # -1    — unlimited (CuPy default; fastest but accumulates workspace,
   #          which on Bluestein-fallback grid sizes can reach tens of GB).
+  # null  — same as -1 (no bound).
   fft-cache-size: auto
 mkl:
   # The number of threads to use for mkl
--- a/abtem/core/abtem.yaml
+++ b/abtem/core/abtem.yaml
@@ -34,8 +34,9 @@ cupy:
   # 0 MB  — disable caching entirely (workspace freed after every FFT call;
   #          use when VRAM is tight and large-grid OOMs occur).
   # > 0   — bound cached workspace to this many bytes (e.g. "512 MB").
-  # -1    — unlimited (CuPy default; fastest but accumulates workspace).
-  fft-cache-size: -1
+  # -1    — unlimited (CuPy default; fastest but accumulates workspace,
+  #          which on Bluestein-fallback grid sizes can reach tens of GB).
+  fft-cache-size: 2 GB
 mkl:
   # The number of threads to use for mkl
   threads: 2
--- a/abtem/core/abtem.yaml
+++ b/abtem/core/abtem.yaml
@@ -21,6 +21,13 @@ dask:
   # Automatically distribute gpu computations over all visible gpus by starting a
   # dask-cuda cluster. Requires the optional dask-cuda package.
   multi-gpu: false
+  # Optional RMM memory-pool size per multi-gpu worker (e.g. "20 GB").
+  # null disables the RMM pool.
+  multi-gpu-rmm-pool: null
+  # Optional subset of GPUs for the multi-gpu cluster, as a list of device
+  # indices (e.g. [0, 1]) or a comma-separated string (e.g. "0,1").
+  # null spans all visible GPUs.
+  multi-gpu-devices: null
 cupy:
   # Maximum GPU memory (bytes) used by the cuFFT plan cache.
   # https://docs.cupy.dev/en/stable/user_guide/fft.html#fft-plan-cache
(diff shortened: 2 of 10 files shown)


pzeiger:
@TomaSusi: what do you think about this? Any questions/concerns? I will try to benchmark if/how computation time is affected by these changes before making this a proper PR.

TomaSusi:
Will have a deeper look on Monday, but a quick question before that: this all stacks on top of the improvements I made in https://github.com/abTEM/abTEM/pull/269?

In particular, I took quite a lot of care that calculations do not OOM on my single RTX4090 (unless a single potential slice + overheads, which I was also estimating, didn’t fit in VRAM), so I am wondering if some of this is so architecture dependent that that was not sufficient.

In any case, it sounds like you maybe have made some better justified estimates, and multi-GPU I did not touch, so this is very welcome! Would be good to benchmark on multiple systems though to make sure this is robust (and preserves performance) across architectures.

pzeiger:
Hi @TomaSusi!

Yes, this stacks on top of it. I have made sure to base everything on the latest dev state on Friday. Most of it is 

How large grids did you try? And where they all by chance *FFT-friendly* as defined in PR #347? I have observed that large *FFT-unfriendly* grids lead to OOMs on data center GPUs, but I could not reliably benchmark this...

I have added benchmark numbers in the original PR. It looks like there is no performance degradation, but, on the contrary, run times actually improved slightly with a larger relative improvement to smaller calculations. Claude attributes this mostly to removing a unnecessary deepcopy of an array somewhere in the multislice part.

Best,
Paul

TomaSusi:
> How large grids did you try? And where they all by chance _FFT-friendly_ as defined in PR #347? I have observed that large _FFT-unfriendly_ grids lead to OOMs on data center GPUs, but I could not reliably benchmark this...

I went up to 16384^2 grids, but yes these were all FFT-friendly sizes. My benchmark numbers (on an RTX 4090) are in the original PR:
https://github.com/abTEM/abTEM/pull/269

pzeiger:
Hi @TomaSusi,

I believe this is in a good state now. I have extensively benchmarked and tested the changes. Here is a summary of how this PR evolved since I opened it. It grew from 8 to 11 commits, and the description was rewritten around the original failure log and a complete benchmark matrix:

- Rebased on current `dev` (includes #350).
- The most important addition: **abTEM configuration now propagates to dask-cuda workers.** Previously `abtem.config.set` in the client never reached them, so any distributed computation with non-default configuration silently used the yaml defaults — we found this because our float64 production scans were actually computing in float32 inside the workers (discovered via cuFFT plan-size forensics, then confirmed by the scaling arithmetic: physically impossible better-than-linear "speedups" collapsed to a clean 3.86× on 4 GPUs once fixed). The bug predates this PR and affects the existing compute path on `dev` as well.
- The cuFFT plan-cache bound went through an A/B-driven evolution: a flat 2 GB (measured to cost ~7 % on fast-radix scans) became `auto` = 25 % of device memory, resolved per device. Along the way a latent CuPy semantic surfaced — a single plan larger than the bound makes CuPy raise rather than run uncached — now handled by a fallback that runs oversized plans uncached with an actionable warning.
- The benchmark matrix in the description covers serial/distributed × `dev`/this branch at true float64: `dev` OOMs serially and silently downgrades precision when distributed; this branch completes in 3089 s serially and 799 s on 4 GPUs (3.86×, matching the compute path's validated 3.90×), with bit-identical checksums across branches and agreement between serial and distributed runs.
- One commit (the fft helpers + Bluestein grid warning) is shared verbatim with #347; whichever merges first, the other rebases clean.

I am happy to split anything out if you would prefer smaller reviews/changes.

Best,
Paul

TomaSusi:
## Review notes

Ran an automated review across the diff (`origin/dev...multi-gpu-hardening`). Several of these were independently flagged by multiple review passes and verified against the actual code paths — details below.

### Correctness

**1. Nested `.compute()` calls lose the synchronous-scheduler guarantee on GPU** (`abtem/potentials/iam.py:2279`)
The old `_gpu_scheduler_guard` used `dask.config.set(scheduler="synchronous")`, which is process-wide — any nested `.compute()` anywhere in the call stack inherited it. The new `_resolve_gpu_scheduler` only returns a `kwargs` dict threaded into one explicit `dask.compute()` call, so a `.compute()` invoked from inside a task body (e.g. `CrystalPotential.generate_slices()`'s `_pool_array.compute()` when `potential_unit` is a precomputed, multi-config `PotentialArray`) silently falls back to the threaded scheduler — reopening the concurrent-CUDA-context hazard the old guard existed to prevent.

**2. Config push races a worker's in-flight task** (`abtem/core/backend.py:177`)
`_apply_config_snapshot` does `config_dict.clear()` then `.update(snapshot)` with no lock (abTEM already has a `config_lock` for this). `client.run()` executes on the worker's event-loop thread, distinct from the task-executor thread pool. A task calling `config.get("precision")` (e.g. inside `get_dtype()`) during the clear/update window can observe the momentarily-empty dict and raise `RuntimeError("Invalid dtype: None")`.

**3. Config-push memoization skips restarted/new workers** (`abtem/core/backend.py:199`)
The skip check `token = (id(client), repr(snapshot))` tracks client identity + config content, not which workers actually received the push. `LocalCUDACluster` runs workers under a `Nanny` by default — a restarted worker keeps default config indefinitely since the client/token is unchanged. Separately, if a cluster is torn down and recreated, CPython can reuse the old `Client`'s `id()`, causing the new cluster's workers to be silently skipped too.

**4. Config-push fix is GPU-only, but described as general** (`abtem/array.py:594`)
`push_config_to_workers` is only reachable via the `is_gpu` branch. The PR description says "the client's merged configuration is now pushed to the workers on every dispatch" — but a CPU-based `distributed` cluster (`LocalCluster`, `SLURMCluster`, etc.) still has the exact silently-wrong-config bug this PR frames as fixed generally. Worth either narrowing the PR language or extending the fix to the CPU path.

**5. `cupy.fft-cache-size = -1` can't undo an earlier bound** (`abtem/core/fft.py:320`)
Once `"auto"` (the new default) has bounded the plan cache to a positive memsize, later setting `-1` hits the `limit < 0` branch, which calls neither `set_size` nor `set_memsize` — the earlier bound just persists instead of restoring "unlimited." This defeats the exact remedy `_cupy_fft_with_cache_fallback`'s own warning tells users to apply.

**6. `cupy.fft-cache-size: null` crashes** (`abtem/core/fft.py:321`)
Falls through to `int(None)` → unhandled `TypeError`. Worth a guard, especially since the two sibling keys added in this same PR (`multi-gpu-rmm-pool`, `multi-gpu-devices`) establish `null` as the "disable/default" convention right next to this one.

**7. `to_zarr(compute=False)` still nests a compute on GPU** (`abtem/array.py:410`)
Pre-existing gap, and disclosed in the commit message ("compute=False still returns the delayed write unchanged") — flagging since it's untested under `device="gpu"` and easy to lose track of now that the `compute=True` path looks fixed.

**8. New Bluestein-overhead logic sits inside a silent catch-all** (`abtem/core/chunks.py:630`, lower confidence)
The new `is_fast_fft_size` check in `estimate_scan_batch_size` is wrapped by `except (ImportError, Exception):`, which would silently fall back to the old fixed heuristic on any exception. Low probability of triggering in practice, but no diagnostic trail if it does.

### Cleanup

**9. `is_fast_fft_size`/`next_fast_fft_size` duplicate `Grid.round_to_power`** (`abtem/core/grid.py:385`)
Same "fast radix = {2,3,5,7}" logic, different algorithm (coarser power-rounding vs. general smooth-number search). Worth consolidating so grid-sizing and the new FFT-cost warnings can't drift apart.

**10. isort ordering** (`abtem/array.py:44`)
`push_config_to_workers` is inserted first in the `from abtem.core.backend import (...)` block rather than alphabetically — `isort --check-only` fails on this as committed.

pzeiger:
Thanks for the review, this is very helpful. I think these are all valid points and I will work on resolving them.

pzeiger:
Hi @TomaSusi,

*(This reply — like the fix commits it describes — was written by Claude; Paul reviewed and posted it.)*

Thanks — this was a genuinely useful review; 8 of the 10 points are fixed in four new commits (781b9440, db7839da, d6dfe5f9, 63226c7c), and the remaining two are deferred with reasons below. Point by point:

1. **Fixed** (781b9440). One nuance worth recording: the kwargs-only gap was not new on the `.compute()` path — `dev`'s `_compute` also injected `scheduler="synchronous"` as a kwarg only, so nested computes there have always read the threaded default. The regression this PR introduced was limited to the `to_zarr` path, where the old guard did set the global config. Both paths now mirror a *forced* synchronous scheduler into the dask configuration for the duration of the outer compute — and only then, so distributed client routing is untouched. Test included, using exactly the nested-compute-reads-config mechanism you described.
2. **Fixed** (db7839da), with one refinement over locking alone: `config.get` readers do not take `config_lock`, so a locked clear-then-update would still expose the empty window to them. The applier now updates under the lock with update-then-prune ordering — lockless readers always observe a fully-populated dict.
3. **Fixed** (db7839da). The snapshot now travels as a named worker plugin — `setup` runs at registration, on join, and on Nanny restart — replacing the one-shot `client.run`, and the memoization token is keyed on `client.id` (the unique string) instead of `id(client)`. The identical-config test now also scales the cluster up and asserts the new worker received the configuration.
4. **Extended rather than narrowed** (db7839da): the push now happens on every compute dispatch to any active distributed client, CPU clusters included, so the PR description's general wording is now accurate. An end-to-end CPU-LocalCluster test covers it.
5. **Fixed** (d6dfe5f9): `-1` now explicitly restores the unlimited memsize. While there, a positive bound also re-enables a cache previously disabled with `"0 MB"` — the same class of one-way transition.
6. **Fixed** (d6dfe5f9): `null` is accepted as "no bound", matching the sibling keys' convention, with the yaml comment updated.
7. **Deferred, now documented** (63226c7c): agreed it is easy to lose track of — the docstring now states that the `compute=False` delayed write executes its nested compute under the caller's scheduler, outside the multi-GPU bootstrap and the forced synchronous scheduler. Restructuring that path is real design work for a pre-existing, rarely used branch; happy to file a follow-up issue if you agree.
8. **Fixed** (63226c7c): the radix check is hoisted out of the `try` (it cannot fail and needs no GPU), which now guards only the CUDA memory probe, and the redundant `(ImportError, Exception)` is narrowed.
9. **Deferred to #347 deliberately**: `Grid.round_to_fast_fft()` there is the consolidation — it supersedes what `round_to_power`'s docstring promises (the implementation actually produces pure powers: 2623 → 4096, +56 %
   pixels vs +0.08 % for 2625), and deprecating/redirecting `round_to_power` is flagged in that PR. Doing it here would duplicate and entangle the two PRs; the fft helpers commit is the only intended overlap.
10. **Fixed** (63226c7c): import block alphabetized.

Full battery after the fixes: 574 passed (+6 GPU skips) plus the focused
suites; the four commits are additive on top of the rebased branch, no
force-push.

Best,
Paul

TomaSusi:
Thank you Paul, this is starting to look very close to ready! 

Reviewed the four fix commits (781b9440, db7839da, d6dfe5f9, 63226c7c) against the original 10 points — 8 are correctly and solidly fixed, and I agree with deferring #7 (to_zarr compute=False) and #9 (round_to_power, properly #347's job).

One new issue slipped in with the point-8 cleanup, though:

**[`abtem/core/chunks.py:475`](https://github.com/abTEM/abTEM/blob/63226c7ce011b5139874b83bfb7fbed4b98a6ad8/abtem/core/chunks.py#L475), in `estimate_potential_chunk_size`** — the hoisted-out-of-try radix-check block got copy-pasted into this function as well, but here `overhead` is dead: the per-slice cost a few lines down still uses a hardcoded `slice_bytes * 5` and never reads it. Before this commit `estimate_potential_chunk_size` had no FFT-friendliness logic at all; now it computes `is_fast_fft_size` over `gpts` on every GPU call and discards the result. Beyond the wasted work, it's a bit of an attractive nuisance — it mirrors the *real* usage in the sibling `estimate_scan_batch_size` closely enough that a future reader could assume it's already wired in here too and skip adding the multiplier where it's actually needed. `test_potential_chunking.py` doesn't parametrize an FFT-unfriendly grid on GPU, so nothing caught it.

Worth either wiring `overhead` into `effective_per_slice` here too (if the Bluestein-overhead reasoning is meant to apply to potential slices as well as scan probes) or dropping the dead block.

🤖 Written by Claude Code on Toma's behalf

TomaSusi:
I'm now working to update the public documentation pages. Would you like to update https://github.com/abTEM/doc/pull/19 once this is finished and merged – I expect we can do this right away now :)

TomaSusi (approved):
Although I did not directly test these changes, based on the PR discussion and my reading of the docstrings this seems more than good enough for now.

pzeiger:
Hi @TomaSusi,

*(This reply — like the commit it describes — was written by Claude; Paul reviewed and posted it.)*

Confirmed and fixed in b584db59: the block was dead in `estimate_potential_chunk_size`, and the right resolution is the second one you offered — **drop it**, not wire it in.

The reason is that the Bluestein factor is not a property this estimate can use. Every FFT that touches a potential slice is a *single, unbatched* 2D transform: `_FieldBuilder.build()` writes one slice at a time into the preallocated chunk, `PotentialArray.generate_slices()` hands the multislice loop one slice at a time, the projection integrals in `integrals.py` transform a bare `(ny, nx)` array, and `multislice_step` bandlimits one transmission function per step. So the Bluestein workspace is *constant in `chunk_size`*, whereas `effective_per_slice` is multiplied by it — wiring `overhead` in would have halved the chunk on an FFT-unfriendly grid to pay for a cost that does not grow with the chunk. The probe batch in `estimate_scan_batch_size` is the one quantity whose FFT genuinely *is* batched over the dimension being sized, which is why the factor belongs there alone. And what residual plan cost does exist is already visible to `estimate_potential_chunk_size` for free: unlike its sibling it runs at *computation* time and reads the live free-VRAM figure, so a resident Bluestein plan has already been subtracted from the budget it is dividing up.

That reasoning is now a comment where the block used to be, which should also defuse the attractive-nuisance concern — a reader arriving from `estimate_scan_batch_size` is told why the factor is missing rather than left to assume it was forgotten.

On the coverage gap: `test_potential_chunking.py` now parametrises the GPU path over an FFT-unfriendly grid, pinning the *invariance* rather than a multiplier. 2623 × 2271 (Bluestein) and 2625 × 2268 (7-smooth) differ in area by 0.06 % and must give the same chunk size — 117 against a mocked 40 GB device, fixed by bytes alone; a doubled overhead would have given 58. It uses the same `_install_fake_cupy` stand-in as the existing scan-batch tests, so it runs in CI without a GPU.

One clarification while we are here, since the wording in the review invited it: `gpts` in `estimate_scan_batch_size` is the **wavefunction grid each probe lives on, not the scan grid**. The scan positions are the FFT *batch* dimension and have no bearing on whether cuFFT falls back to Bluestein — only the transform lengths `ny`/`nx` do. The comment there now says so explicitly, so the two functions cannot be misread as making claims about scan sampling.

Separately: I have updated abTEM/doc#19 to cover this PR, and left the details in a comment there. Two points need your judgement rather than just review — the July "super-linear ~4.3x on four GPUs" tip is **removed**, because that benchmark predates the configuration-propagation fix and so almost certainly compared a float64 single-GPU baseline against float32 workers (it is replaced by measured equal-precision numbers); and #19 may want retargeting from `main` to `v1.1.0`, which is your call rather than mine.

Test status unchanged otherwise: 1581 passed on this branch, with the only failures being the pre-existing ROCm/gfx1151 ones on our dev box (8 in `test_realspace_multislice.py`, 2 in `test_measure.py`, plus a `test_prism_upsample.py` segfault) — all of which reproduce identically on the branch point.

Best,
Paul

TomaSusi:
Reviewed b584db59 — and I verified the reasoning rather than taking it on faith, because "drop it" vs "wire it in" hinged entirely on whether the claim about unbatched transforms was true. It is:

- [`integrals.py:493`](https://github.com/abTEM/abTEM/blob/b584db5983b6fcd2f0c9742b3be5e42e9bc6f2a2/abtem/integrals.py#L493) allocates `temp_array = xp.zeros(gpts)` — a bare `(ny, nx)` array — and transforms that; `generate_slices` drives `generate_chunks(..., chunks=1)`, so slices really are emitted one at a time; and `conventional_multislice_step` bandlimits a single transmission function per step. So the Bluestein workspace is constant in `chunk_size` while `effective_per_slice` is multiplied by it — folding the factor in would indeed have been wrong. Good call, and the comment left in its place is the right amount of explanation for the next reader arriving from the sibling function.
- The test arithmetic is exact: 2623 = 43×61 and 2271 = 3×757 (Bluestein), 2625 = 3·5³·7 and 2268 = 2²·3⁴·7 (7-smooth), areas differing by 0.056 %, and `int(0.35 × 40 GB / (2623·2271·4 × 5))` = 117 for both. Pinning the *invariance* rather than a multiplier is the right shape for that test.
- The `gpts`-is-the-wavefunction-grid-not-the-scan-grid clarification is accurate and worth having — thanks for pre-empting that misreading.

Also thanks for handling abTEM/doc#19. Agreed on removing the July super-linear tip — that comparison can't have been like-for-like given the propagation bug. I'll take the `main` vs `v1.1.0` retargeting question over there.

**One last thing before this is ready to merge**, and it is purely test hygiene — no product code involved:

Running `pytest test/test_potential_chunking.py test/test_fft.py test/test_multigpu_logic.py` **serially** gives a clean 102 passed / 10 skipped. Under `-n auto` (my default, and how parallel CI would run it) it fails **2 of the 3 distributed tests on every run**, with the failing pair varying run to run:

```
RuntimeError: Cluster failed to start: Port 8787 is already in use.
PytestUnraisableExceptionWarning: Exception ignored in: <socket.socket fd=16, ...>
```

Root cause, confirmed empirically: `dashboard_address=None` does **not** disable the dashboard in distributed 2026.6.0. `LocalCluster` still starts bokeh and tries the **default port 8787 first**, falling back to an ephemeral port only when the bind fails. I ran four concurrent clusters with `dashboard_address=None` and one grabbed `:8787` while the rest got random ports — so when two start simultaneously they can both observe 8787 free, and the loser raises instead of retrying. The leaked-socket warning is downstream of the failed startup, and it is fatal only because `pyproject.toml` sets `filterwarnings = ["error"]`.

The three new distributed tests use `dashboard_address=None`:

- `test_multigpu_logic.py:470` (`test_push_config_to_workers`)
- `test_multigpu_logic.py:500` (`test_workers_see_identical_config`)
- `test_multigpu_logic.py:571` (`test_cpu_distributed_compute_pushes_config`)

The pre-existing test at `test_multigpu_logic.py:58` already gets this right with `dashboard_address=":0"`. Switching the three to `":0"` gave me 3/3 clean `-n auto` runs (28 passed, 4 skipped each).

Worth landing since otherwise these tests are flaky for anyone running parallel pytest — but with that fixed I think we are good to merge.

TomaSusi:
Verified 8a32240f — that was the last open item, and I think this is ready to merge.

- The three targeted files under `-n auto`, three runs: **102 passed, 10 skipped** every time (was 2 failed / 2 errors per run before).
- Full suite under `-n auto`: **1187 passed, 800 skipped, 7 failed** — all 7 in `test_gpaw.py` with `ValueError: ... grid too small`. I checked out the merge base (2f931826) and ran `test_gpaw.py` there: the same 7 fail. Pre-existing environment issue on my machine, unrelated to this PR (and distinct from the ROCm/gfx1151 ones on your dev box).

So: all 10 original findings fixed or deliberately deferred with reasons I agree with, the follow-up dead-code issue resolved in b584db59, and the test flakiness gone. Thanks for the unusually thorough turnaround on all of this — the configuration-propagation bug in particular was a good catch, and the fact that it predates this PR makes it worth calling out in the release notes.

Two loose ends, neither blocking the merge:

- abTEM/doc#19 — I will answer the `main` vs `v1.1.0` retargeting question over there.
- The shared fft-helpers commit means whichever of this and #347 lands second needs a rebase. I will sequence them.

TomaSusi:
Now that #347 (fast-fft-grid) and autogrid are on `dev`, this needs a merge. I resolved both conflicts locally and ran the suite, so here is exactly what is involved — including one trap where the mechanical resolution is silently wrong.

Both conflicts are in the **fft-helpers commit the two PRs shared verbatim**, exactly the overlap you flagged. #347 landed first and its later review rounds improved those helpers, so `dev`'s copy now strictly supersedes this branch's.

### `abtem/core/fft.py` — 6 hunks

Five are pure "take `dev`": the `_FAST_FFT_PRIMES` scipy/radix-11 rationale comment, the `next_fast_fft_size` docstring, the new `_warned_slow_fft_shapes_lock`, the rewritten `_warn_slow_fft_size` (now axis-aware via `_transform_lengths`, thread-safe, with the smarter remedy text), and the `dev`-only `warn_if_slow_gpu_fft`.

The sixth needs **both sides** — this is the one to watch:

```python
_configure_cufft_cache()
_warn_slow_fft_size(x.shape, func_name, kwargs)                                # dev's new signature
return _cupy_fft_with_cache_fallback(getattr(cp.fft, func_name), x, **kwargs)  # this PR's wrapper
```

Taking `dev` wholesale here drops `_cupy_fft_with_cache_fallback` from the dispatch path. It would still be defined, still be tested in isolation, and never actually be called — the oversized-plan fallback would be dead code, with no marker left behind to notice it.

### `test/test_fft.py` — 2 hunks

Same trap, and worse. `dev`'s expanded helper tests supersede this branch's, but the four cuFFT-plan-cache tests unique to this PR sit *inside* the conflict region, so a plain "take theirs" deletes them silently:

- `test_cufft_cache_auto_resolves_device_relative`
- `test_oversized_plan_bypasses_cache`
- `test_unrelated_runtime_error_propagates`
- `test_cufft_cache_config_edge_values`

They need re-appending after the resolution. All 15 tests should be present when you are done.

### One thing that auto-merged, worth a second look

`_configure_cufft_cache` did **not** conflict. `dev` still carries the old pre-#346 version — default `"0 MB"`, no per-device keying, and the `-1`-is-a-no-op bug fixed here as review point 5. Git took this branch's improved version because `dev` never touched that function since the merge base, which is the right outcome — but it is the kind of thing a rebase could quietly get backwards. Worth confirming `raw = config.get("cupy.fft-cache-size", "auto")` survives whichever way you go.

### Verification

Full suite on my merged tree under `-n auto`: **1315 passed, 821 skipped, 7 failed** — all 7 are the `test_gpaw.py` `... grid too small` errors, which I confirmed fail identically on plain `origin/dev`. Nothing attributable to the merge.

It would also be good to run multi-GPU tests one more time before we merge.

pzeiger:
Hi @TomaSusi,

*(This reply — like the merge it describes — was written by Claude; Paul reviewed and posted it.)*

Merged in bbfba8b5. Your map was exact, including both traps, and following it directly was much faster than rediscovering them — thank you for writing it out rather than just saying "needs a rebase".

**`abtem/core/fft.py`** — five hunks taken from `dev` unchanged, as you listed. The sixth kept both sides:

```python
_configure_cufft_cache()
_warn_slow_fft_size(x.shape, func_name, kwargs)                                # dev's signature
return _cupy_fft_with_cache_fallback(getattr(cp.fft, func_name), x, **kwargs)  # this PR's wrapper
```

You are right that taking `dev` there is silently wrong rather than loudly wrong: `_cupy_fft_with_cache_fallback` would still be defined and still be green in its own tests, so nothing would have failed — the oversized-plan fallback would just never run.

**`test/test_fft.py`** — `dev`'s helper tests taken, then the four plan-cache tests re-appended. All 15 present, and I checked the names against your list rather than the count alone, since the count would also have been 15 if a helper test had gone missing instead.

**`_configure_cufft_cache`** — confirmed. `raw = config.get("cupy.fft-cache-size", "auto")` survives, with the per-device keying and the `-1`-is-a-no-op fix. Worth noting for anyone reading later that this one is safe *because* it auto-merged: git had no reason to prefer `dev`'s copy, since `dev` has not touched the function since the merge base. A rebase resolving hunk by hunk is where it could have gone backwards.

`abtem/core/abtem.yaml` auto-merged cleanly — `grid.round-to-fast-fft` from #347 and this branch's `cupy`/`dask` keys are all present, which was the other thing I wanted to see with my own eyes.

**Verification.** Full suite on the merged tree: **1730 passed, 318 skipped**, with the same 10 ROCm/gfx1151 failures this box has on the branch point. The multi-GPU tests you asked about specifically: `test_multigpu_logic.py`, `test_multigpu.py`, `test_fft.py`, `test_potential_chunking.py` and `test_commensurate_gpts.py` give 137 passed / 6 skipped serially, and the parallel gate is three clean `-n auto` runs at 118 passed each. CI is green on 3.11 through 3.14.

One unrelated thing I noticed while checking the merged `abtem.yaml`, since it touches a file this PR edits heavily. #356 reports `import abtem` crashing on a Windows GBK locale because `config.py` opens `abtem.yaml` without an explicit encoding, and the reported failure is `byte 0x94`. That is the third byte of an em dash: `—` is `e2 80 94` in UTF-8, a GBK decoder consumes `e2 80` as a plausible double-byte character and then hits `0x94` with no valid trail byte. Every non-ASCII character in that file is an em dash, and there is no other source of `0x94` in it, so those comment dashes are the whole cause.

`dev` has three of them, all in the `cupy.fft-cache-size` comment block from #269, so the bug predates this PR. This PR takes it to six — the `auto` line, the "ceiling, not a reservation" clause and the `null` line — all in the same block. So it does not introduce the crash, but it does add three more ways to trip it.

It is worth stressing that being inside comments does not spare them, which is the intuitive-but-wrong reading here: `open(fn)` with no `encoding=` decodes the whole file with the locale codec, and `yaml.safe_load(f)` pulls the stream through that decoder, so the file never reaches the parser. Comments are invisible to YAML but perfectly visible to the codec. Confirmed on this branch: `yaml.safe_load(open(fn, encoding="gbk"))` raises at position 1331, and the identical file with all `#` comments stripped loads cleanly — every YAML *value* in it is pure ASCII.

So #356 is the right fix and worth landing rather than left open. Transliterating our three dashes would not actually help: the three from #269 would still crash the same box. If you would nonetheless like all six turned into plain hyphens as belt-and-braces, say the word — but it is a workaround for a reader bug, not a fix for it.

Best,
Paul

TomaSusi:
Great, merging now :)

PR #347, opened 2026-08-21
Fast-FFT grid sizing: opt-in rounding, combined with commensurate auto-sampling from PR 274

pzeiger:
Hi!

While running large-scale HAADF simulations on Perlmutter, Claude and I found that a lot of performance is lost when the automatic grid lands on an *FFT-unfriendly* size: `gpts = ceil(extent / sampling)` regularly produces
sizes like 2623 = 43×61, and FFT libraries (FFTW, MKL, cuFFT) only have fast kernels for sizes whose prime factors are all in {2, 3, 5, 7}.

The focus of this PR is making the *default* grid choice do the right thing. As discussed, it is rebased on #274, and `sampling='auto'` now produces grids that are commensurate with the lattice *and* FFT-fast wherever the two are
compatible — commensurability wins when they are not, since aliasing is a physics error while a slow FFT only costs time. For structures with no commensurate grid (rutile, brookite), the fallback now aligns the grid to the structure's translational period, which fixed a latent sub-pixel alignment issue along the way. Explicit numeric sampling keeps its exact current behaviour — the rounding is available there as an opt-in (`grid.round-to-fast-fft`, or `Grid.round_to_fast_fft()`), so no existing script changes results.

Measured on production-scale scans: 3–6× faster on CPU, up to 3× end-to-end on an A100 at float64. (FFT-friendly grids also need far less cuFFT plan workspace at float64, which for our original workload made the difference between OOM and completion — details in the benchmarks below.)

Best,
Paul

-------------

# Claude's summary

zed batch; all measured below; the issue surfaced while diagnosing an OOM in a production 4×A100 HAADF run). CPU FFTs (FFTW, pocketfft, MKL) are also measurably slower on such lengths, so the feature applies to both device paths — it lives in `Grid`, which is device-agnostic, and 7-smooth lengths are optimal for every backend abTEM uses.

## Targets the autogrid branch — one grid, both conditions

As suggested in review, this PR now **targets `autogrid` (#274)** rather than `dev`: it contains only the fast-FFT work and the integration, sitting directly on the living autogrid branch (including its recent slice-thickness fixes, the multi-configuration AtomsEnsemble skip, and the new `test/test_commensurate_potential.py` suite). Merge order becomes: #347 → `autogrid`, then #274 carries everything to `dev`. The integration makes `sampling='auto'` fulfil **both** conditions:

- **Commensurate and fast.** When the GCD search finds a commensurate grid period p, the grid becomes the smallest multiple of p that is 7-smooth and not coarser than the target sampling. A multiple of p is a fast FFT size exactly when p and the multiplier both are — so when p itself contains a prime factor larger than 7, no fast multiple exists and commensurability takes precedence unchanged (aliasing of the nuclear potential is a physics error; a slow FFT only costs time).
- **Incommensurate structures: translational period preserved.** Structures with irrational internal parameters (rutile u = 0.306, brookite) have no exactly commensurate grid, and re-testing exposed a latent constraint: gpts must remain a multiple of the unit-cell repeat count, or symmetry-equivalent atoms in different cells land at different sub-pixel offsets and receive different discretised potentials. The previous `ceil`-derived fallback satisfied this only by coincidence of where `ceil` landed. The fallback now detects the translational period from the atom-plane set directly and picks the best-plane-aligned fast multiple of it. For rutile this beats the raw target size outright: 294 px = 2·3·7² aligns the planes to ~0.01 px (u = 0.306 ≈ 30/98), versus ~0.15 px at the old fallback size.
- **Translation-invariant fallback scoring.** The incommensurate fallback ranks candidate fast sizes by *alignability* — the smallest achievable max plane-to-gridpoint distance over all grid-origin phases, computed from the largest circular gap of the plane residues — which depends only on relative positions. A rigidly shifted structure therefore gets the same grid, preserving the invariant autogrid's own test suite checks (an earlier absolute-position score failed `test_gpts_is_translation_invariant` for rutile/brookite).
- **Composes with the AtomsEnsemble skip:** ensemble (and non-periodic) grids have no commensurability constraint, so their target-derived sizes are simply rounded up to fast lengths — rounding there is free.
- `commensurate_gpts(..., round_to_fast_fft=False)` reproduces #274's behaviour exactly, for direct comparison.

Fast grids are the **default for `sampling='auto'`** — it is a new code path with no existing scripts to preserve — while the plain numeric-sampling path stays strictly opt-in (below).

## What this adds

- **`grid.round-to-fast-fft` config option (default `false`).** When enabled, gpts *derived from a requested sampling* are rounded up to the next 7-smooth length inside `Grid._adjust_gpts`. Guarantees:
  - rounding is **upward only** — the realized sampling is never coarser than requested;
  - **explicitly given `gpts` are never altered** (only the sampling→gpts derivation is touched);
  - **endpoint grids are exempt** (they are not periodic FFT grids);
  - idempotent — already-fast sizes pass through unchanged.

  Example (the case above): off → (2623, 2271); on → (2625, 2304) (= 3·5³·7, 2⁸·3²), sampling 0.0499/0.0493 Å, +1.5 % pixels. `probe.grid.match(potential)` propagates the rounded grid consistently.

- **FFT-fast commensurate `sampling='auto'`** (see previous section): commensurate with the lattice *and* 7-smooth wherever the two are compatible, with a principled fallback for incommensurate structures.

- **`Grid.round_to_fast_fft()`** method for explicit one-shot use, mirroring `round_to_power()`. Note `round_to_power`'s docstring promises sizes "factored into small primes" but the implementation produces *pure* powers of a single base (2623 → 4096, +56 % pixels vs. +0.08 % for 2625); the new method provides what that docstring describes. A follow-up could deprecate or redirect `round_to_power`.

- **`is_fast_fft_size()` / `next_fast_fft_size()`** helpers in `abtem.core.fft`, plus a one-time `UserWarning` when a large (≥ 1024²) GPU FFT runs on a Bluestein-fallback shape, suggesting the nearest fast grid. The helpers commit is shared verbatim with the `multi-gpu-hardening` branch (#346) — whichever merges first, the other rebases to an empty commit.

## Review round 1 (addressed)

- **Why not `scipy.fft.next_fast_len`?** Deliberate divergence, now documented in code and pinned by a test: scipy targets pocketfft, whose kernels go up to radix 11, so it returns **11-smooth** lengths (`next_fast_len(121) == 121 == 11²`, `next_fast_len(4619) == 4620 = 2²·3·5·7·11`). cuFFT documents optimized kernels only for `2^a·3^b·5^c·7^d`, so factor-11 grids fall off exactly the GPU fast path this feature protects. {2, 3, 5, 7} is the intersection guaranteed fast on every backend abTEM uses. CuPy independently made the same call: `cupyx.scipy.fft.next_fast_len` is 7-smooth, agrees with our helper on every probed value, and its docstring notes the deliberate difference from scipy; a test pins the agreement (skipped without cupy).
- **Warn-once race:** fixed with a lock around the check-then-add on the warned-shapes set (the warning itself stays outside); barrier-synchronized 8-thread test added.
- **O(n² log n) period search:** reworked. A shift mapping the plane set onto itself must map the first plane onto some plane, so candidates are only the differences to the first plane, filtered to near-integer L/s and tested smallest-first (first hit = largest m). Typically O(n log n); a 2400-plane incommensurate supercell now takes 1.9 ms.

## Review round 2 (addressed)

All six findings confirmed; none refuted. Two were worse than reported, one was benign, and three more problems in the surrounding code turned up while fixing them.

- **Periodic wrap in the plane matching** (finding 1) — real, and reachable on ordinary crystals. Not at the low end (the caller's `x > L - tolerance -> 0` snap keeps input planes away from L), but at the high end, because that snap protects the *input* planes and not the *shifted* ones — and `orthogonalize_cell` returns rutile's L/2 plane one ulp short. Measured through the public API: rutile in `plane='xz'` gave gpts (98,60) unshifted vs (108,60) shifted by 1e-6 Å, breaking the documented translation invariance, and calcite 6×5×1 came out with an odd gpts against a true period of 450, spreading symmetry-equivalent atoms over 0.44 px instead of 2e-6 px. Fixed by indexing modulo n; checked against an independent brute-force periodic oracle over ~1.3 M invariance evaluations with zero disagreements.
- **The 64-candidate cap** (finding 3) — removed outright rather than made graceful. Reachable on tiled disordered cells (an amorphous support film repeated to cover the field of view). No cap is needed: the candidate count is bounded by how many distinct `L/m` are more than `tolerance` apart, not by the number of planes, so the search is sub-quadratic in atom count at fixed cell size — worst case constructible inside a plausible cell is 167 ms, against ~1.2 s just to build the potential array. Each distinct m is now tested once at the **exact** shift `L/m` rather than at the difference that suggested it, which is both faster and more correct: the difference carries the plane's float noise, and an off-period difference rounding to m could suppress the genuine shift. (Findings 1 and 3 compound: on calcite the wrap bug rejected every genuine candidate, the search then exhausted the cap, and the result was "no period" at all.)
- **The config governed only half the feature** (finding 2) — the option is now tri-state: `'auto'` (new default) rounds only grids abTEM derives itself, `True` additionally rounds gpts derived from a numeric sampling, `False` never rounds. Numeric-sampling behaviour is unchanged in the default mode, and `False` reproduces the plain commensurate grid exactly. `'auto'` is truthy, so every read goes through a mode helper and an unrecognized value raises instead of silently meaning "on".
- **The warning ignored leading transformed axes** (finding 5) — worse than reported: both the fastness check *and* the size gate were computed from `shape[-2:]`, so a slow leading axis was invisible on the 3D structure-factor path, whose grid is essentially never 7-smooth. Lengths now come from the function and its `axes`/`axis`/`s` arguments; `s` entries of `None` are tolerated (a raise here would break the transform, not just the diagnostic), and the "set gpts" remedy is dropped where it does not apply.
- **Warn-cache hygiene** (finding 6) — real but benign (a realistic GPU scan produced one 130-byte entry). Fixed anyway: the size and fastness checks now precede the lock, so only shapes that actually warn are recorded and a fast-grid run never takes the lock.
- **`round_to_power`** (finding 4) — a discoverability trap rather than a defect: zero callers, and its docstring promised what `round_to_fast_fft` actually does. Behaviour left alone (dead but public); the docstring now describes it accurately, cross-references the new method, and notes the float-`log` artifact that overshoots for 125, 15625 and 16807. Deprecating it is a maintainer decision, not this PR's.

Found while fixing the above, not raised in review:

- The fallback's documented +12 % window was **not enforced** — the candidate that terminated the search was still scored and, being the largest, usually won: rutile took 1.17× the target size (+38 % pixels) for a 2.6 pm alignment gain. Now tested before scoring, with candidates compared by misalignment in Ångström rather than in fractions of a grid spacing.
- Grids that were **already fast** were enlarged anyway (Cu fcc 6×6: 432 = 2⁴·3³ → 480, +23.5 % pixels, equivalent-atom spread at float noise either way). The nearest commensurate grid is now kept when already fast, so Si, Cu and MoS₂ keep autogrid's own grid; Au (164 = 2²·41) and SrTiO₃ (234 = 2·3²·13) still move. Rounding the multiplier **up** in the remaining cases keeps the grid from ever coming out coarser than with rounding off.
- `grid.round-to-fast-fft: True` rounded **GridScan probe positions**, changing the probe count and scan step for a grid that is never Fourier transformed (and disagreeing with `LineScan`, which does not use `Grid`). Grids now carry an `fft_grid` flag; `GridScan` opts out.

## Review round 3 (addressed)

- **The fallback's overshoot window was not a bound.** Moving the guard cannot fix it: candidates ascend, so if the first fast multiple of the period `m` overshoots, every one does and there is no in-window candidate to find (`m=64` against a target of 72 lands on 128, +78 %). This branch now gets the same escape the other one already had — when no fast multiple fits the window, keep the period constraint and drop fastness. The documented window is a real bound again, and the two branches behave alike. As with that pre-existing branch, the escape can land marginally coarser than the target; the alternative is +78 % pixels.
- **The 3D structure-factor transform escaped the diagnostic.** Confirmed, but it cannot simply be routed through `abtem.core.fft.ifftn`: the FFTW backend behind that wrapper hardcodes `axes=(-2, -1)`, so wrapped `fftn`/`ifftn` silently perform a **2D** transform on 3D input under the default `fft: fftw` config (verified identical on `origin/autogrid` — pre-existing, not introduced here). The transform therefore stays on `xp.fft` and requests the diagnostic explicitly through a GPU-only helper. The wrapper bug is left alone as out of scope for this PR.
- **`Grid.round_to_fast_fft()` ignored `fft_grid`.** The automatic path already skipped grids that are never Fourier transformed; the explicit method now does too.

## Defaults: opt-in for numeric sampling, on for `sampling='auto'`

Rounding numeric-sampling grids by default would change gpts — and therefore results bit-for-bit — for every existing sampling-specified script: regression baselines, published numbers, and shape-compatibility with stored data (zarr archives, S-matrix caches) would all silently shift. Off-by-default preserves exact backward compatibility there; the GPU warning steers users toward enabling it. `sampling='auto'`, by contrast, is new in this PR series and ships fast-by-default from the start. With 1.1 already carrying a numerics-breaking propagator change (#298), flipping the numeric-sampling default could be considered too, but that is left as a deliberate follow-up decision.

## Benchmarks

Measured on an AMD Strix Halo dev box: CPU = FFTW, single thread, `FFTW_ESTIMATE` planning (autotuning excluded); GPU = Radeon 8050S iGPU via CuPy 13.5.1 / ROCm hipFFT. `float32`/`complex64`; median of repeats after warm-up. "Small" is 10 Å at 0.03 Å sampling, "large" is the 131.1 × 113.5 Å production cell at 0.05 Å. The end-to-end case is a 4-slice HAADF probe scan (60 keV, 31.5 mrad, annular 70–200 mrad).

| benchmark | device | ceil gpts | fast gpts | ceil | fast | speedup | extra pixels |
|---|---|---|---|---|---|---|---|
| fft2+ifft2, batch 8 | CPU | 334² | 336² | 44.0 ms | 6.4 ms | **6.9×** | +1.2 % |
| fft2+ifft2, batch 4 | CPU | 2623×2271 | 2625×2304 | 1486.6 ms | 421.4 ms | **3.5×** | +1.5 % |
| HAADF scan, 64 pos | CPU | 334² | 336² | 2328.5 ms | 591.8 ms | **3.9×** | +1.2 % |
| HAADF scan, 4 pos | CPU | 2623×2271 | 2625×2304 | 10310.1 ms | 3334.6 ms | **3.1×** | +1.5 % |
| fft2+ifft2, batch 8 | GPU | 334² | 336² | 3.0 ms | 1.5 ms | **2.0×** | +1.2 % |
| fft2+ifft2, batch 4 | GPU | 2623×2271 | 2625×2304 | 140.1 ms | 99.8 ms | **1.4×** | +1.5 % |
| HAADF scan, 64 pos | GPU | 334² | 336² | 179.1 ms | 97.2 ms | **1.8×** | +1.2 % |
| HAADF scan, 4 pos | GPU | 2623×2271 | 2625×2304 | 1199.8 ms | 892.1 ms | **1.3×** | +1.5 % |

The same benchmark on a Perlmutter A100 (cuFFT, CUDA 40 GB):

| benchmark | ceil gpts | fast gpts | ceil | fast | speedup |
|---|---|---|---|---|---|
| fft2+ifft2, batch 8 | 334² | 336² | 0.2 ms | 0.2 ms | (launch-bound) |
| fft2+ifft2, batch 4 | 2623×2271 | 2625×2304 | 5.1 ms | 2.8 ms | **1.8×** |
| HAADF scan, 64 pos | 334² | 336² | 32.6 ms | 31.5 ms | 1.04× |
| HAADF scan, 4 pos | 2623×2271 | 2625×2304 | 97.7 ms | 83.7 ms | 1.17× |

| memory (after one round-trip) | data | pool | ratio |
|---|---|---|---|
| ceil 2623×2271, batch 4 | 190.6 MB | 953.1 MB | 5.0× |
| fast 2625×2304, batch 4 | 193.5 MB | 774.1 MB | 4.0× |

### Production-scale memory and speed (float64, serial 40 GB A100)

The decisive test of the memory claim: the full production workload (101,871-position HAADF scan, float64) with *only the grid changed*, run on plain `dev` and on the `multi-gpu-hardening` branch (#346, whose bounded plan cache runs the oversized ceil-grid plan uncached):

| branch | ceil (2623×2271) | fast (2625×2304) |
|---|---|---|
| dev | **OOM** (pool 41.2 GB — the original production crash) | **completes: 21.43 GB peak**, 1041 s |
| #346 branch | completes: 31.32 GB peak, 3095 s | 21.43 GB peak, **1027 s** |

Isolated plan workspace for a 16-probe batch: **11.0 GB (ceil) vs 1.55 GB (fast)** at complex128 — 7.1×; 0.76 GB vs ≈0 at complex64. The ceil-vs-fast peak difference (31.3 → 21.4 GB) is almost exactly that transient 11 GB workspace. End-to-end at this scale the fast grid is also **3.0× faster** — well above the small-benchmark GPU ratios, because the Bluestein planning cost compounds over ~6400 batches. Fast-grid results are bit-identical across branches (same checksum on dev and #346).

Reading: the CPU gains (3–6×) are the headline and reproduce on both machines. On the A100 the raw large-grid FFT gap is 1.8× — larger than on ROCm, cuFFT's Bluestein being the costlier fallback — while the end-to-end rows are diluted at this benchmark size because non-FFT overheads dominate on so fast a GPU (the 4-position scan spends ~10–20 ms of its 98 ms in FFTs); production-scale scans amortize toward the raw ratio. On memory, the effect at complex64 is real but modest: the round-trip holds ~3× data in live arrays, so the retained plan workspace is ≈ 2× data for the Bluestein shape vs ≈ 1× for the fast shape (~20 % pool reduction here). At complex128 it is anything but modest — see the production-scale table above: 7× plan workspace, 32 % peak reduction, and OOM-vs-completes on plain dev. CPU numbers use `FFTW_ESTIMATE`; `FFTW_MEASURE` narrows the CPU gap somewhat in exchange for per-shape planning time.

## Testing

- Autogrid's own `test/test_commensurate_potential.py` passes in full on the combined grids: all 19 lattice types (hexagonal included — coverage the old root-level scripts skipped), translation invariance (rutile/brookite via the translational-period fallback), slabs with vacuum, supercell sizes, parametrizations, axis orientations, and the ensemble behavior (expected gpts adapted to the fast-rounded values, with a comment).
- New `test/test_commensurate_gpts.py`: combined commensurate+fast conditions, upward-only sampling, smallest-fast-multiple selection, `round_to_fast_fft=False` equivalence to #274, both fallback regimes (fast translational period → fast multiple; non-smooth period → commensurability wins), fallback timing guard, end-to-end `Potential(sampling='auto')` for Si and Au.
- New unit tests in `test_grid.py` (default-off regression, refinement direction, already-fast passthrough, explicit-gpts immunity, endpoint exemption, sampling-setter path, the explicit method) and `test_fft.py` (helpers, warning, thread-safety, scipy-divergence and cupyx-agreement pins).
- The suite passes with `grid.round-to-fast-fft` set to `false` and to `true` alike — it previously failed both ways, because several tests read the ambient config instead of pinning it.
- Every new test was checked to actually **fail** with the fix it guards reverted. Three that could not fail were rewritten, including round 1's thread-safety test, which passed against the unlocked code and now observes 8 warnings without the lock and 1 with it.
- Full suite on the current base: no new failures — the failing set is byte-identical to unmodified `autogrid` (GPU/ROCm and GPAW-version cases that fail there too).

### DIFF
--- a/abtem/bloch/dynamical.py
+++ b/abtem/bloch/dynamical.py
@@ -47,7 +47,7 @@ from abtem.core.constants import kappa
 from abtem.core.diagnostics import TqdmWrapper
 from abtem.core.energy import energy2sigma, energy2wavelength
 from abtem.core.ensemble import Ensemble, _wrap_with_array, unpack_blockwise_args
-from abtem.core.fft import fft_interpolate
+from abtem.core.fft import fft_interpolate, warn_if_slow_gpu_fft
 from abtem.core.grid import Grid
 from abtem.core.utils import CopyMixin, get_dtype
 from abtem.distributions import BaseDistribution, validate_distribution
@@ -282,6 +282,12 @@ def structure_factor_to_potential(
     """
     xp = get_array_module(structure_factor)
     structure_factor = structure_factor_1d_to_3d(structure_factor, hkl, gpts)
+    # Deliberately xp.fft rather than abtem.core.fft.ifftn: the FFTW backend
+    # behind that wrapper only ever transforms the trailing two axes, so it
+    # would silently turn this 3D transform into a 2D one on CPU. The slow-FFT
+    # diagnostic is requested explicitly instead -- this grid follows from
+    # g_max and the cell, so it is essentially never a fast length.
+    warn_if_slow_gpu_fft(structure_factor, "ifftn")
     potential = xp.fft.ifftn(structure_factor)
     potential = potential * np.prod(potential.shape) / kappa
     potential -= potential.min()
--- a/abtem/core/abtem.yaml
+++ b/abtem/core/abtem.yaml
@@ -43,6 +43,13 @@ fftw:
   planning_timelimit: 60
   # Whether to allow falling back to not using wisdom if the cache fails
   allow_fallback: true
+grid:
+  # Round gpts derived from a requested sampling up to the next fast FFT length
+  # (all prime factors in {2, 3, 5, 7}). The realized sampling then becomes at
+  # most the requested one, never coarser. Explicitly given gpts are never
+  # altered. Fast lengths avoid the slow, memory-hungry Bluestein FFT fallback
+  # on GPU and speed up CPU FFTs as well.
+  round-to-fast-fft: false
 warnings:
   # Show the dask warning about the blockwise performance when the number are increased dramatically
   dask-blockwise-performance: false
--- a/abtem/core/abtem.yaml
+++ b/abtem/core/abtem.yaml
@@ -44,12 +44,15 @@ fftw:
   # Whether to allow falling back to not using wisdom if the cache fails
   allow_fallback: true
 grid:
-  # Round gpts derived from a requested sampling up to the next fast FFT length
-  # (all prime factors in {2, 3, 5, 7}). The realized sampling then becomes at
-  # most the requested one, never coarser. Explicitly given gpts are never
-  # altered. Fast lengths avoid the slow, memory-hungry Bluestein FFT fallback
-  # on GPU and speed up CPU FFTs as well.
-  round-to-fast-fft: false
+  # Round automatically derived gpts up to the next fast FFT length (all prime
+  # factors in {2, 3, 5, 7}). Fast lengths avoid the slow, memory-hungry
+  # Bluestein FFT fallback on GPU and speed up CPU FFTs as well. Rounding is
+  # upward only, so a grid derived from a numeric sampling never comes out
+  # coarser than requested. gpts given explicitly are never altered.
+  #   'auto'  round only grids abTEM derives on its own (sampling='auto')
+  #   true    additionally round gpts derived from a numeric sampling
+  #   false   never round
+  round-to-fast-fft: auto
 warnings:
   # Show the dask warning about the blockwise performance when the number are increased dramatically
   dask-blockwise-performance: false
--- a/abtem/core/fft.py
+++ b/abtem/core/fft.py
@@ -182,7 +182,8 @@ def _warn_slow_fft_size(shape, func_name: str = "fft2", kwargs=None):
     # user sets with gpts; for anything else (e.g. the 3D structure-factor
     # grid, which follows from g_max and the cell) that advice would be wrong.
     remedy = (
-        ", e.g. by setting gpts explicitly instead of the sampling"
+        ", e.g. with abtem.config.set({'grid.round-to-fast-fft': True}) or by "
+        "setting gpts explicitly instead of the sampling"
         if len(lengths) == 2 and len(shape) >= 2 and tuple(shape[-2:]) == lengths
         else ""
     )
--- a/abtem/core/fft.py
+++ b/abtem/core/fft.py
@@ -199,6 +199,21 @@ def _warn_slow_fft_size(shape, func_name: str = "fft2", kwargs=None):
     )
 
 
+def warn_if_slow_gpu_fft(x, func_name: str = "fft2", **kwargs):
+    """
+    Emit the slow-FFT diagnostic for a transform run outside ``_fft_dispatch``.
+
+    A few transforms call ``xp.fft`` directly rather than through the wrappers
+    in this module -- ``structure_factor_to_potential`` does, because the FFTW
+    backend here only ever transforms the trailing two axes and would silently
+    turn its 3D transform into a 2D one. They still deserve the diagnostic, so
+    they can call this alongside. Only GPU arrays are considered: the Bluestein
+    workspace this warns about is a cuFFT concern.
+    """
+    if cp is not None and isinstance(x, cp.ndarray):
+        _warn_slow_fft_size(x.shape, func_name, kwargs)
+
+
 def _raise_fft_lib_not_present(lib_name: str):
     raise RuntimeError(
         f"FFT library {lib_name} not present. Install this package or change the FFT"
--- a/abtem/core/fft.py
+++ b/abtem/core/fft.py
@@ -1,5 +1,6 @@
 """Module for handling Fourier transforms and convolution in *ab*TEM."""
 
+import threading
 import warnings
 from itertools import product as _product
 from typing import Tuple, TypeVar, overload
@@ -103,6 +104,9 @@ def next_fast_fft_size(n: int) -> int:
 
 
 _warned_slow_fft_shapes: set = set()
+# _fft_dispatch runs concurrently in dask worker threads, so the check-then-add
+# on the shape set must be atomic or the same shape can warn more than once.
+_warned_slow_fft_shapes_lock = threading.Lock()
 
 # Bluestein overhead only matters for large transforms; small odd-sized arrays
 # (e.g. interpolated measurements) are not worth a warning.
@@ -112,9 +116,12 @@ _SLOW_FFT_WARN_MIN_ELEMENTS = 1024 * 1024
 def _warn_slow_fft_size(shape):
     """Warn once per large 2D transform shape whose lengths force Bluestein."""
     dims = tuple(int(n) for n in shape[-2:])
-    if len(dims) < 2 or dims in _warned_slow_fft_shapes:
+    if len(dims) < 2:
         return
-    _warned_slow_fft_shapes.add(dims)
+    with _warned_slow_fft_shapes_lock:
+        if dims in _warned_slow_fft_shapes:
+            return
+        _warned_slow_fft_shapes.add(dims)
     if dims[0] * dims[1] < _SLOW_FFT_WARN_MIN_ELEMENTS:
         return
     if all(is_fast_fft_size(n) for n in dims):
--- a/abtem/core/fft.py
+++ b/abtem/core/fft.py
@@ -37,6 +37,90 @@ except ImportError:
     cp = None
 
 
+_FAST_FFT_PRIMES = (2, 3, 5, 7)
+
+
+def is_fast_fft_size(n: int) -> bool:
+    """
+    Whether an FFT of length ``n`` runs on fast radix kernels.
+
+    FFT libraries only ship optimized kernels for lengths whose prime factors
+    are small (2, 3, 5 and 7 are supported everywhere). A length with a larger
+    prime factor triggers a generic fallback -- on cuFFT the Bluestein
+    algorithm, which pads internally to a power of two, costing several times
+    the arithmetic and, on GPU, a workspace of several times the transform
+    size.
+
+    Parameters
+    ----------
+    n : int
+        The transform length.
+
+    Returns
+    -------
+    bool
+        True if ``n`` factorizes completely into 2, 3, 5 and 7.
+    """
+    n = int(n)
+    if n < 1:
+        return False
+    for p in _FAST_FFT_PRIMES:
+        while n % p == 0:
+            n //= p
+    return n == 1
+
+
+def next_fast_fft_size(n: int) -> int:
+    """
+    The smallest length >= ``n`` whose prime factors are all in {2, 3, 5, 7}.
+
+    Useful for choosing grid sizes (``gpts``) that avoid the slow
+    large-workspace Bluestein fallback on GPU.
+
+    Parameters
+    ----------
+    n : int
+        The minimum transform length.
+
+    Returns
+    -------
+    int
+        The next fast transform length.
+    """
+    n = max(1, int(n))
+    while not is_fast_fft_size(n):
+        n += 1
+    return n
+
+
+_warned_slow_fft_shapes: set = set()
+
+# Bluestein overhead only matters for large transforms; small odd-sized arrays
+# (e.g. interpolated measurements) are not worth a warning.
+_SLOW_FFT_WARN_MIN_ELEMENTS = 1024 * 1024
+
+
+def _warn_slow_fft_size(shape):
+    """Warn once per large 2D transform shape whose lengths force Bluestein."""
+    dims = tuple(int(n) for n in shape[-2:])
+    if len(dims) < 2 or dims in _warned_slow_fft_shapes:
+        return
+    _warned_slow_fft_shapes.add(dims)
+    if dims[0] * dims[1] < _SLOW_FFT_WARN_MIN_ELEMENTS:
+        return
+    if all(is_fast_fft_size(n) for n in dims):
+        return
+    suggestion = " x ".join(str(next_fast_fft_size(n)) for n in dims)
+    warnings.warn(
+        f"FFT size {dims[0]} x {dims[1]} contains prime factors larger than 7; "
+        "cuFFT falls back to the Bluestein algorithm, which is several times "
+        "slower and allocates a workspace of several times the array size. "
+        f"Consider adjusting the grid to {suggestion}, e.g. by setting gpts "
+        "explicitly instead of the sampling.",
+        UserWarning,
+    )
+
+
 def _raise_fft_lib_not_present(lib_name: str):
     raise RuntimeError(
         f"FFT library {lib_name} not present. Install this package or change the FFT"
@@ -262,6 +346,7 @@ def _fft_dispatch(
 
     if isinstance(x, cp.ndarray):
         _configure_cufft_cache()
+        _warn_slow_fft_size(x.shape)
         return getattr(cp.fft, func_name)(x, **kwargs)
 
 
--- a/abtem/core/fft.py
+++ b/abtem/core/fft.py
@@ -104,14 +104,6 @@ def next_fast_fft_size(n: int) -> int:
     return n
 
 
-def _prev_fast_fft_size(n: int) -> int:
-    """The largest length <= ``n`` whose prime factors are all in {2, 3, 5, 7}."""
-    n = max(1, int(n))
-    while not is_fast_fft_size(n):
-        n -= 1
-    return n
-
-
 _warned_slow_fft_shapes: set = set()
 # _fft_dispatch runs concurrently in dask worker threads, so the check-then-add
 # on the shape set must be atomic or the same shape can warn more than once.
@@ -134,17 +126,27 @@ def _transform_lengths(shape, func_name: str = "fft2", kwargs=None) -> tuple[int
     """
     kwargs = kwargs or {}
     axes = kwargs.get("axes")
+    # An explicit `s` overrides the array lengths (None entries keep theirs).
+    s = kwargs.get("s")
 
     if axes is None:
         if func_name.endswith("fftn"):
-            axes = tuple(range(len(shape)))
+            # numpy and cupy apply `s` to the TRAILING len(s) axes when no axes
+            # are given, so deriving the axes from the array rank alone would
+            # pair `s` with the wrong ones.
+            count = len(s) if s is not None else len(shape)
+            axes = tuple(range(len(shape) - count, len(shape)))
         elif func_name.endswith("fft2"):
             axes = (-2, -1)
         else:
             axes = (kwargs.get("axis", -1),)
 
-    # An explicit `s` overrides the array lengths (None entries keep theirs).
-    s = kwargs.get("s")
+    # Drop axes the array does not have. cupy's fft2 quietly reduces the axes
+    # for arrays of fewer than two dimensions, and this runs ahead of every GPU
+    # transform: raising here would break a transform that would otherwise
+    # succeed, rather than merely skipping a diagnostic.
+    ndim = len(shape)
+    axes = tuple(axis for axis in axes if -ndim <= axis < ndim)
 
     lengths = []
     for i, axis in enumerate(axes):
@@ -182,8 +184,9 @@ def _warn_slow_fft_size(shape, func_name: str = "fft2", kwargs=None):
     # user sets with gpts; for anything else (e.g. the 3D structure-factor
     # grid, which follows from g_max and the cell) that advice would be wrong.
     remedy = (
-        ", e.g. with abtem.config.set({'grid.round-to-fast-fft': True}) or by "
-        "setting gpts explicitly instead of the sampling"
+        ", e.g. by setting gpts explicitly, or -- if this grid was derived from "
+        "a numeric sampling -- with "
+        "abtem.config.set({'grid.round-to-fast-fft': True})"
         if len(lengths) == 2 and len(shape) >= 2 and tuple(shape[-2:]) == lengths
         else ""
     )
--- a/abtem/core/fft.py
+++ b/abtem/core/fft.py
@@ -37,6 +37,13 @@ except ImportError:
     cp = None
 
 
+# Deliberately NOT scipy.fft.next_fast_len's set: scipy targets pocketfft,
+# whose kernels go up to radix 11, so next_fast_len returns 11-smooth lengths
+# (next_fast_len(121) == 121 == 11**2). cuFFT documents optimized kernels only
+# for sizes of the form 2^a * 3^b * 5^c * 7^d, so a factor-11 length falls off
+# exactly the GPU fast path this module exists to protect. {2, 3, 5, 7} is the
+# intersection guaranteed fast on every backend abTEM uses (FFTW, pocketfft,
+# MKL, cuFFT, hipFFT).
 _FAST_FFT_PRIMES = (2, 3, 5, 7)
 
 
@@ -75,7 +82,9 @@ def next_fast_fft_size(n: int) -> int:
     The smallest length >= ``n`` whose prime factors are all in {2, 3, 5, 7}.
 
     Useful for choosing grid sizes (``gpts``) that avoid the slow
-    large-workspace Bluestein fallback on GPU.
+    large-workspace Bluestein fallback on GPU. Stricter than
+    ``scipy.fft.next_fast_len``, which returns 11-smooth lengths (fast for
+    pocketfft on CPU, but off cuFFT's documented fast path).
 
     Parameters
     ----------
--- a/abtem/core/fft.py
+++ b/abtem/core/fft.py
@@ -1,5 +1,6 @@
 """Module for handling Fourier transforms and convolution in *ab*TEM."""
 
+import math
 import threading
 import warnings
 from itertools import product as _product
@@ -113,26 +114,75 @@ _warned_slow_fft_shapes_lock = threading.Lock()
 _SLOW_FFT_WARN_MIN_ELEMENTS = 1024 * 1024
 
 
-def _warn_slow_fft_size(shape):
-    """Warn once per large 2D transform shape whose lengths force Bluestein."""
-    dims = tuple(int(n) for n in shape[-2:])
-    if len(dims) < 2:
+def _transform_lengths(shape, func_name: str = "fft2", kwargs=None) -> tuple[int, ...]:
+    """
+    The array lengths a transform actually acts on.
+
+    Only the transformed axes matter for the Bluestein fallback: batch axes are
+    looped over, whatever their length. Which axes those are depends on the
+    function -- ``fftn`` transforms all of them unless ``axes`` says otherwise,
+    ``fft2`` the trailing two, ``fft`` a single one -- so reading the trailing
+    two axes unconditionally misreports every non-2D transform.
+    """
+    kwargs = kwargs or {}
+    axes = kwargs.get("axes")
+
+    if axes is None:
+        if func_name.endswith("fftn"):
+            axes = tuple(range(len(shape)))
+        elif func_name.endswith("fft2"):
+            axes = (-2, -1)
+        else:
+            axes = (kwargs.get("axis", -1),)
+
+    # An explicit `s` overrides the array lengths (None entries keep theirs).
+    s = kwargs.get("s")
+
+    lengths = []
+    for i, axis in enumerate(axes):
+        length = shape[axis]
+        if s is not None and i < len(s) and s[i] is not None:
+            length = s[i]
+        lengths.append(int(length))
+
+    return tuple(lengths)
+
+
+def _warn_slow_fft_size(shape, func_name: str = "fft2", kwargs=None):
+    """Warn once per large transform whose transformed lengths force Bluestein."""
+    lengths = _transform_lengths(shape, func_name, kwargs)
+
+    if not lengths:
         return
+
+    # Test cheaply first, and record only shapes that actually warn: the
+    # memo then holds at most one entry per warning ever emitted, and a run
+    # whose grid is fast never touches the lock on the FFT hot path.
+    if math.prod(lengths) < _SLOW_FFT_WARN_MIN_ELEMENTS:
+        return
+    if all(is_fast_fft_size(n) for n in lengths):
+        return
+
     with _warned_slow_fft_shapes_lock:
-        if dims in _warned_slow_fft_shapes:
+        if lengths in _warned_slow_fft_shapes:
             return
-        _warned_slow_fft_shapes.add(dims)
-    if dims[0] * dims[1] < _SLOW_FFT_WARN_MIN_ELEMENTS:
-        return
-    if all(is_fast_fft_size(n) for n in dims):
-        return
-    suggestion = " x ".join(str(next_fast_fft_size(n)) for n in dims)
+        _warned_slow_fft_shapes.add(lengths)
+
+    shown = " x ".join(str(n) for n in lengths)
+    suggestion = " x ".join(str(next_fast_fft_size(n)) for n in lengths)
+    # Only a 2D transform of the trailing axes is the wave-function grid the
+    # user sets with gpts; for anything else (e.g. the 3D structure-factor
+    # grid, which follows from g_max and the cell) that advice would be wrong.
+    remedy = (
+        ", e.g. by setting gpts explicitly instead of the sampling"
+        if len(lengths) == 2 and len(shape) >= 2 and tuple(shape[-2:]) == lengths
+        else ""
+    )
     warnings.warn(
-        f"FFT size {dims[0]} x {dims[1]} contains prime factors larger than 7; "
+        f"FFT size {shown} contains prime factors larger than 7; "
         "cuFFT falls back to the Bluestein algorithm, which is several times "
         "slower and allocates a workspace of several times the array size. "
-        f"Consider adjusting the grid to {suggestion}, e.g. by setting gpts "
-        "explicitly instead of the sampling.",
+        f"Consider adjusting the transformed grid to {suggestion}{remedy}.",
         UserWarning,
     )
 
@@ -362,7 +412,7 @@ def _fft_dispatch(
 
     if isinstance(x, cp.ndarray):
         _configure_cufft_cache()
-        _warn_slow_fft_size(x.shape)
+        _warn_slow_fft_size(x.shape, func_name, kwargs)
         return getattr(cp.fft, func_name)(x, **kwargs)
 
 
--- a/abtem/core/fft.py
+++ b/abtem/core/fft.py
@@ -104,6 +104,14 @@ def next_fast_fft_size(n: int) -> int:
     return n
 
 
+def _prev_fast_fft_size(n: int) -> int:
+    """The largest length <= ``n`` whose prime factors are all in {2, 3, 5, 7}."""
+    n = max(1, int(n))
+    while not is_fast_fft_size(n):
+        n -= 1
+    return n
+
+
 _warned_slow_fft_shapes: set = set()
 # _fft_dispatch runs concurrently in dask worker threads, so the check-then-add
 # on the shape set must be atomic or the same shape can warn more than once.
--- a/abtem/core/grid.py
+++ b/abtem/core/grid.py
@@ -14,6 +14,43 @@ from abtem.core.backend import device_name_from_array_module, get_array_module
 from abtem.core.utils import CopyMixin, EqualityMixin, get_dtype
 
 
+def _fast_fft_rounding_mode() -> str:
+    """
+    Normalize the ``grid.round-to-fast-fft`` configuration to a mode string.
+
+    Returns
+    -------
+    str
+        ``'never'``, ``'auto'`` (only grids abTEM derives on its own) or
+        ``'always'`` (additionally grids derived from a numeric sampling).
+    """
+    value = config.get("grid.round-to-fast-fft", "auto")
+
+    if value is True:
+        return "always"
+    if value is False:
+        return "never"
+    if isinstance(value, str) and value.lower() == "auto":
+        return "auto"
+
+    raise ValueError(
+        "configuration 'grid.round-to-fast-fft' must be True, False or 'auto', "
+        f"got {value!r}"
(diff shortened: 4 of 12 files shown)


TomaSusi:
Thanks Paul! I was aware that powers of two (or some of the other factors) would be particularly good, but I did not except such big differences – I wonder if we’ve done some of the initial testing with intel’s MKL… in any case, AMD and plain pyFFT are more common and what we should optimize for.

This work quite heavily intersects with another sampling issue that users often stumble upon: having the sampling not be commensurate with the atoms can lead to fairly big differences due to aliasing of the nuclear Coulomb potential.

To address that, I have the autogrid PR that is pretty much ready to merge. I suggest you rebase this addition on that branch and we try to fulfill both conditions, even if it brings the grid spacing a bit further away from the requested sampling.

https://github.com/abTEM/abTEM/pull/274

pzeiger:
Hi @TomaSusi!

Yes, I saw your work on the commensurate sampling and also on supporting non-orthogonal supercells. I think combining all of these improvements can be a very nice improvement. I suppose that the optimal grid choice gets much simpler for non-orthogonal cells in general, since the grid is observing more symmetries of the structure than an orthogonal grid does. So I think the combination of this PR with the other two will work out well, but I think orthogonal cells might remain a headache if one wants to combine all options.

The safest thing is probably to keep the current behavior for now and test on a bunch of different settings how well the three improvements work together for non-orthogonal and orthogonal cells. And then one could potentially contemplate changing the default behavior for *ab*TEM v2.0 to obey the best working combination for the different grid settings.

For now I will rebase this on your work in #274 and double re-test.

Best,
Paul

TomaSusi:
I already decided that the next version on dev is anyways going to be 1.1, and it _will_ contain a numerics-breaking change as we switch the default propagator to the exact one:
https://github.com/abTEM/abTEM/pull/298

So I'm very open to making clearly justified improvements even if we lose exact numerical reproducibility between 1.0.x and 1.1.x

pzeiger:
Oh, I see! I guess we can maybe fold it all in in such case =)

pzeiger:
Hi @TomaSusi!

Claude and I have now rebased this on latest `dev` and included also your PR #274, so this PR now contains both features plus the combination.

I have updated the PR text with new findings and the summary is as follows:
- `sampling='auto'` now tries to fulfill both conditions, commensurability and *FFT-friendliness*. When the commensurate search finds a period p, the grid becomes the smallest multiple of p that is also 7-smooth (meaning all prime factors are ≤ 7) and not coarser than the target. Since a multiple of p is a fast FFT size exactly when p and the multiplier both are, the case where p itself contains a prime factor larger than 7 is cleanly decided: no fast multiple exists, and commensurability takes precedence unchanged.
- all of your three verification suites pass as well as new pytest coverage in `test/test_commensurate_gpts.py` for the combined conditions and both fallback regimes, and the existing grid/fft/potential tests. Calling `commensurate_gpts(..., round_to_fast_fft=False)` reproduces the previous autogrid behaviour exactly, so the two are easy to compare.
- re-testing turned up something interesting in the fallback path for structures that have no exactly commensurate grid (irrational internal parameters, e.g. rutile, brookite). Naively rounding the fallback size to a fast length broke your verification scripts, and digging into why revealed that the symmetry-equivalence they check actually requires gpts to stay a multiple of the unit-cell repeat count — otherwise equivalent atoms in different cells land at different sub-pixel offsets and see different discretised potentials. The previous `ceil`-derived fallback satisfied this only by coincidence of where `ceil` landed. The fallback now detects that translational period from the plane set directly and picks the best-plane-aligned fast multiple of it. Nice side effect: for rutile this beats the raw target size outright — 294 = 2·3·7² aligns the planes to ~0.01 px (u = 0.306 ≈ 30/98), versus ~0.15 px before. You may want to consider the same constraint for the fallback in #274 independently of the FFT question.

I think also that my previous comment about the orthogonal cells was maybe too defensive and I think Claude has solved my issue through the decision logic that p needs to be 7-smooth for a fast grid to exist. That should hold for all cell shapes and there is nothing particular about orthogonal cells, which should cause headaches here.

Best,
Paul

pzeiger:
Hi @TomaSusi,

I have rebased this on latest dev and updated the PR description and added a few commits after running a proper benchmark campaign on Perlmutter:

- The description now leads with the *default* behaviour (`sampling='auto'` producing commensurate + FFT-fast grids) rather than with my original memory and compute savings story.
- The benchmark section gained production-scale float64 numbers: 3–6× on CPU, and 3.0× end-to-end on an A100 (3095 s → 1027 s for a 101k-position HAADF scan). The earlier small benchmarks suggested only 1.2–1.8× on GPU — the Bluestein planning cost compounds at scale.
- One correction to my original post: I quoted "10–11 GB" for the FFT-friendly grid on the workload that OOM'ed; the properly measured number is 21.4 GB peak. The controlled comparison (same hardware, same scan, only the grid changed: OOM at 41 GB vs completion at 21.4 GB) is now in the description. The mechanism is cuFFT plan workspace — 11.0 GB for the Bluestein grid vs 1.55 GB for the fast grid at complex128 — while at float32 the memory effect is modest, so memory is now presented as a documented float64-specific effect rather than a general claim.
- The benchmark scripts behind these numbers run against this branch and against #346 (which shares the fft-helpers commit; whichever merges first, the other rebases clean).

With this I think I am happy and you could merge it any time convenient for you. Please let me know if I can do anything else.

Best,
Paul

TomaSusi:
Reviewed the PR-specific changes (separating out what's inherited from #274, which is unaffected here). A few things worth a look, all in the new fast-FFT code specifically:

**1. `is_fast_fft_size`/`next_fast_fft_size` reimplement `scipy.fft.next_fast_len`** (`abtem/core/fft.py`) — scipy is already a hard dependency (used in `abtem/core/backend.py`), and `scipy.fft.next_fast_len(n, real=False)` already finds the next FFT-friendly length. Checked: `next_fast_len(2623) == 2625`, matching this PR's own test case exactly. Might be worth using it directly rather than hand-rolling the prime-factorization loop, unless there's a reason the 2/3/5/7-only restriction needs to diverge from scipy's set.

**2. `_warned_slow_fft_shapes` has a check-then-add race** (`abtem/core/fft.py`, `_warn_slow_fft_size`) — it's a plain module-level `set()`, but `_fft_dispatch` can be called concurrently from multiple dask worker threads via `da.map_blocks`. Two threads hitting a new slow shape at the same time can both pass the `not in` check before either `.add()`s, so the same shape can warn more than once — the "warn once" tests only cover the single-threaded case.

**3. `translational_period_count`'s fallback search is O(n² log n)** (`abtem/slicing.py`, inside `commensurate_gpts`'s `fallback_gpts`) — it searches `m` down from `len(unique_x)` to 2, doing an O(n log n) sort+searchsorted each iteration. For a large supercell of a genuinely incommensurate structure (rutile/brookite-like, repeated many times), this hits close to the worst case and could make `Potential(..., sampling='auto')` construction noticeably slow, uncached, on every call.

TomaSusi:
I'm fixing some issues that came up with my original autogrid implementation in that branch... might it make sense to make this PR target autogrid instead of dev..?

pzeiger:
Hi @TomaSusi,
I think this is a great suggestions. I have discussed a bit with Claude and the significantly expanded tests and aditional bug fixes are well worth it. I will change the target to autogrid and rebase the changes on top of it.
//Paul

TomaSusi:
Thank you! I just pushed my last fix, so I am done for now with autogrid.

pzeiger:
Hi @TomaSusi!

*(This reply was written by Claude — Paul reviewed and posted it.)*

All three review points are addressed, and the PR now targets `autogrid` as you suggested — rebased onto your current tip (including the slice-thickness fixes and the multi-configuration `AtomsEnsemble` refinement you pushed tonight, which briefly conflicted; resolved, and GitHub reports the PR mergeable again). The branch carries only the fast-FFT work and the integration; your call on #274 then brings everything to `dev` together. If you push more to `autogrid`, we'll re-rebase — the overlap is confined to two docstring/comment regions in `iam.py` and resolves mechanically.

On `scipy.fft.next_fast_len`: good catch that they coincide on 2623 → 2625, but the sets genuinely differ — scipy targets pocketfft, whose kernels go up to radix 11, so it returns 11-smooth lengths: `next_fast_len(121) == 121` (= 11²) and `next_fast_len(4619) == 4620` (= 2²·3·5·7·11). cuFFT documents optimized kernels only for sizes of the form 2^a·3^b·5^c·7^d, so a factor-11 grid falls off exactly the GPU fast path that motivated this PR. {2, 3, 5, 7} is the intersection that is fast on every backend *ab*TEM uses, so the stricter set stays, with the divergence documented in a comment and pinned by a test (so a future "simplify to scipy" refactor trips the test rather than silently regressing GPU grids). Nice independent confirmation: CuPy faced exactly this choice and landed on the same side — `cupyx.scipy.fft.next_fast_len` is 7-smooth and agrees with the helper on every value we probed, and its docstring explicitly notes it may differ from scipy "as pocketfft's prime factors are different" from cuFFT's. A second test pins the helper to cupyx's output (skipped when cupy is absent), so drift on either side fails loudly. If your MKL/AMD testing ever shows factor-11 fine on the GPUs too, we're happy to revisit.

One CI note: since `autogrid` forked before your numpy 2.5 fix from #343, this week's numpy 2.5.2 release briefly turned CI red here — so the branch carries your `02140cfb` as a verbatim cherry-pick, and CI is green again. The cherry-pick drops out automatically once `autogrid` catches up with `dev`; whenever you rebase it onto current `dev` (needed before merging #274 anyway), just force-push and we'll rebase this branch on top — the overlap is confined to the two `iam.py` doc regions and resolves mechanically.

The warn-once race is fixed with a lock around the check-then-add (the warning itself is emitted outside the lock), with a barrier-synchronized 8-thread test.

The period search is reworked: a shift that maps the plane set onto itself must map the first plane onto *some* plane, so the only candidates are the differences to the first plane, further filtered to near-integer L/s and tested smallest-first (the first valid shift is the minimal period, i.e. the largest m). Typically O(n log n) now — a 2400-plane incommensurate supercell takes 1.9 ms — with a candidate cap so pathological difference sets cannot go quadratic.

Two things your new test suite caught during the retarget, for the record:

- `test_gpts_is_translation_invariant` failed for rutile/brookite against the original fallback, which scored candidate sizes by distance of the planes to grid points at a fixed origin — not translation-invariant. The fallback now scores *alignability*: the smallest achievable max plane-to-gridpoint distance over all grid-origin phases, computed from the largest circular gap of the plane residues, which depends only on relative positions. Shifted structures now provably get the same grid. Nice test.
- `test_atoms_ensemble_..._skip_commensurate_search` asserted the plain `ceil`-derived gpts; ensemble grids (like all non-commensurate-constrained grids) are now rounded up to fast sizes — rounding is free there since nothing constrains the grid — so the expected values in that test are adapted accordingly (with a comment).

Everything green on the retargeted branch: your `test_commensurate_potential.py` in full (all 19 lattices, hexagonal included), `test_commensurate_gpts.py`, and the grid/fft suites — 151 tests.

Best,
Paul

TomaSusi:
Ran a structured review of your latest commits (fast-fft-grid vs its current base, autogrid) -- 7 independent finder passes plus verification against the checked-out branch, including a direct Python repro for the most serious one. All of the below are in code this PR adds; none trace back to autogrid.

**1. `plane_set_invariant_under` misses the wraparound neighbor (abtem/slicing.py) -- confirmed by repro**

The neighbor search never checks across the 0/L boundary: when a shifted plane lands before `unique_x[0]` or after `unique_x[-1]`, both `np.clip(idx-1, ...)` and `np.clip(idx, ...)` collapse to the same element, so the true nearest plane on the other side of the wrap is never compared. Repro: `unique_x=[1.0, 9.9995]`, `L=10.0`, shifted point `0.0003` -- genuinely `0.0008` from `unique_x[-1]=9.9995` via wraparound (well under `tolerance=1e-3`) -- but the coded logic computes distance `0.9997` (comparing only to `unique_x[0]=1.0`) and wrongly concludes the shift is not invariant. Since a plane sitting near the cell origin/edge is the common case (there's already code snapping `x > L - tolerance` back to `0`), this can under-report the true translational period `m` on ordinary structures, which defeats the point of `fallback_gpts`: keeping `gpts` a multiple of `m` so symmetry-equivalent atoms see identical discretized potentials.

**2. `Potential(sampling='auto')` ignores `config['grid.round-to-fast-fft']`**

`Grid._adjust_gpts` only rounds to a fast FFT size when that config flag is explicitly enabled (default `False`). Neither `commensurate_gpts(...)` call in `Potential.__init__` passes `round_to_fast_fft=False`, so both silently default to `True`, and the non-periodic/multi-config-ensemble branch calls `next_fast_fft_size` directly with no config check at all. A user who sets the flag to `False` expecting no rounding anywhere still gets a rounded-up, finer-than-requested grid from `sampling='auto'`.

**3. Period search gives up silently past 64 candidates (plausible)**

`translational_period_count` returns `m=1` ("no period") once 64 near-integer-ratio candidates fail the invariance check, rather than the best-known valid (if coarser) period. For a plane set with many spurious near-integer ratios ahead of a genuinely valid, coarser period in the candidate order, this silently drops the symmetry guarantee instead of falling back more gracefully.

**4. `Grid.round_to_fast_fft()` duplicates the already-dead `Grid.round_to_power()`**

`round_to_power` already exists for this exact purpose ("Fourier transforms are faster for arrays whose size can be factored into small primes (2, 3, 5 and 7)") and has zero callers anywhere in the repo. Rather than fixing/reusing it -- it only rounds to a single power, so it'd round 2623 to 4096 instead of 2625 -- this PR adds a fully parallel method plus new config/fft.py machinery, leaving the stale dead method in place.

**5. Slow-FFT warning only inspects `shape[-2:]`, missing leading axes in 3D transforms (plausible)**

`fft_interpolate`'s 3D path calls `fftn(array, axes=axes)` with a 3-element `axes` tuple, but `_warn_slow_fft_size(x.shape)` only extracts the trailing two dims -- a non-fast-FFT length on the leading transformed axis never triggers a warning, even though it genuinely forces the Bluestein fallback on GPU for exactly the 3D-potential code path this feature is meant to help with.

**6. Warn-cache locks/grows for every shape, not just slow ones (plausible, minor)**

`_warned_slow_fft_shapes.add(dims)` happens inside the lock before the size/fast-check, so the set accumulates an entry for every distinct GPU FFT shape ever dispatched (not just slow ones) for the life of the process, and every 2D GPU FFT call takes the global lock rather than only new/slow shapes.

🤖 Written by Claude Code on Toma's behalf

pzeiger:
Hi @TomaSusi!

*(This reply was written by Claude — Paul reviewed and posted it.)*

Thanks for the review which found a bunch more issues. All six issues you have found are addressed, and two of them were worse than they looked. Verdicts first, details below: **1 confirmed and reachable on ordinary crystals, 2 confirmed, 3 confirmed, 4 confirmed as a trap rather than a defect, 5 confirmed and worse than stated, 6 confirmed but benign.** Nothing was refuted.

**1 — wraparound: confirmed, and it fires on plain rutile.** Not on the low end though: the caller's `x > L - tolerance -> 0` snap keeps input planes away from L, so the repro as written cannot come out of `commensurate_gpts`. The reachable half is the high end, because the snap protects the *input* planes and not the *shifted* ones — and `orthogonalize_cell` supplies exactly that, returning rutile's L/2 plane one ulp short. Consequences measured through the public API: rutile in `plane='xz'` gave gpts (98, 60) unshifted and (108, 60) shifted by 1e-6 Å, i.e. the documented translation invariance was broken; and calcite 6×5×1 came out with an odd gpts against a true period of 450, spreading symmetry-equivalent atoms over 0.44 px instead of 2e-6 px. Fixed by indexing modulo n. Verified against an independent brute-force periodic oracle over ~1.3 M invariance evaluations: zero disagreements, and zero cases where the new code claims a period the old one correctly rejected.

**3 — the 64-candidate cap: confirmed, and it needs no cap at all.** Reachable on tiled disordered cells — an amorphous support film repeated to cover the field of view is the routine case, and it reproduces on 11 of 12 random seeds. Removing the cap outright is safe: the candidate count is bounded not by the number of planes but by how many distinct `L/m` are more than `tolerance` apart, so the loop is sub-quadratic in atom count at fixed cell size. Worst case constructible inside a plausible cell (≤ 200 Å, ≤ 4000 atoms) is 167 ms, against ~1.2 s just to build the potential array. Two things came out of testing that: each distinct m is now tested once at the **exact** shift `L/m` rather than at the plane difference that suggested it — the difference carries the plane's own float noise, and testing at it let an off-period difference rounding to m suppress the genuine `L/m` shift. Incidentally, findings 1 and 3 compound: on calcite the wrap bug rejected every genuine candidate (450, then 225, then 150…) and the search then burned through the cap and returned "no period" at all.

**2 — config: confirmed.** The two paths genuinely want different defaults, so the option is now tri-state: `'auto'` (new default) rounds only grids abTEM derives itself, `True` additionally rounds gpts derived from a numeric sampling, `False` never rounds. Numeric-sampling behaviour is unchanged in the default mode, and `False` now reproduces the plain commensurate grid **exactly** — verified over a survey of structures, so the two are directly comparable. Since `'auto'` is truthy, every read goes through a mode helper and an unrecognized value raises rather than silently meaning "on".

**5 — 3D transforms: confirmed, and worse than stated.** Both the fastness check *and* the size gate were computed from `shape[-2:]`, so on the structure-factor path (`StructureFactor(..., device='gpu').get_projected_potential` → `slice_potential`) a slow leading axis was entirely invisible — and that grid is essentially never 7-smooth by construction. The transformed lengths are now derived from the function and its `axes`/`axis`/`s` arguments. Two details worth flagging: `s` entries of `None` are tolerated, because this sits on the hot path of every GPU FFT and a raise there breaks the transform rather than merely skipping a diagnostic; and the "set gpts explicitly" remedy is dropped where it does not apply, since the structure-factor grid follows from `g_max` and the cell.

**6 — warn cache: confirmed, benign, fixed anyway.** In a realistic GPU STEM scan the memo held exactly one 130-byte entry, so the growth claim overstates it — the trailing-2D key is essentially constant for a whole run. But reordering costs nothing: the size and fastness checks now run before the lock, so only shapes that actually warn are recorded (200 distinct fast shapes now leave the memo empty) and a run on a fast grid never touches the lock.

**4 — `round_to_power`: confirmed as a trap, not a defect.** Zero callers, and 2623 → 4096 as you say. Since it is dead but public, this PR leaves the behaviour alone and fixes the docstring, which was the actual problem: it now says it produces a *pure* power of one base, cross-references `round_to_fast_fft` as the method that does what the old text promised, and notes the float-`log` artifact that overshoots for 125, 15625 and 16807. Deprecating or redirecting it is your call, not this PR's — happy to do either.

Three things came out of re-reviewing the surrounding code that you did not raise, all now fixed:

- The fallback's documented ±12 % window was not enforced — the candidate that *terminated* the search was still scored, and being the largest it usually won, so rutile took 1.17× the target size (38 % more pixels) for a 2.6 pm alignment gain. The window is now tested before scoring, and candidates are compared by misalignment in Ångström rather than in fractions of a grid spacing, which is not comparable across different grid sizes.
- Grids that were **already** fast were enlarged anyway: Cu fcc 6×6 went from 432 = 2⁴·3³ to 480, +23.5 % pixels, with the equivalent-atom spread at float noise either way. The nearest commensurate grid is now kept when it is already fast, so Si, Cu and MoS₂ keep your branch's own grid unchanged, while Au (164 = 2²·41) and SrTiO₃ (234 = 2·3²·13) still move.
- `grid.round-to-fast-fft: True` rounded **GridScan probe positions**, silently changing the number of probes and the scan step for a grid that is never Fourier transformed — and disagreeing with `LineScan`, which does not use `Grid`. Grids now carry an `fft_grid` flag and `GridScan` opts out.

On testing: the whole suite passes with the option set to `false` and to `true` alike (it previously failed both ways, because several tests read the ambient config instead of pinning it), and every new test was checked to actually fail with the fix it guards reverted — three that could not fail have been rewritten, including the thread-safety test from the last round, which passed against the unlocked code and now observes 8 warnings without the lock and 1 with it.

Best,
Paul

TomaSusi:
I reviewed the round-3 fix commits (with Sonnet 5), then had a second model (Opus 5) independently try to refute each candidate finding before reporting back (running real repros against the checkout rather than reasoning from the diff alone). Two of the original five findings didn't survive that pass -- noting them below too, since being wrong about those is useful signal.

**1. The documented `+-12%` overshoot window in `fallback_gpts` isn't actually a bound -- it's unenforceable in the cases that need it most.**

Reproduced through the public API: 56 planes (u=0.306, L=29.75, target 0.05) gives `gpts=(672,672)` against `n_target=595`, a 12.94% overshoot; with `m=64, n_target=72` it reaches 77.8%. My first pass blamed the `if best is not None` guard for letting the first candidate through unchecked -- that's true, but it's not fixable by moving the guard earlier, since the first candidate is already the *smallest* fast multiple of `m`. There's no better in-window candidate to find. The actual gap is that unlike the "`m` itself isn't a fast size" branch (which falls back to `round(n_target/m)*m`), there's no equivalent escape here when the only fast candidate overshoots badly -- so the comment/docstring's stated bound isn't honored in that case. Might be worth either documenting that the window is best-effort, or adding the same fallback the other branch already has.

**2. `structure_factor_to_potential` (abtem/bloch/dynamical.py:285) calls raw `xp.fft.ifftn` instead of the wrapped version, so it doesn't reach the slow-FFT warning.**

Low severity -- missing diagnostic, not wrong results, and GPU-only. `slice_potential` does call the wrapped `fft_interpolate` for a resize step, but only when `gpts is not None and gpts != potential_3d.shape[:2]`, so coverage is partial: the case where `gpts` is omitted or already matches (i.e. the primary 3D structure-factor transform itself) still goes uncovered.

**3. `Grid.round_to_fast_fft()` (the public method, not the automatic path) ignores `self._fft_grid` and always rounds.**

Confirmed as fact -- `Grid(..., gpts=(2623,2623), fft_grid=False).round_to_fast_fft()` still returns `(2625,2625)`. But it's a nit: only one caller in the repo (a test), and `GridScan` (the only `fft_grid=False` site) never calls it, so it's latent rather than live. Given the `fft_grid` design is otherwise intentional as-is, this probably just wants the method to respect the flag too, or its docstring to say plainly that it always rounds regardless.

**Retracted from my first pass, for the record:**
- *"round() instead of ceil() lets `sampling='auto'` return a coarser-than-target grid"* -- checked against `origin/autogrid` directly: identical code, identical output (`gpts=(400,400)`, sampling 0.050125) on both branches. Pre-existing, not a #347 regression, and the docstring's actual promise ("never coarser than commensurability alone") isn't violated either.
- *"The warning remedy suggests a config option that can't help users with explicit gpts"* -- the remedy text already hedges this exactly ("...or -- if this grid was derived from a numeric sampling -- with abtem.config.set(...)"), and the other half of the sentence (set gpts to the size printed in the message) is the advice that does apply. Not a defect.

🤖 Written by Claude Code on Toma's behalf

pzeiger:
Hi @TomaSusi!

*(This reply was written by Claude — Paul reviewed and posted it.)*

All three addressed. The retractions are appreciated too — and finding 2 turned out to have a sting in the tail.

**1 — the overshoot window is now a real bound.** You were right that moving the guard cannot fix it: candidates ascend, so if the first one overshoots then every fast multiple of `m` does, and there is nothing better to find. Rather than document the window as best-effort I gave this branch the same escape the other one already has: when no fast multiple fits the window, keep the period constraint alone and drop fastness (`round(n_target/m)*m`). That is the trade the `m` is-not-fast branch has always made, so the two branches now behave alike, and your `m=64, n_target=72` case goes to 64 instead of 128. Worth noting explicitly, since it is the same class of thing your retracted finding was about: the escape can land marginally *coarser* than the target, exactly as that pre-existing branch can — the alternative here is +78 % pixels.

**2 — confirmed, but the suggested fix would have silently changed CPU results.** Routing `structure_factor_to_potential` through `abtem.core.fft.ifftn` breaks `test_potential_from_structure_factor` by ~84 units, because the FFTW backend behind that wrapper hardcodes `axes=(-2, -1)` in `_new_fftw_object` — so wrapped `fftn`/`ifftn` silently perform a **2D** transform on a 3D array whenever `fft: fftw` is configured, which is the default. Verified directly against `origin/autogrid`, so this is pre-existing and not something #347 introduces:

```
wrapped ifftn vs np.fft.ifftn(x)                 -> max diff 0.585
wrapped ifftn vs np.fft.ifftn(x, axes=(-2, -1))  -> max diff 0.0
```

So the transform stays on `xp.fft` and asks for the diagnostic explicitly instead, via a helper that only considers GPU arrays (the Bluestein workspace is a cuFFT concern). That gets the coverage you wanted without touching numerics. **The wrapper bug itself I have deliberately left alone** — it is outside this PR, and any fix changes results for whoever currently calls wrapped `fftn`/`ifftn` on 3D data. Happy to open a separate issue or PR for it if you want; it looked worth flagging either way.

**3 — fixed.** `Grid.round_to_fast_fft()` now returns non-FFT grids unchanged, so `fft_grid` means the same thing in the explicit method as in the automatic path, with the docstring saying so. Agreed it was latent — `GridScan` is the only `fft_grid=False` site and never calls it.

Each fix has a test that fails without it, and the full suite's failure set is unchanged from `origin/autogrid`.

Best,
Paul

pzeiger:
Hi @TomaSusi!

Do you have already documentation updates scheduled for this branch? Should I contribute something similar to what Claude and I did in connection with PR #346?

Best,
Paul

TomaSusi:
Yes please. I didn’t yet do any documentation for autogrid, but that will touch a lot of the code – I am planning to make ‘auto’ the default with reasonable target values (say 0.05 Å for lateral sampling and 1 Å for the slice thickness). 

But we should definitely do a separate appendix on the details of all this and the GPU implications, so if you’d like to contribute that, this would be much appreciated :)

TomaSusi:
> Each fix has a test that fails without it, and the full suite's failure set is unchanged from `origin/autogrid`.

Did you perhaps forget to push?-)

TomaSusi:
Verified all three against `bf97cfac` -- read the actual diff and ran the tests against the checked-out branch (not just confirming they exist).

**1. Overshoot window -- fixed.** The fallback to `round(n_target/m)*m` when no in-window fast candidate exists mirrors the existing not-fast-`m` branch exactly. Reproduced: the `m=64, n_target=72` case that used to overshoot to 128 (+78%) now returns 64.

**2. Structure-factor warning -- fixed, and glad you caught the FFTW trap.** `warn_if_slow_gpu_fft` requesting the diagnostic explicitly while leaving the transform on raw `xp.fft` is the right call -- confirmed `_transform_lengths` correctly reads an axes-less `ifftn` as transforming all 3 axes, and `test_potential_from_structure_factor` (the numerics-sensitive one) still passes.

**3. `round_to_fast_fft()` -- fixed.** Now respects `fft_grid` consistently with the automatic path.

Ran the new targeted tests plus the full commensurate/grid/fft/bloch/potentials suite: 195 passed, 38 skipped, no regressions. Nothing outstanding from my side.

🤖 Written by Claude Code on Toma's behalf

TomaSusi (approved):

PR #352, opened 2026-08-25
Add interpolate_line/interpolate_line_at_position for potentials (skew-aware)

TomaSusi:
## Summary
- Adds `interpolate_line` / `interpolate_line_at_position` to potentials (`BaseField`/`FieldArray`), primarily for projected (thickness-summed) potentials, with an optional 3D mode that interpolates a (possibly tilted) line through the full built potential volume.
- The projected path reuses `Images.interpolate_line` via `project()`; the 3D path is a new spline-interpolation path through the potential array.
- Fixes a pre-existing bug in `_BaseMeasurement2D.interpolate_line`'s lazy/dask path: the wrong axes were dropped whenever the measurement had ensemble axes (e.g. frozen phonons), silently corrupting results.
- Fixes a pre-existing bug on this branch (`skewpixels`) in the same method's skewed-cell handling: the Cartesian -> fractional-pixel conversion used a transposed inverse-cell matrix, landing on the wrong lattice position for any non-symmetric skewed cell.
- Makes the new 3D interpolation path skew-aware, using the potential's effective in-plane cell (skewed cell if present, else a diagonal cell from `extent`) for both fractional/Cartesian conversions.
- Adds a `logscale` option to `RealSpaceLineProfiles`/`ReciprocalSpaceLineProfiles.show()` for a log-scaled y-axis (matching the `logscale` kwarg already used by the 2D `show()` methods for color-normalization; unambiguous here since a line plot has no color scale). Needed for `explode=True`/`common_scale` grids, where there's no single `Axes` to pre-configure with `set_yscale("log")` before calling `show(ax=...)`. Along the way, fixed a real bug in `LinesArtist.get_logscale`/`set_logscale` (a getter that mutated state instead of reading it, and a setter with no way to switch back to linear -- both previously unreachable, since nothing called them), and made the y-limit auto-margin log-aware so it no longer emits a spurious `UserWarning` about non-positive log-axis limits.

## Test plan
- [x] New unit tests in `test/test_potentials.py`: projected/3D consistency, lazy vs. eager, fractional coordinates, `width`/`margin` rejected for the 3D path, `interpolate_line_at_position`.
- [x] New skew-specific tests on a hexagonal (non-orthogonal) cell verifying atom-location accuracy in both the projected and 3D paths, lazy/eager consistency, and fractional-coordinate consistency.
- [x] New tests in `test/test_measure.py` for `logscale`: plain overlay, and the exploded/`common_scale` grid case with a warnings-as-errors check confirming no spurious `UserWarning` and positive y-limits across all panels.
- [x] Full `test/` suite: 1038 passed, 661 skipped, 7 failed (all pre-existing GPAW-environment failures, unrelated, reproduced identically on `skewpixels` before this branch).

Draft - opening for early visibility/feedback before finalizing.


TomaSusi:
## Showing where a line profile was taken

`interpolate_line`/`interpolate_line_at_position` record `start`/`end` (and `width`,
if given) in the returned `RealSpaceLineProfiles`/`ReciprocalSpaceLineProfiles`
metadata. `add_to_plot` uses that to draw the line's location on a matplotlib plot:

```python
import abtem
from ase.build import bulk

atoms = bulk("Si", cubic=True) * (4, 4, 4)
potential = abtem.Potential(atoms, sampling=0.05, slice_thickness=1.0)

profile = potential.interpolate_line(start=(0, 0), end=(10, 10), gpts=200)

visualization = potential.project().show()
profile.add_to_plot(visualization, color="r")
```

This also works directly on a plain `Axes` instead of a `Visualization`, with
`interpolate_line_at_position`, and with a nonzero `width` (drawn as the averaging
rectangle rather than a line):

```python
fig, ax = plt.subplots()
potential.project().show(ax=ax)

profile = potential.interpolate_line_at_position(
    center=(5, 5), angle=30, extent=4, width=1.0
)
profile.add_to_plot(ax, color="r")
```

A 3D (non-projected) line profile's `start`/`end` carry a depth (`z`) coordinate;
`add_to_plot` draws its lateral `(x, y)` projection, so overlaying it on the
projected potential still works the same way:

```python
profile = potential.interpolate_line(
    start=(0, 0, 0), end=(5, 5, potential.thickness), gpts=200, projected=False
)
visualization = potential.project().show()
profile.add_to_plot(visualization, color="b")
```

Same for `ReciprocalSpaceLineProfiles` over a `DiffractionPatterns.show()` plot.

pzeiger:
Heads-up from #378: it wraps `_polar_detector_bins` in an `lru_cache` keyed on its (coerced, hashable) arguments, via a canonical `_polar_bins_key` helper. This branch adds `cell=`/`wavelength=` parameters to that same function, and an `ase.Cell` is not hashable — so on rebase the new parameters need to go through the key helper in a hashable form, e.g. `tuple(np.asarray(cell).ravel().tolist())` for the cell (and `float(wavelength)`), extending `_polar_bins_key` rather than the call sites, so the host and device caches stay in lockstep. Also note the cached geometry arrays are now shared between callers and marked read-only — any code that wants to modify them should `astype`/copy first (the existing in-tree callers already do). Happy to help with the rebase when this is ready.

Issue #373, opened 2026-09-07
Separate one-phonon and multi-phonon channels in energy-resolved phonon-loss simulations


TomaSusi:
## Summary

Energy-resolved phonon-loss simulations built on energy-binned frozen-phonon snapshots (as in #324) attribute **all** scattering from a snapshot to that snapshot's energy bin. Because scattering amplitude is nonlinear in displacement, each snapshot carries multi-phonon contributions as well as one-phonon ones — and those get filed under the wrong energy. In an S(q, E) map this shows up as the dispersion **duplicated at q where no mode of that energy exists**.

<img width="432" height="288" alt="Image" src="https://github.com/user-attachments/assets/e4706d29-e783-44bf-8bfa-bc274b0f8f9f" />

Multi-phonon scattering is real physics and interesting in its own right, so this isn't an argument for removing it. The proposal is to make the two channels **separable**, keeping the current (all-orders) behaviour as the default and adding an option to isolate one-phonon — or, equally, to isolate multi-phonon for study.

## The physics

For a snapshot with atoms `a = 1, ..., N`, the kinematic (single/first-order Born) scattering amplitude is

```
F(Q) = Σ_a f_a(Q) exp(iQ·R_a) exp(iQ·u_a)
```

where:

- `Q` — the scattering vector (momentum transfer), in Å⁻¹.
- `f_a(Q)` — the atomic scattering factor (form factor) of atom `a`: the Fourier transform of its own potential/electron density, independent of displacement.
- `R_a` — the *equilibrium* (time-averaged, undisplaced) position of atom `a`.
- `u_a` — the *instantaneous displacement* of atom `a` from `R_a` in this particular frozen-phonon configuration (a superposition of whichever phonon modes are populated in it).

The first exponential, `exp(iQ·R_a)`, alone gives the familiar static structure factor — Bragg peaks where `Q` equals a reciprocal lattice vector, zero elsewhere. The second exponential, `exp(iQ·u_a)`, is **not linear in `u`**, and is the source of the problem: expanding it mixes together every order of scattering off the displacement field, all attributed to whatever energy the displacement `u_a` corresponds to.

For a single mode,

```
u_a = A e cos(q·R_a + φ)
```

with:

- `A` — the mode's (real, scalar) amplitude.
- `e` — the mode's (unit) displacement/polarisation vector.
- `q` — the phonon wavevector.
[...]

PR #374, opened 2026-09-08
Separate one-phonon from multi-phonon scattering via parity projection

TomaSusi:
## Summary

Implements the parity-projection option from #373 to separate one-phonon from multi-phonon scattering in energy-resolved phonon-loss simulations, plus two opt-in extensions that came out of review: **rest-field damping** and **classical snapshot statistics**.

**Background.** `EnergyResolvedAtomsEnsemble`-based phonon-loss simulations (#324) attribute all scattering from a frozen-phonon snapshot to that snapshot's energy bin. Scattering amplitude is nonlinear in displacement (`exp(iQ·u)`), so multi-phonon contributions leak into the wrong energy bin — visible in S(q, E) maps as a spurious dispersion duplicated at q where no mode of that energy exists, growing with Q.

**Fix.** For each snapshot, also propagate its displacement-reversed twin (`2·R_eq − R`), giving

```
psi_odd  = (psi_real − psi_twin) / 2     # u, u³, ...  -> one-phonon
psi_even = (psi_real + psi_twin) / 2     # u⁰, u², ... -> elastic + multi-phonon
```

The one-phonon channel is `mean_j |FT psi_odd_j|²` — no coherent subtraction is needed, since the elastic term is even in the displacement and cancels exactly. The multi-phonon channel is the *variance* of the even part across configurations, which removes the Debye–Waller-damped elastic wave. This also sidesteps the catastrophic-cancellation floor of the existing `I_incoherent − I_coherent` path.

image

image

## API

Four ways to run the pipeline, cheapest first.

```python
import abtem
from abtem.detectors import SpectralSlitDetector
from abtem.inelastic.phonons import EnergyResolvedAtomsEnsemble
from abtem.measurements import (
    momentum_resolved_spectrum,
    phonon_loss_diffraction_patterns,
    unfold_loss_gain,
)

potential_kwargs = dict(sampling=0.05, slice_thickness=eq_atoms.cell[2, 2])
pw = abtem.PlaneWave(energy=60e3)

# 1. Parity projection (two runs per snapshot): all three channels.
ensemble = EnergyResolvedAtomsEnsemble(
    snapshots, energies, equilibrium_atoms=eq_atoms, parity_projection=True,
)  # forces ensemble_mean=False; adds a leading PhononParityAxis("real", "twin")
dp = phonon_loss_diffraction_patterns(
    pw.multislice(abtem.Potential(ensemble, **potential_kwargs)),
    max_angle=100, block_direct=True,
)
dp_all, dp_one, dp_multi = dp[0], dp[1], dp[2]      # "Phonon order" axis
dp_one_signed = unfold_loss_gain(dp_one, temperature=300.0)

# 2. Rest fields (four runs per snapshot): the one-phonon channel alone,
#    Debye-Waller damped by all modes outside the bin.
ensemble_rest = EnergyResolvedAtomsEnsemble(
    snapshots, energies, equilibrium_atoms=eq_atoms, parity_projection=True,
    rest_snapshots=rest_fields,
)
dp_one_damped = phonon_loss_diffraction_patterns(
    pw.multislice(abtem.Potential(ensemble_rest, **potential_kwargs)),
    max_angle=100, block_direct=True, temperature=300.0,
)

# 3. Molecular-dynamics snapshots: unfold with classical statistics.
dp_one_md = unfold_loss_gain(dp_one, temperature=300.0, snapshot_statistics="classical")

# 4. Rest fields plus the static reference (six runs per snapshot): all
#    three channels, damped.
ensemble_full = EnergyResolvedAtomsEnsemble(
    snapshots, energies, equilibrium_atoms=eq_atoms, parity_projection=True,
    rest_snapshots=rest_fields, rest_static_reference=True,
)

detector = SpectralSlitDetector(width=4.0, q_min=0.0, q_max=80.0)   # mrad
spectrum = momentum_resolved_spectrum(dp_one_damped, detector).compute()
spectrum.show(logscale=True)
```

| case | runs/snapshot | returns | `temperature=` |
|---|---|---|---|
| 1, parity only | 2 | "Phonon order" axis (all, one, multi) | raises; use `unfold_loss_gain` on the one slot |
| 2, rest fields | 4 | one-phonon channel, plain `DiffractionPatterns` | allowed |
| 4, rest + static reference | 6 | "Phonon order" axis, damped | raises, as in 1 |

`all == one + multi` holds exactly in cases 1 and 4. Non-parity usage (`parity_projection=False`, the default) is byte-for-byte unchanged.

### Rest fields (`rest_snapshots`, `rest_static_reference`)

A bin-restricted snapshot lacks the Debye–Waller damping of its one-phonon amplitude by all the *other* modes. Supplying a displacement field for the rest of the spectrum propagates each snapshot at the four structures `R_eq ± u_bin ± u_rest`; averaging the exit waves over the rest axis keeps the part even in `u_rest` (the damping) and cancels the mis-binned one-bin-plus-one-rest-phonon term, which is odd. `rest_static_reference=True` adds the per-realization reference `R_eq ± u_rest`, without which the multi channel would be dominated by two-rest-phonon fluctuations of order `u_rest⁴`.

Drawing the rest field from the full thermal ensemble (bin modes included) double-counts the bin modes' own damping — a relative error of about `2·M_bin` in the one-phonon intensity, i.e. `2M / n_bins` for bins of comparable weight. Negligible for narrow bins.

### Classical snapshot statistics (`snapshot_statistics`)

The loss/gain unfolding splits each bin's intensity with `(n+1)/(2n+1)` and `n/(2n+1)`. Those weights are right when the amplitudes were drawn with the Bose–Einstein variance `2n+1` (zero-point motion included, as `rattling(quantum=True)` does). Molecular-dynamics snapshots carry the equipartition variance `2k_BT/E` instead, so their bin intensity is the classical total `1/x` with `x = E / (2k_BT)`, and the correct weights are

```
loss = x (n + 1),    gain = x n,    loss + gain = x coth x
```

`x coth x` is the standard quantum correction of a classical spectrum. Both sets obey detailed balance `loss/gain = exp(E/k_BT)` exactly, tend to ½ each at high temperature, and the classical weights restore the zero-point contribution at low temperature. Default `"quantum"` is unchanged.

## Validation

Full pipeline on a real graphene supercell (336 atoms, single-slice, 100 keV), synthetic single-mode displacement field swept over `Q·u = 0.1 → 1.6`:

| Q·u | ghost/one-phonon (existing `I_tds`) | one-phonon-channel suppression of ghost | one-phonon bias |
|---|---|---|---|
| 0.10 | 8.2×10⁻⁴ | 50× | 0.8% |
| 0.41 | 1.3×10⁻² | 15,500× | 0.03% |
| 1.63 | 2.5×10⁻¹ | 2,600,000× | 1.4% |

- Ghost/one-phonon ratio grows as `(Q·u)²`, matching the predicted Bessel-function scaling.
- Suppression grows dramatically with `Q·u` — strongest exactly where the standard method is worst.
- One-phonon intensity preserved to <1.4% in every case.
- `real`/`twin` exit-wave branches cross-checked bit-for-bit against independent direct `multislice` calls on the same displaced structures.
- Rest-field damping checked per-pixel against `exp(−Q²σ²)` at the one-phonon peaks, with the residual confirmed to be the small positive two-rest-phonon term and the total weight confirmed *not* conserved (the lost weight is a loss, not a redistribution).
- Classical/quantum weights verified to satisfy detailed balance, `x coth x`, and both limiting forms to machine precision.

## Also in this PR

**Memory**: the initial version broadcast a shared static exit wave to the full per-configuration shape and carried it as a third parity member, costing ~1/3 extra exit-wave memory once materialized. The multi channel was then reworked as the variance of the even part, which needs no static reference at all — so the static exit wave was **removed entirely** rather than merely un-replicated, and `all = one + multi` became exact. A production calculation reported >10× lower memory use after this change.

**Correctness fixes from review** (@pzeiger, two rounds):
- Minimum-image displacement check wraps along every non-degenerate cell vector rather than only `pbc`-flagged ones — hand-built ASE `Atoms` default to `pbc=False`, so bulk cells wrapped across a boundary were rejected as multi-ångström displacements.
- `EnergyResolvedAtomsEnsemble.__getitem__` rebuilds the energy axis from the sliced energies instead of copying the parent's, which made the next `Potential` raise an ordinal-axis size mismatch; `ensemble[i, j]` now works.
- `unfold_loss_gain` accepts a `MomentumResolvedSpectrum`, as its docstring already promised.
- `rest_snapshots` refuses pre-built snapshot arrays and cells that differ from `equilibrium_atoms`.
- The `N ≥ 2` guard is scoped to the channels that actually form a variance, so the four-run one-phonon-only mode accepts a single configuration; `snapshot_statistics` is validated at the public entry points so a typo can't be silently ignored when `temperature=None`.

**Visualization fixes**, found while looking at the new "Phonon order" ensemble with `show(explode=True)`:
- `AxesGrid._set_caxes_locators`'s per-panel colorbar loop iterated `(ny outer, nx inner)`, but `self._caxes` is shaped `(ncols, nrows, ncbars)` — placing every colorbar in the right row but the wrong column, invisible on a square grid.
- `MomentumResolvedSpectrum.show(explode=True)` repeated `"Label = value"` on every panel instead of only the first along each exploded axis.
- `show(explode=True)` gained a `suptitle=` parameter for a whole-figure title distinct from the per-panel `title=`.
- `MomentumResolvedSpectrum`'s `pcolormesh` is rasterized by default, so saved PDFs no longer show a fine grid overlaid on the data.

## Test plan
- [x] `test/test_energy_resolved_parity.py` (29 tests): ensemble shape/axes, twin construction, `ensemble_mean` auto-forcing, `equilibrium_atoms` and minimum-image validation, `rest_snapshots` validation, indexing, dask blockwise machinery on the 3D/4D layouts.
- [x] `test/test_parity_projection_multislice.py` (9 tests): real (small) graphene multislice, exit-wave branches matching independent direct `multislice` calls, rest-field Debye–Waller damping against `exp(−Q²σ²)`, multi-channel damping ratio, four-run vs six-run one-phonon channel bit-for-bit, single-configuration one-phonon-only mode.
- [x] `test/test_phonon_loss.py` (30 tests): "Phonon order" ensemble shape/axes, `"all"` matches the direct non-parity call, `"one"`/`"multi"` isolate independently-constructed known signals, validation errors, temperature unfolding, classical/quantum statistics through both entry points, lazy/eager consistency.
- [x] `test/test_axes_grid.py`, `test/test_spectral_detectors.py`: colorbar column placement (fails on old code), non-square grids, `suptitle`, exploded-panel title convention, rasterized `pcolormesh`.
- [x] Full suite after merging current `dev`: see CI.

🤖 Written by [Claude Code](https://claude.com/claude-code) on behalf of [Toma](https://github.com/tomasusi)


TomaSusi:
## Slice-thickness sensitivity for materials with real z-extent

Production testing turned up a case worth documenting: parity projection separates one-phonon from multi-phonon cleanly for monolayer graphene or hBN, but not as cleanly for monolayer MoS2 — until slice thickness is reduced.

**The odd/even decomposition itself is dimension-agnostic** — nothing in `psi_odd = [psi(+u) - psi(-u)] / 2` assumes planarity, and it's exact for whatever function the multislice code computes, regardless of z-extent.

**More likely explanation**: ordinary multislice slice-thickness truncation error — a separate, well-known approximation (phase-grating + Fresnel propagation is only accurate to leading order in `slice_thickness`) independent of parity projection or projection mode. It's negligible for hBN, which sits in essentially one z-plane, but MoS2's S–Mo–S sandwich spans ~3.1 Å, so a coarse slice thickness genuinely under-resolves the static structure before phonon displacement is even added.

This matters more for parity projection specifically than for an ordinary summed-intensity calculation: the method works by differencing two large, nearly-identical complex exit waves (`psi(+u)`, `psi(-u)`). Any inconsistency in how the two displacement signs are discretized near a slice boundary doesn't cancel in that difference — it shows up as leakage between the odd and even channels. A total-intensity slice-thickness convergence check (the usual way to pick this parameter) doesn't probe this at all; it can look converged while the much smaller odd/even difference is still dominated by discretization noise.

**Practical takeaway**: for materials with genuine out-of-plane structure, converge `slice_thickness` against the parity-separated ratio itself (e.g. spurious/duplicate-branch intensity vs. one-phonon-peak intensity), not against total S(q, E) — it likely needs to be finer than total-intensity convergence alone would suggest. Confirmed empirically: smaller slice thickness improves the MoS2 separation.

pzeiger:
Hi Toma,

Claude and I have reviewed this PR in the last days. The parity trick is a really nice idea and the one-phonon (odd) channel is exactly right to leading order: I checked it bit-for-bit against a manual `mean|(psi_real − psi_twin)/2|²` on a real graphene multislice. Two things are wrong, though, one in the physics of the multi-phonon channel and one regression in shared plotting code, plus a few smaller points. Every finding below was reproduced numerically on a live worktree, not read off the diff.

I have put fixes with tests on the branch `parity-projection-fixes` (four commits on top of your head, one per point, [compare view](https://github.com/abTEM/abTEM/compare/claude/parity_projection_phonons...parity-projection-fixes)), so you can cherry-pick, merge, or just take them as a reference — your call. Full CPU test suite passes on that branch (1403 passed).

1. **Multi-phonon channel contains the elastic Debye–Waller term** (inline comment on `measurements.py`). `mean_j |psi_diff_j|²` needs `− |mean_j psi_diff_j|²`; without it the channel carries an O(u⁴) elastic term on the reciprocal-lattice points, 40× the true multi-phonon intensity at higher-order spots on a graphene test. This could be the reason for the streaks at Gamma-points in the multi phonon channel of the figure you have shared with me. Fixed as the variance of the even part in [e5a94a0a](https://github.com/abTEM/abTEM/commit/e5a94a0a8d484bf346b61b84c1000cb53b895e1c). With that form the static exit wave drops out algebraically; a complex128 upcast at the diffraction-pattern step gives identical numerics, so that commit also removes the static-wave machinery (second multislice run, `Waves.static_exit_wave`, zarr companion arrays) and makes `"all"` = `one + multi`, which is exactly the ordinary TDS estimator over the full ±u set.
2. **`suptitle` breaks the exploded-plot title chain in `Visualization`** (inline comment on `visualizations.py`). `show(explode=True, title="foo")` now ignores the custom title, and with a suptitle the auto column titles vanish, for every measurement type. Fixed in [65765137](https://github.com/abTEM/abTEM/commit/657651374b7bbf19c2624929a70fc84e0a029701).
3. **`temperature` unfolding applied to the multi channel with one-phonon Bose weights** (inline comment on `measurements.py`). Now raises with an explanation, and a public `abtem.measurements.unfold_loss_gain(measurement, temperature)` lets one unfold the `"one"` slot deliberately: [a4d419b1](https://github.com/abTEM/abTEM/commit/a4d419b1fe845c2b566d345ffa3d469a0ddebc96).
4. **No validation of `equilibrium_atoms` against the snapshots** (inline comment on `phonons.py`). Count, species and a minimum-image displacement bound: [ce71d615](https://github.com/abTEM/abTEM/commit/ce71d615ffa3463c7ec4bc2cd8e70f398a18b52f).
5. **`static_exit_wave` lifecycle** (inline comment on `waves.py`) — moot after (1), noted for the record.

Usage after the fixes, including detailed balance applied afterwards to the one-phonon channel only (run end-to-end on the fix branch):

```python
import abtem
from abtem.detectors import SpectralSlitDetector
from abtem.inelastic.phonons import EnergyResolvedAtomsEnsemble
from abtem.measurements import (
    momentum_resolved_spectrum,
    phonon_loss_diffraction_patterns,
    unfold_loss_gain,
)

# snapshots: list (one per energy bin) of lists of displaced ASE Atoms,
# all displacements of eq_atoms; energies start at 0 and increase [eV]
ensemble = EnergyResolvedAtomsEnsemble(
    snapshots, energies, equilibrium_atoms=eq_atoms, parity_projection=True
)
potential = abtem.Potential(ensemble, sampling=0.05, slice_thickness=eq_atoms.cell[2, 2])
exit_waves = abtem.PlaneWave(energy=60e3).multislice(potential, lazy=True)

dp = phonon_loss_diffraction_patterns(exit_waves, max_angle=100, block_direct=True)
# leading "Phonon order" axis: ("all", "one", "multi"); all == one + multi exactly
dp_all, dp_one, dp_multi = dp[0], dp[1], dp[2]

# Detailed balance on the one-phonon channel only. The multi channel's
# energy axis is the bin's mode energy, not the energy transfer (its
# two-phonon processes sit at +2E, 0, -2E), so it must not be unfolded
# with one-phonon weights -- passing temperature= to the call above raises.
dp_one_signed = unfold_loss_gain(dp_one, temperature=300.0)  # E axis -> (-E_max ... 0 ... +E_max)

detector = SpectralSlitDetector(width=4.0, q_min=0.0, q_max=80.0)  # mrad
spectrum = momentum_resolved_spectrum(dp_one_signed, detector).compute()
spectrum.show(logscale=True)
```

One physics remark beyond the code, since the PR text touches on it: the odd channel removes the coherent-subtraction artefact but not the under-damping of bin-restricted snapshots. Its one-phonon amplitude still lacks the Debye–Waller factor of the modes outside the bin, `e^{-M_other(Q)}`, which for the graphene numbers in #373 (Q·u_rms ≈ 0.35 at the outer spots) is a ~12 % intensity effect, larger than any of the residual higher-order terms. So it complements the Zeiger et al. smearing correction rather than replacing it. Happy to discuss separately — there is a cheap way to recover that factor with a second sign average over the rest modes, which I can write up if there is interest.

abtem/measurements.py
```diff
@@ -6280,6 +6314,172 @@ def _broadcast(weight):
...
+        psi_odd, **waves_real._copy_kwargs(exclude=("array",))
+    )
+    I_one_phonon = waves_odd.diffraction_patterns(**dp_kwargs).array.mean(
+        axis=fp_axis_idx
+    )
+
+    # --- "multi_phonon": (real + twin)/2 - static, isolates u^2 and higher ---
+    # static_array is (nx, ny) -- broadcasts against (..., nx, ny) for free,
+    # never physically replicated across the energy/config axes.
+    psi_diff = (waves_real.array + waves_twin.array) / 2 - static_array
```
pzeiger: This channel is `mean_j |psi_diff_j|²`, but `mean_j psi_diff_j` is not zero. `psi_even = (psi(+u) + psi(−u))/2` averages to the bin's coherent, Debye–Waller-damped elastic wave, while `psi_static = psi(0)` is the undamped one; kinematically `<psi_diff>(Q) = Σ_a f_a e^{iQ·R_a} (e^{−M_a(Q)} − 1)`, non-zero exactly on the reciprocal-lattice points, amplitude O(u²), intensity O(u⁴) — the same order as the two-phonon signal, but elastic (zero energy loss). So `|<psi_diff>|²` sits inside `"multi"` as a bright spot on every (especially higher-order) G.

There is an exact identity that exposes it. For the symmetric 2N set `{+u_j, −u_j}` the ordinary estimator decomposes as `TDS_all(±u) = mean_j|psi_odd_j|² + [ mean_j|psi_even_j|² − |mean_j psi_even_j|² ]`, i.e. `one + multi_correct`, because the odd part of the mean and the odd–even cross term cancel pairwise. Measured on a graphene 6×3 supercell, 100 keV, 8 configs, σ = 0.05 Å, float64, relative to the one-phonon maximum: `all(±u) − (one + multi_PR)` = 0.41 max; `all(±u) − (one + multi_corrected)` = 1.2e−9. At a reciprocal-lattice pixel with static I/I₀ = 1.4e−5 the PR value is 0.448 vs 0.0116 corrected (≈ 40×), and the integrated excess is 1.09× the genuine multi-phonon weight. The synthetic test `test_one_phonon_and_multi_phonon_isolate_known_signals` mirrors the implementation's formula rather than testing physics, and the toy in #373 only measured this channel at q₀ and 2q₀, never at a G, which is why it went unnoticed.

Fix: `multi = mean_j |FT psi_even_j|² − |FT mean_j psi_even_j|²`. The static wave then cancels algebraically; it was only a precision aid. Evaluating this channel with the exit waves upcast to complex128 for the diffraction-pattern step is numerically identical to the static-wave design (2e−6 vs a float64 reference, in a float32 session; the naive float32 variance is off by 0.17), so [e5a94a0a](https://github.com/abTEM/abTEM/commit/e5a94a0a8d484bf346b61b84c1000cb53b895e1c) does that and removes the static-wave plumbing. Two exact tests replace the mirroring one: `all == one + multi` and `all ==` the plain estimator on the real+twin snapshots, in float64 to round-off, plus a float32-vs-float64 guard. One dask gotcha found on the way: the lazy diffraction-pattern path stamps the *configured* dtype as meta, and dask reductions accumulate in the meta dtype, so an explicit `astype(float64)` is needed or the lazy path silently downcasts before the subtraction (the lazy-vs-eager test catches it).

abtem/measurements.py
```diff
@@ -6280,6 +6314,172 @@ def _broadcast(weight):
...
+    if fp_axis_idx is None or energy_axis_idx is None:
+        raise ValueError(
+            "exit_waves must have both a FrozenPhononsAxis and an "
+            "EnergyLossAxis in addition to the PhononParityAxis."
+        )
+
+    dp_kwargs = dict(max_angle=max_angle, parity=parity, fftshift=True)
+
+    # --- "all": identical to the non-parity path, forced to component="tds" ---
+    dp_all = phonon_loss_diffraction_patterns(
```
pzeiger: `"all"` is computed from the N real configurations only, discarding the N twins, which are equally valid draws (φ → φ + π), and it costs two extra diffraction-pattern passes. With the multi fix above, `one + multi` over the ±u pairs *is* the ordinary estimator over all 2N members, exactly (see the identity in the previous comment), with twice the statistics and antithetic-sampling variance reduction of the coherent part for free. With 8 configs the current `"all"` differs from `one + multi` by 0.13–0.24 of the one-phonon maximum purely from N-vs-2N sampling noise, which a user will read as a bug. [e5a94a0a](https://github.com/abTEM/abTEM/commit/e5a94a0a8d484bf346b61b84c1000cb53b895e1c) sets `all = one + multi`; the bit-equality test against a real-only call becomes the exact-identity test.

abtem/waves.py
```diff
@@ -389,6 +389,104 @@ def reduce_ensemble(
     return reduced_output
 
 
+def _add_parity_projection_static_branch(
```
pzeiger: Moot if the multi-channel change is taken (the static wave is not needed at all), but for the record, the attribute lifecycle as designed is fragile: `compute()` keeps `_static_exit_wave` but as an un-computed dask array on an eager object (`_compute` swaps `_array` in place), so `phonon_loss_diffraction_patterns(ew.compute())` returns a lazy result from eager input; `to_cpu()`/`copy_to_device()`, `rechunk`, `ensure_lazy` and slicing drop it and the downstream call raises — GPU users routinely `to_cpu()` before analysis. If it were kept, promoting it to a `Waves` constructor keyword so `_copy_kwargs` carries it would be the robust variant. [e5a94a0a](https://github.com/abTEM/abTEM/commit/e5a94a0a8d484bf346b61b84c1000cb53b895e1c) removes this function, the three multislice hooks, `_build_equilibrium_potential`, the property, and the zarr companion arrays in `array.py`.

abtem/measurements.py
```diff
@@ -6280,6 +6314,172 @@ def _broadcast(weight):
...
+        remaining_energy_axis_idx = next(
+            i for i, ax in enumerate(remaining_axes) if isinstance(ax, EnergyLossAxis)
+        )
+        e_values = np.asarray(
+            waves_real.ensemble_axes_metadata[energy_axis_idx].values, dtype=float
+        )
+        I_one_phonon, _ = _thermal_weight_tds(
+            I_one_phonon, e_values, remaining_energy_axis_idx, temperature
+        )
+        I_multi_phonon, _ = _thermal_weight_tds(
```
pzeiger: `_thermal_weight_tds` moves bin E to ±E with the one-phonon weights `(n+1)/(2n+1)` and `n/(2n+1)`. For the multi channel that is wrong twice: its same-mode two-phonon processes transfer +2E, 0 and −2E with weights `(n+1)²`, `2n(n+1)`, `n²` over `(2n+1)²` — nothing at ±E — and two-mode pairs in the bin land at ≈ E_i + E_j ≈ 2E and ≈ 0 likewise. The `"multi"` slot of a temperature-unfolded result is therefore placed at the wrong energy with the wrong weights; it is exactly the mis-attribution #373 set out to fix, now isolated but still labelled E. Correct two-phonon weights need an extended (±2E_max) energy axis, which I would leave out of this PR. [a4d419b1](https://github.com/abTEM/abTEM/commit/a4d419b1fe845c2b566d345ffa3d469a0ddebc96) raises for `temperature` with a parity axis, with the explanation in the message, and exposes the existing weighting as a public `unfold_loss_gain(measurement, temperature)` so the `"one"` slot can be unfolded on purpose (see the snippet in the summary).

abtem/inelastic/phonons.py
```diff
@@ -720,13 +754,31 @@ def __init__(
...
+            # stack it as a new leading axis. A `snapshots.ndim == 3` input
+            # here means this is instead a dask/chunk reconstruction of an
+            # ensemble that already has the parity axis baked in (see
+            # `_from_partition_args_func` below) -- must not be re-twinned.
+            eq_positions = equilibrium_atoms.positions
+            twin = np.empty_like(snapshots)
+            for index in np.ndindex(snapshots.shape):
+                atoms = snapshots[index]
+                twin_atoms = atoms.copy()
+                twin_atoms.positions = 2 * eq_positions - atoms.positions
```
pzeiger, line 831: Nothing checks that `equilibrium_atoms` matches the snapshots. A wrong atom count surfaces as a raw numpy broadcasting error; a different ordering or species sequence produces meaningless twins silently. [ce71d615](https://github.com/abTEM/abTEM/commit/ce71d615ffa3463c7ec4bc2cd8e70f398a18b52f) checks count and species and bounds the minimum-image displacement of every atom (`max_displacement=1.0` Å by default, `None` disables), with a message that points at misordering as the usual cause. Minimum image, so snapshots wrapped back into the cell pass; the twin `2R_eq − R` is correct under periodic wrapping anyway, since `R_eq − u − L` wraps to `R_eq − u`.

abtem/visualize/visualizations.py
```diff
@@ -242,6 +243,9 @@ def __init__(
         if isinstance(title, str):
             self.set_column_titles(title)
 
+        if suptitle:
```
pzeiger: This `if` was inserted between `if isinstance(title, str):` and its `elif`, so the `elif title and len(explode) > 0` now hangs off `if suptitle`. Measured on an `Images` ensemble with `show(explode=True, ...)`: on `dev`, `title="foo"` gives column titles `['foo','foo','foo']`; on this branch it gives `['Phonon order = all','one','multi']` (custom title overwritten); `title=True, suptitle="sup"` gives `[]` (auto titles lost). This affects every exploded `show()` through `Visualization`, not just the new axis, and the added suptitle test does not assert on column titles, which is why it passed. [65765137](https://github.com/abTEM/abTEM/commit/657651374b7bbf19c2624929a70fc84e0a029701) moves the `suptitle` call after the chain and adds a regression test that fails on the old code.

abtem/measurements.py
pzeiger: This channel is `mean_j |psi_diff_j|²`, but `mean_j psi_diff_j` is not zero. `psi_even = (psi(+u) + psi(−u))/2` averages to the bin's coherent, Debye–Waller-damped elastic wave, while `psi_static = psi(0)` is the undamped one; kinematically `<psi_diff>(Q) = Σ_a f_a e^{iQ·R_a} (e^{−M_a(Q)} − 1)`, non-zero exactly on the reciprocal-lattice points, amplitude O(u²), intensity O(u⁴) — the same order as the two-phonon signal, but elastic (zero energy loss). So `|<psi_diff>|²` sits inside `"multi"` as a bright spot on every (especially higher-order) G.

There is an exact identity that exposes it. For the symmetric 2N set `{+u_j, −u_j}` the ordinary estimator decomposes as `TDS_all(±u) = mean_j|psi_odd_j|² + [ mean_j|psi_even_j|² − |mean_j psi_even_j|² ]`, i.e. `one + multi_correct`, because the odd part of the mean and the odd–even cross term cancel pairwise. Measured on a graphene 6×3 supercell, 100 keV, 8 configs, σ = 0.05 Å, float64, relative to the one-phonon maximum: `all(±u) − (one + multi_PR)` = 0.41 max; `all(±u) − (one + multi_corrected)` = 1.2e−9. At a reciprocal-lattice pixel with static I/I₀ = 1.4e−5 the PR value is 0.448 vs 0.0116 corrected (≈ 40×), and the integrated excess is 1.09× the genuine multi-phonon weight. The synthetic test `test_one_phonon_and_multi_phonon_isolate_known_signals` mirrors the implementation's formula rather than testing physics, and the toy in #373 only measured this channel at q₀ and 2q₀, never at a G, which is why it went unnoticed.

Fix: `multi = mean_j |FT psi_even_j|² − |FT mean_j psi_even_j|²`. The static wave then cancels algebraically; it was only a precision aid. Evaluating this channel with the exit waves upcast to complex128 for the diffraction-pattern step is numerically identical to the static-wave design (2e−6 vs a float64 reference, in a float32 session; the naive float32 variance is off by 0.17), so [e5a94a0a](https://github.com/abTEM/abTEM/commit/e5a94a0a8d484bf346b61b84c1000cb53b895e1c) does that and removes the static-wave plumbing. Two exact tests replace the mirroring one: `all == one + multi` and `all ==` the plain estimator on the real+twin snapshots, in float64 to round-off, plus a float32-vs-float64 guard. One dask gotcha found on the way: the lazy diffraction-pattern path stamps the *configured* dtype as meta, and dask reductions accumulate in the meta dtype, so an explicit `astype(float64)` is needed or the lazy path silently downcasts before the subtraction (the lazy-vs-eager test catches it).

pzeiger (commented):
Hi Toma,

Claude and I have reviewed this PR in the last days. The parity trick is a really nice idea and the one-phonon (odd) channel is exactly right to leading order: I checked it bit-for-bit against a manual `mean|(psi_real − psi_twin)/2|²` on a real graphene multislice. Two things are wrong, though, one in the physics of the multi-phonon channel and one regression in shared plotting code, plus a few smaller points. Every finding below was reproduced numerically on a live worktree, not read off the diff.

I have put fixes with tests on the branch `parity-projection-fixes` (four commits on top of your head, one per point, [compare view](https://github.com/abTEM/abTEM/compare/claude/parity_projection_phonons...parity-projection-fixes)), so you can cherry-pick, merge, or just take them as a reference — your call. Full CPU test suite passes on that branch (1403 passed).

1. **Multi-phonon channel contains the elastic Debye–Waller term** (inline comment on `measurements.py`). `mean_j |psi_diff_j|²` needs `− |mean_j psi_diff_j|²`; without it the channel carries an O(u⁴) elastic term on the reciprocal-lattice points, 40× the true multi-phonon intensity at higher-order spots on a graphene test. This could be the reason for the streaks at Gamma-points in the multi phonon channel of the figure you have shared with me. Fixed as the variance of the even part in [e5a94a0a](https://github.com/abTEM/abTEM/commit/e5a94a0a8d484bf346b61b84c1000cb53b895e1c). With that form the static exit wave drops out algebraically; a complex128 upcast at the diffraction-pattern step gives identical numerics, so that commit also removes the static-wave machinery (second multislice run, `Waves.static_exit_wave`, zarr companion arrays) and makes `"all"` = `one + multi`, which is exactly the ordinary TDS estimator over the full ±u set.
2. **`suptitle` breaks the exploded-plot title chain in `Visualization`** (inline comment on `visualizations.py`). `show(explode=True, title="foo")` now ignores the custom title, and with a suptitle the auto column titles vanish, for every measurement type. Fixed in [65765137](https://github.com/abTEM/abTEM/commit/657651374b7bbf19c2624929a70fc84e0a029701).
3. **`temperature` unfolding applied to the multi channel with one-phonon Bose weights** (inline comment on `measurements.py`). Now raises with an explanation, and a public `abtem.measurements.unfold_loss_gain(measurement, temperature)` lets one unfold the `"one"` slot deliberately: [a4d419b1](https://github.com/abTEM/abTEM/commit/a4d419b1fe845c2b566d345ffa3d469a0ddebc96).
4. **No validation of `equilibrium_atoms` against the snapshots** (inline comment on `phonons.py`). Count, species and a minimum-image displacement bound: [ce71d615](https://github.com/abTEM/abTEM/commit/ce71d615ffa3463c7ec4bc2cd8e70f398a18b52f).
5. **`static_exit_wave` lifecycle** (inline comment on `waves.py`) — moot after (1), noted for the record.

Usage after the fixes, including detailed balance applied afterwards to the one-phonon channel only (run end-to-end on the fix branch):

```python
import abtem
from abtem.detectors import SpectralSlitDetector
from abtem.inelastic.phonons import EnergyResolvedAtomsEnsemble
from abtem.measurements import (
    momentum_resolved_spectrum,
    phonon_loss_diffraction_patterns,
    unfold_loss_gain,
)

# snapshots: list (one per energy bin) of lists of displaced ASE Atoms,
# all displacements of eq_atoms; energies start at 0 and increase [eV]
ensemble = EnergyResolvedAtomsEnsemble(
    snapshots, energies, equilibrium_atoms=eq_atoms, parity_projection=True
)
potential = abtem.Potential(ensemble, sampling=0.05, slice_thickness=eq_atoms.cell[2, 2])
exit_waves = abtem.PlaneWave(energy=60e3).multislice(potential, lazy=True)

dp = phonon_loss_diffraction_patterns(exit_waves, max_angle=100, block_direct=True)
# leading "Phonon order" axis: ("all", "one", "multi"); all == one + multi exactly
dp_all, dp_one, dp_multi = dp[0], dp[1], dp[2]

# Detailed balance on the one-phonon channel only. The multi channel's
# energy axis is the bin's mode energy, not the energy transfer (its
# two-phonon processes sit at +2E, 0, -2E), so it must not be unfolded
# with one-phonon weights -- passing temperature= to the call above raises.
dp_one_signed = unfold_loss_gain(dp_one, temperature=300.0)  # E axis -> (-E_max ... 0 ... +E_max)

detector = SpectralSlitDetector(width=4.0, q_min=0.0, q_max=80.0)  # mrad
spectrum = momentum_resolved_spectrum(dp_one_signed, detector).compute()
spectrum.show(logscale=True)
```

One physics remark beyond the code, since the PR text touches on it: the odd channel removes the coherent-subtraction artefact but not the under-damping of bin-restricted snapshots. Its one-phonon amplitude still lacks the Debye–Waller factor of the modes outside the bin, `e^{-M_other(Q)}`, which for the graphene numbers in #373 (Q·u_rms ≈ 0.35 at the outer spots) is a ~12 % intensity effect, larger than any of the residual higher-order terms. So it complements the Zeiger et al. smearing correction rather than replacing it. Happy to discuss separately — there is a cheap way to recover that factor with a second sign average over the rest modes, which I can write up if there is interest.

TomaSusi:
Thanks a lot @pzeiger, really critical findings and improvements. They should all now be implemented, and I ran a further round of review – could you set your agent on this one more time to confirm?

pzeiger:
Thanks Toma. Re-reviewed the head `fd904e38`: all five points of the first round are in as proposed, and the four new commits were read, tested individually, and then attacked with inputs their tests do not cover. Verdict: ready apart from two regressions, both small and both fixed with tests on `parity-projection-fixes-2` (three commits on top of your head, [compare](https://github.com/abTEM/abTEM/compare/claude/parity_projection_phonons...parity-projection-fixes-2)), plus two pre-existing defects found on the way and fixed there as well.

**The four new commits**

- `758be138`, minimum-image check via `find_mic`: right diagnosis (the rank-3 requirement skipped the wrap for ASE's 2D builders), but the replacement keys the wrap on `atoms.pbc`, and hand-built ASE `Atoms` default to `pbc=False`. A bulk cell with an ordinary displacement wrapped across a boundary is now rejected as a 9.95 Å displacement (reproduced with `Atoms("H2", cell=[10, 10, 10])`, no `pbc`), where the previous version accepted it. abTEM treats the potential as periodic along every cell vector regardless of `pbc`, so the wrap directions should come from the cell: wrap along every non-zero cell vector, skip the zero one. That keeps the rank-2 case and restores the bulk `pbc=False` case; the fix branch does exactly that with a test for each.
- `2fd401c2`, `block_direct=True` resolving to a 1 mrad radius: correct, a real bug, and it was in the code before this PR as well. Both call sites fixed and tested.
- `4ac6f6a5`, eight cleanups: seven are fine. The new `ensemble[:, k]` indexing is broken for its own use case: `__getitem__` copies the parent's `ensemble_axes_metadata` into the slice, so the slice's `EnergyLossAxis` keeps all energies while its snapshots keep one, and building a `Potential` from it raises an ordinal-axis size mismatch. The pre-existing non-parity `ensemble[k]` has the same defect, and `ensemble[i, j]` crashes on a bare `Atoms` object. All three fixed in `72aa303f` (the constructor rebuilds the axis from the sliced energies; one test covers the three forms, each followed by a multislice). Two minor notes besides: (a) in the parity `__getitem__`, `item == slice(None)` on a numpy index array evaluates elementwise and raises a `ValueError` instead of the documented `NotImplementedError`; an `isinstance(item, slice)` guard fixes it (included in `f971e480` with a test). (b) `_ensemble_mean_forced` is a dataclass field, so `axis_to_dict` serializes it into zarr like `_ensemble_mean`; old files load fine (the field defaults), but files written by this version will not load in an older abTEM, whose `axis_from_dict` passes the unknown key to the constructor. Same forward-compatibility trade-off as `_ensemble_mean` already makes, so just noting it.
- `fd904e38`, scoped warning test: fine.

The third commit on that branch, `f0586ddf`, fixes a defect in the earlier fixes rather than in the new commits: `unfold_loss_gain` promised to accept a `MomentumResolvedSpectrum` but raised on one, since the spectrum's energy is a base axis; it now unfolds along that axis, and unfolding before or after `momentum_resolved_spectrum` gives the same result.

**Checks run on `fd904e38`**: the seven touched test files (energy-resolved parity, phonon loss, parity multislice, axes grid, spectral detectors, array, waves) pass, 382 tests, plus the adversarial cases above; also verified that the private `_validated` flag does propagate through dask chunk reconstruction, that a `Probe.scan` with a `CustomScan` over a parity potential flows through `phonon_loss_diffraction_patterns` with its positions axis intact, and that a parity result survives a zarr round trip.

**Follow-up, separate from this PR.** The one-phonon channel of a bin-restricted snapshot lacks the Debye–Waller factor of the modes outside the bin (about 12 % in intensity at the outer spots for graphene), which parity projection cannot supply. A second sign average over a rest displacement field recovers it: `EnergyResolvedAtomsEnsemble(..., rest_snapshots=...)` propagates every snapshot at `R_eq ± u_bin ± u_rest` on a new `PhononRestParityAxis`, and `phonon_loss_diffraction_patterns` averages the exit waves over it before the projection. The part even in `u_rest` carries the damping; the part odd in `u_rest`, one bin phonon plus one rest phonon, is mis-binned and cancels exactly. In this mode only the one-phonon channel is returned (a plain `DiffractionPatterns`, as `component="tds"` gives, and `temperature=` is allowed for it): the multi channel would be swamped by two-rest-phonon fluctuations, order `u_rest⁴`, unless a per-realization reference `R_eq ± u_rest` is subtracted, which `rest_static_reference=True` adds as a third parity member at six runs per snapshot, giving back all three channels. Verified on graphene against `exp(-Q² σ²)` at the one-phonon peaks, and with the reference the multi/one ratio stays that of the bin-only run. It is on `parity-rest-displacements`, merged with your current head ([compare](https://github.com/abTEM/abTEM/compare/claude/parity_projection_phonons...parity-rest-displacements)), and can go in here or as its own PR after this one merges, whichever you prefer; four runs per snapshot instead of two (six with the reference), opt-in, nothing changes without the new argument.

pzeiger:
One more follow-up, separate from the review points: a `snapshot_statistics` argument for the loss/gain unfolding, so that molecular-dynamics snapshots can be unfolded correctly.

The unfolding in this PR splits each bin's intensity into loss and gain with `(n+1)/(2n+1)` and `n/(2n+1)`. Those weights are right when the snapshot amplitudes were drawn with the Bose–Einstein variance `2n+1` per mode, zero-point motion included, as `rattling` does with `quantum=True`: the bin intensity is then the quantum total `(n+1) + n`, and the weights only redistribute it. Snapshots from molecular dynamics carry the equipartition variance `2 k_B T / E` instead, so their bin intensity is the classical total `1/x` with `x = E / (2 k_B T)`, and splitting it with the quantum weights leaves it classical: too weak at low temperature, with the spurious `1/E²` scaling that #373 mentions from Zeiger et al. 2023. For that case the weights have to be

```
loss = x (n + 1),    gain = x n,    loss + gain = x coth x
```

`x coth x` is the standard quantum correction of a classical spectrum; both sets of weights obey detailed balance `loss / gain = exp(E / k_B T)` exactly; the classical weights go to ½ each at high temperature and restore the zero-point contribution at low temperature. The zero-energy bin passes through unweighted in both modes, as before.

API: `phonon_loss_diffraction_patterns(..., temperature=T, snapshot_statistics="classical")` and `unfold_loss_gain(measurement, T, snapshot_statistics="classical")`. The default `"quantum"` leaves every existing call unchanged; an unknown value raises at all entry points. Tests check that classical equals quantum times `x coth x` on every energy and both sides, detailed balance in both modes, the two temperature limits, and that both entry points agree.

One limitation the correction cannot remove, stated in the docstring: it acts on each bin's intensity, not on the snapshots, so a molecular-dynamics displacement field still carries the classical Debye–Waller factor, which is too small at low temperature. That is a property of classical sampling itself.

It is on `phonon-loss-classical-statistics`, one commit on top of the fix branch ([compare](https://github.com/abTEM/abTEM/compare/parity-projection-fixes-2...phonon-loss-classical-statistics)). Happy to include it in this PR: if you want it here, say so and it gets merged into `claude/parity_projection_phonons` together with the fix branch, tests included; otherwise it becomes a small PR of its own once this one is in.

TomaSusi:
Let's please include all the fixes into this PR @pzeiger :)

Could you also provide an updated API example for the new fixes?

pzeiger:
Done: everything offered is now on `claude/parity_projection_phonons`, taken as "all" to mean the three branches, since the second follow-up was offered for inclusion in exactly these words: the fix branch (`f971e480`, `72aa303f`, `f0586ddf`), the rest parity (`parity-rest-displacements`, merged as `78f94fcf`), and the classical snapshot statistics (`b08495c4`, merged as `c1e34446`), plus one integration commit (`4ecf3422`) that passes `snapshot_statistics` through the rest-parity one-phonon path. If the rest parity or the classical statistics were not meant to be in, say so and they come out again as separate PRs; both are opt-in and change nothing without their arguments.

Checked on the combined branch, not only on each piece: the seven test files the PR touches (energy-resolved parity, phonon loss, parity multislice, spectral detectors, axes grid, array, waves) pass, and the example below runs end to end.

**Updated API example.** Four ways to run the pipeline, from cheapest to most complete.

```python
import abtem
from abtem.detectors import SpectralSlitDetector
from abtem.inelastic.phonons import EnergyResolvedAtomsEnsemble
from abtem.measurements import (
    momentum_resolved_spectrum,
    phonon_loss_diffraction_patterns,
    unfold_loss_gain,
)

# snapshots: list (one per energy bin) of lists of displaced ASE Atoms, all
# displacements of eq_atoms; energies start at 0 and increase [eV].
# rest_fields: one displacement field per configuration for all modes
# outside the bin (or drawn from the full thermal ensemble), also
# displacements of eq_atoms; needed for cases 2 and 4 only.
potential_kwargs = dict(sampling=0.05, slice_thickness=eq_atoms.cell[2, 2])
pw = abtem.PlaneWave(energy=60e3)

# 1. Parity projection (two runs per snapshot): all three channels.
ensemble = EnergyResolvedAtomsEnsemble(
    snapshots, energies, equilibrium_atoms=eq_atoms, parity_projection=True
)
dp = phonon_loss_diffraction_patterns(
    pw.multislice(abtem.Potential(ensemble, **potential_kwargs)),
    max_angle=100, block_direct=True,
)
dp_all, dp_one, dp_multi = dp[0], dp[1], dp[2]      # "Phonon order" axis
dp_one_signed = unfold_loss_gain(dp_one, temperature=300.0)   # one-phonon only

# 2. Rest fields (four runs per snapshot): the one-phonon channel alone,
#    Debye-Waller damped by the other modes; temperature= is allowed here.
ensemble_rest = EnergyResolvedAtomsEnsemble(
    snapshots, energies, equilibrium_atoms=eq_atoms, parity_projection=True,
    rest_snapshots=rest_fields,
)
dp_one_damped = phonon_loss_diffraction_patterns(
    pw.multislice(abtem.Potential(ensemble_rest, **potential_kwargs)),
    max_angle=100, block_direct=True, temperature=300.0,
)

# 3. Molecular-dynamics snapshots: unfold with classical statistics.
dp_one_md = unfold_loss_gain(dp_one, temperature=300.0, snapshot_statistics="classical")

# 4. Rest fields plus the static reference (six runs per snapshot): all
#    three channels, damped.
ensemble_full = EnergyResolvedAtomsEnsemble(
    snapshots, energies, equilibrium_atoms=eq_atoms, parity_projection=True,
    rest_snapshots=rest_fields, rest_static_reference=True,
)
dp_full = phonon_loss_diffraction_patterns(
    pw.multislice(abtem.Potential(ensemble_full, **potential_kwargs)),
    max_angle=100, block_direct=True,
)

detector = SpectralSlitDetector(width=4.0, q_min=0.0, q_max=80.0)   # mrad
spectrum = momentum_resolved_spectrum(dp_one_damped, detector).compute()
spectrum.show(logscale=True)
```

What each call returns:

| case | runs per snapshot | `phonon_loss_diffraction_patterns` returns | `temperature=` |
|---|---|---|---|
| 1, parity only | 2 | "Phonon order" axis (all, one, multi) | raises; use `unfold_loss_gain` on the one slot |
| 2, rest fields | 4 | one-phonon channel, plain `DiffractionPatterns` | allowed |
| 4, rest fields + static reference | 6 | "Phonon order" axis (all, one, multi), damped | raises, as in 1 |

`snapshot_statistics="classical"` applies wherever `temperature=` is used, on both `phonon_loss_diffraction_patterns` and `unfold_loss_gain`; the default `"quantum"` is unchanged. `all == one + multi` holds exactly in cases 1 and 4.

TomaSusi:
Thanks @pzeiger — reviewed the combined branch and pushed two small fixes, a `dev` merge and a duplication cleanup; head now `b18f5d34`.

**Review of the merged work.** The physics holds up. I checked the classical weights numerically rather than by eye: detailed balance `loss/gain = exp(E/k_BT)`, the `x coth x` sum, the conversion of a classical `1/x` intensity into `(n+1)` and `n`, and both limiting forms all hold to machine precision across 10–1000 K. The rest-parity decomposition is right too — averaging over the `("plus","minus")` axis keeps the part even in `u_rest` and cancels the odd mis-binned term, and the `static` member is the correct per-realization reference for removing the `O(u_rest⁴)` two-rest-phonon fluctuations. `test_rest_parity_damps_one_phonon_channel` checking the ratio per-pixel against `exp(−Q²σ²)`, asserting the residual is a small *positive* bias and that total weight is not conserved, is exactly the right assertion set.

**Two defects found, both small, fixed in `4f210f84`.**

1. The `N ≥ 2` guard fired before the one-phonon-only branch, so the four-run rest-parity mode rejected a single configuration per energy — citing the multi-phonon channel, which that mode does not compute. Only the variance needs `N ≥ 2`; `mean_j |FT psi_odd_j|²` is well defined for one configuration. Now scoped with `and not one_phonon_only`, mirroring how the non-parity path already scopes its own guard to `component in ("tds", "all")`.

2. `snapshot_statistics` was validated only inside `_loss_gain_weights`, which the unfolding alone reaches, so with `temperature=None` a misspelled name was accepted and silently ignored (`snapshot_statistics="clasical"` returned a normal result). Now validated at both public entry points, next to the existing `component` check and for the same stated reason.

New tests cover both: a single-configuration case for the four-run mode that also asserts the guard still fires for the two modes that do form the variance, and a `temperature=None` case added to `test_invalid_statistics_rejected`.

**`dev` merged in** (`7a8a7aab`). The branch was 140 commits behind, so CI had been green against a stale base. One conflict, in `test/conftest.py`: two unrelated additions landing adjacent at end of file — dev's GPU `xdist_group` logic belongs inside `pytest_collection_modifyitems`, this branch's figure-closing fixture is a top-level autouse fixture that follows it. Both kept; dev's fixture deduplication (#426) does not provide an equivalent, so ours is still needed.

Worth noting that `cc7294c8` on dev — asserting non-size-1 dimensions match before reshaping a per-block result in `WavesBuilder` — lands directly on the machinery this PR stresses with its extra leading axes. It's a useful safety net here rather than a risk; the lazy parity tests pass under it.

Full suite on the merge: **1855 passed, 863 skipped, 0 failed** (skips are GPU tests with no device available, plus `--runslow`). CI went green on all four Python versions — the first green run against a current `dev` base.

**PR description rewritten.** The old one documented `exit_waves.static_exit_wave` at length, including four references in the test plan — that attribute was removed when the multi channel became the even-part variance (`grep -rn static_exit_wave abtem/` is now empty, and a test asserts its absence). It also predated all three new opt-in arguments. The description now covers the four API cases, both new physics sections, and the corrected memory story.

**Duplication cleanup** (`b18f5d34`). Two leftovers from the earlier review round, both in `measurements.py`:

- `_find_fp_and_energy_axes()` replaces the `FrozenPhononsAxis`/`EnergyLossAxis` scan that existed in two copies. It is a lookup only, returning `None` for an absent axis, with the raising left to the callers — the two error messages are genuinely different and both worth keeping (the non-parity path names each axis separately and suggests `ensemble_mean=False`; the parity path reports them together and relative to the parity axis). The docstring records that, so the two do not get merged into one vaguer message later.
- `_require_complex_waves()` replaces the identical `np.iscomplexobj` check and error string at both sites.

This trades ~20 lines for single-sourcing the lookup, which was the original point: a change to how these axes are located would otherwise only reach one copy. The extraction orphaned `FrozenPhononsAxis` from both local import lists; removed. `ruff check` is back to the file's pre-existing warnings with none added, and the full suite is unchanged at 1855 passed / 863 skipped.

On scope: keeping the rest parity and the classical statistics in, as offered — both are opt-in and inert without their arguments.

🤖 Written by [Claude Code](https://claude.com/claude-code) on behalf of [Toma](https://github.com/tomasusi)

pzeiger:
How do the figures look like now that we have made those changes? I would hope that the streaks at Gamma points are gone in the multiphonon background component.

TomaSusi:
I’ve been running a multi-day 3DED production run on our GPU workstation so did not get around to testing yet, but I hope tomorrow 🤞

TomaSusi:
image
image

I got the pipeline working, but neither the 0K nor the temperature-unfolded 300K one-phonon plots no longer show a clear separation between the multiphonon channels.

I suspect my modification of the rattling code to get the `rest_fields` isn't correct, maybe you could share more of that part of your modified workflow?

PR #375, opened 2026-09-08
Core-loss hot path: stop redundant per-chunk work in the scatter/filter path

pzeiger:
Hi Toma!

While profiling production core-loss scans I found several independent inefficiencies in the `TransitionPotentialArray` scatter/filter path. They share one theme — work that is identical for every scatter chunk was being redone per chunk — so they come as one PR.

A note on scope: this branch originally also fixed `TransitionPotential.build()` silently upcasting to complex128 (I had measured it doubling the built array to 31 MB and the serialized dask graph to 33.5 MiB on a production hBN scan); #364 has since fixed exactly that on dev, so this PR keeps only the end-to-end tests for it — built dtype follows the `precision` configuration through the real gpaw pipeline in both directions, `scatter()` preserves the waves' precision, and the float32 build matches a float64 reference to within round-off.

## 1. The local potential was re-uploaded to the device per chunk

`filter_sites` runs once per site chunk and re-uploaded the full-grid local potential from host to device on every call — hundreds of transfers per task in a chunked scan. The device copy is now cached on the object (`_local_potential_on_device`), keyed by device (including the GPU device id, for processes driving several GPUs), and excluded from pickle via `__getstate__` so it cannot ride into dask task graphs or break CPU-only workers. `absolute_threshold` had the same pattern one level up — it re-derived the local potential *and* re-uploaded it on every task — and now uses the same cache (its hard-coded `complex64` was already fixed by #364).

## 2. Sites were filtered once per scatter chunk instead of once per slice

`generate_scattered_waves` applied the threshold filter inside every scatter chunk. The filter copies its survival mask back to the host, which synchronises the device — for a production core-loss scan, ~1400 synchronisations per task. The threshold is applied per site and does not depend on how the sites are chunked, so the surviving set is identical when filtered once up front; the per-chunk filter is then skipped entirely. The grid/accelerator match now runs before the filter, keeping the immediate, informative mismatch error the scatter path used to raise first.

## Also fixed along the way

An adversarial review of the branch caught two latent issues in the touched functions, both fixed with synthetic (gpaw-free) regression tests: `filter_sites` indexed the caller's object with a mask computed over the element-validated subset, raising `IndexError` for mixed-element `Atoms` inputs; and the new gpaw-dependent test files now follow the suite's skip convention so CI (which installs no gpaw) skips instead of erroring.

## Testing

New `test/test_core_loss.py`: the precision pipeline tests described above. New `test/test_site_filtering.py`: filtering once for the whole set selects exactly the same sites as the old per-chunk filtering across `max_batch` settings, and the scan results are identical. New synthetic tests in `test/test_ionization.py`: mask/sites alignment for mixed-element `Atoms`, and the device cache surviving use but not pickle. Full local run against dev @ `a626b5a3`: the core-loss (including the new `test_core_loss_fixes.py` from #364), ionization, phonon-loss and multislice-input suites pass; with gpaw absent the gpaw-dependent tests skip cleanly (verified by blocking the import).

Best,
Paul

### DIFF
--- a/abtem/inelastic/core_loss.py
+++ b/abtem/inelastic/core_loss.py
@@ -28,7 +28,11 @@ except ImportError:
 
 from abtem.array import ArrayObject
 from abtem.core.axes import AxisMetadata, OrdinalAxis
-from abtem.core.backend import copy_to_device, get_array_module
+from abtem.core.backend import (
+    copy_to_device,
+    device_name_from_array_module,
+    get_array_module,
+)
 from abtem.core.chunks import validate_chunks
 from abtem.core.complex import abs2, complex_exponential
 from abtem.core.electron_configurations import electron_configurations
@@ -908,6 +912,7 @@ class TransitionPotentialArray(ArrayObject, BaseTransitionPotential):
         )
 
         self._local_potential = self.local_potential(space="real").sum(0)
+        self._local_potential_device_cache = None
         self._threshold = None
 
     def from_array_and_metadata(self, array, metadata):
@@ -1026,6 +1031,21 @@ class TransitionPotentialArray(ArrayObject, BaseTransitionPotential):
         sites = np.array(sites, dtype=get_dtype())
         return sites
 
+    def _local_potential_on_device(self, like):
+        """The local potential, resident on the device of ``like``.
+
+        Cached per device so repeated calls reuse the upload; the cache is
+        dropped when the local potential itself is recomputed.
+        """
+        device = device_name_from_array_module(get_array_module(like))
+        cache = getattr(self, "_local_potential_device_cache", None)
+        if cache is not None and cache[0] == device:
+            return cache[1]
+
+        on_device = copy_to_device(self._local_potential, like)
+        self._local_potential_device_cache = (device, on_device)
+        return on_device
+
     def filter_sites(self, waves, sites, threshold):
         if hasattr(waves, "build"):
             waves = waves.build(lazy=False)
@@ -1040,7 +1060,11 @@ class TransitionPotentialArray(ArrayObject, BaseTransitionPotential):
                 (validated_sites / xp.array(self.sampling))
             ).astype(int)
 
-            local_potential = copy_to_device(self._local_potential, waves.array)
+            # filter_sites runs once per site chunk, so re-uploading the
+            # local potential here costs a full-grid host-to-device transfer
+            # hundreds of times per task. Keep the device copy on the object,
+            # as scatter() already does for the transition potentials.
+            local_potential = self._local_potential_on_device(waves.array)
 
             # Stream the overlap reduction over sites in chunks. The full
             # (n_sites, *waves_shape, H, W) tensor that the naive computation
--- a/abtem/inelastic/core_loss.py
+++ b/abtem/inelastic/core_loss.py
@@ -1196,6 +1196,18 @@ class TransitionPotentialArray(ArrayObject, BaseTransitionPotential):
     ):
         sites = self.validate_sites(sites)
 
+        # Filter once for the whole set rather than inside every chunk.
+        # filter_sites copies the mask back to the host, which synchronises
+        # the device; doing that per chunk cost one synchronisation per chunk
+        # -- for a production core-loss scan, ~1400 per task. The threshold is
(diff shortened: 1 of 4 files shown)


TomaSusi:

A few findings worth a look before merge:

1. **`abtem/inelastic/core_loss.py:1050`** (`_local_potential_on_device`) — the CUDA device-id lookup is wrapped in a bare `except Exception: pass`, which on failure collapses the per-device cache key back to a single `"gpu"` key. That reopens the cross-GPU cache-collision bug this per-device keying is meant to prevent: in a single process driving multiple GPUs, a transient error on `.id` could cause a call on GPU 1 to receive back an array resident on GPU 0.

2. **`abtem/inelastic/core_loss.py:1220`** (`generate_scattered_waves`) — this adds a `grid.match`/`accelerator.match`/`check_is_defined()` pass up front, but `scatter()` still repeats the identical calls per chunk inside the loop. Net effect is N+1 calls instead of N, which cuts against the PR's stated goal of removing redundant per-chunk work.

3. The single-slot per-object device cache will thrash with no benefit if a threaded scheduler interleaves calls for different GPUs on the same `TransitionPotentialArray` instance (correctness is fine — each read checks the device key — but the optimization provides no speedup in that topology).

4. The new cache test (`test_local_potential_device_cache_survives_use_but_not_pickle`) runs entirely on CPU and doesn't exercise multi-GPU keys or the exception fallback in (1) — worth a test if feasible.

Separately: the `filter_sites` mask-alignment fix in this PR looks like a genuine, well-tested bug fix (reproduced the `IndexError` on `dev` with mixed-element `Atoms` and confirmed the PR fixes it), not just cleanup — nice catch.

pzeiger:
Thanks for the careful look!

1. **Bare-except device fallback — you're right, fixed in 59efb05b.** The current-device lookup and its fallback are gone entirely: the key is now read off the array the caller passes (`like.device.id`, a plain attribute with no failure mode to fall back from), and the upload is allocated under that device. This also removes the subtler variant of the same bug — keying on the current-device *context*, which can differ from the array's device under multi-GPU use.

2. **The extra `grid.match` in `generate_scattered_waves` — deliberate, and cheaper than it looks.** The up-front match runs once per *task*, not per chunk (the per-chunk calls inside `scatter()` are unchanged from dev), so the net cost is one additional set of idempotent µs-scale comparisons per multi-second task. It's needed for correctness ordering: the hoisted `filter_sites` reads `self.sampling` before any scatter call, so without the early match a grid mismatch would produce a wrong-scaled filter or an opaque broadcast error instead of the informative mismatch error the scatter path raises. `scatter()` keeps its own guard because it's independently callable. Happy to restructure if you feel strongly, but I'd argue the constant is the right price for the error behaviour.

3. **Single-slot cache under a threaded multi-device scheduler — acknowledged, deliberate trade-off.** The supported multi-GPU execution path is one process per GPU (dask-cuda, and `is_gpu_dask_client` explicitly rejects multi-threaded workers for CuPy work), where each process only ever sees one device and the single slot is always warm. In the unsupported threaded-multi-device topology the cache stays correct and merely stops helping, which seems acceptable for a code path abTEM doesn't otherwise support.

4. **Test coverage — partially addressed in 59efb05b.** A new test asserts the GPU key carries the concrete device id and that a cpu request replaces the slot (runs on any CuPy box, including ROCm). The exception-fallback test is moot now that there is no fallback branch. A genuine two-GPU key test needs multi-GPU hardware — it's on the cluster verification list for the A100 node runs.

And thanks for independently confirming the `filter_sites` mask fix reproduces on dev — good to have that double-checked.

TomaSusi:
Checked `59efb05b` against the code — the device-id fallback is genuinely gone (keyed on `like.device.id`, no exception path left to fall through), and the new test exercises the GPU key correctly.

On the two remaining points: confirmed `filter_sites` reads `self.sampling` directly, so the early `grid.match` in `generate_scattered_waves` is load-bearing for correctness, not redundant — and `is_gpu_dask_client` does reject multi-threaded workers for CuPy, so the single-slot cache's blind spot is outside abTEM's supported execution model. Both hold up.

Resolving — thanks for the thorough responses.


TomaSusi (approved):

PR #376, opened 2026-09-08
GPU ROCm portability

pzeiger:
Hi Toma,

Running the full GPU test suite on a ROCm machine (Strix Halo gfx1151, ROCm 7.2.0, amd-cupy 13.5.1) surfaced three independent defects in the GPU code path. Two affect every CuPy-HIP platform and one affects CUDA as well. Each gets its own commit; 532 GPU-parameterized tests now pass on that box, up from two hard failure classes plus an interpreter-killing segfault. The branch is rebased on current dev (a626b5a3).

**1. `cupyx.scipy.signal` was never importable through `get_scipy_module` (all platforms, CUDA included).** `cupyx.scipy` exposes its submodules lazily, so `get_scipy_module(array).signal` in the Lorentzian/Voigtian image filters raised `AttributeError` unless something else had already imported `cupyx.scipy.signal`. The fix registers the submodule in `abtem.core.backend` next to the existing `cupyx.scipy.ndimage` import, with the `FutureWarning` that import emits (about the experimental `cupyx.jit` interface it uses internally) suppressed. Fixes `test_lorentzian_filter_images[gpu-*]` and `test_voigtian_filter_images[gpu-*]`.

**2. The real-space multislice Laplacian was abTEM's only `numba.cuda` dependency, and `numba.cuda` cannot initialize on non-CUDA platforms.** The `@cuda.jit` stencil in `abtem/finite_difference.py` is replaced with a `cp.RawModule` kernel following the pattern of `abtem/core/_cuda.py`; CuPy compiles it via NVRTC on CUDA and hipRTC on ROCm alike, so the module no longer imports `numba.cuda` at all. The kernel reproduces the previous stencil to machine precision (checked against the untouched CPU `numba.njit` stencil for complex64/complex128, accuracy orders 2–16, and 2-D through 4-D batch shapes; relative error ≤3e-7 single, ≤6e-16 double). It also switches the launch layout so the fast image axis maps to `threadIdx.x` (coalesced access; the numba kernel had the batch index on x) and strides the grid over the batch so any batch size is supported. Fixes all 8 `test_realspace_multislice` GPU failures on ROCm.

**3. PRISM S-matrix compression segfaulted the interpreter on HIP via `xp.argsort`.** CuPy's sort family (`sort`, `argsort`, `lexsort`, `unique`) is backed by precompiled Thrust device code, which the amd-cupy wheels ship only for MI-series and a subset of RDNA architectures — on an unsupported device the call segfaults at kernel launch inside `libamdhip64` rather than raising, taking pytest down with it. The argsort in `_compress` orders one scalar norm per retained mode, so it now runs on the host: cost is negligible at that size, and it removes the only Thrust dependency in the GPU path. Un-crashes `test_prism_upsample` (78 tests) on ROCm.

**Testing.** On the ROCm box, after the rebase onto a626b5a3: suite-wide `-k gpu -m "not multigpu"` → 532 passed, 152 skipped (`--runslow` gating), 0 failed; full runs of `test_realspace_multislice.py`, `test_measure.py`, and `test_prism_upsample.py` (368 tests, CPU + GPU) all green. A CPU-only environment was simulated by blocking `cupy`/`cupyx` imports: abTEM imports cleanly and the CPU stencil is unaffected. I don't have a CUDA card in this machine, so the RawModule kernel has only run under hipRTC here — the machine-precision equivalence check is dtype- and layout-exhaustive, but a CI or maintainer run on CUDA would be a welcome double-check.

No API or dependency changes; `numba` remains required for the CPU stencil.

Best,
Paul

### DIFF
--- a/abtem/core/backend.py
+++ b/abtem/core/backend.py
@@ -43,6 +43,18 @@ except ImportError:
     cupyx_ndimage = None
 
 
+try:
+    # cupyx.scipy exposes submodules lazily; signal must be imported explicitly
+    # before ``get_scipy_module(...).signal`` can resolve it (the import emits a
+    # FutureWarning about the experimental cupyx.jit interface it uses
+    # internally)
+    with warnings.catch_warnings():
+        warnings.simplefilter("ignore", FutureWarning)
+        import cupyx.scipy.signal  # type: ignore  # noqa: F401
+except ImportError:
+    assert cupyx is None
+
+
 ArrayModule = Union[ModuleType, str]
 
 logger = logging.getLogger(__name__)
--- a/abtem/finite_difference.py
+++ b/abtem/finite_difference.py
@@ -5,10 +5,10 @@ from typing import TYPE_CHECKING, Callable, Sequence
 
 import numpy as np
 import scipy.ndimage  # type: ignore
-from numba import cuda, njit  # type: ignore
+from numba import njit  # type: ignore
 
 from abtem.antialias import AntialiasAperture
-from abtem.core.backend import get_array_module
+from abtem.core.backend import check_cupy_is_installed, cp, get_array_module
 from abtem.core.energy import energy2sigma, energy2wavelength
 from abtem.core.utils import get_dtype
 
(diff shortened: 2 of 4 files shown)


TomaSusi:

Most important: a likely correctness bug, plus a few other things worth checking before merge.

1. **`abtem/finite_difference.py:297`** — the GPU kernel dispatch picks the complex64 kernel for any dtype that isn't exactly `complex128`, rather than validating the input is complex at all. Since `_laplace_operator_stencil` accepts arbitrary dtypes (the CPU/njit path handles any dtype generically), calling it on the GPU with a real float32/float64 array would run the complex64 kernel against a real buffer — reinterpreting pairs of floats as one complex value, which corrupts memory or reads out of bounds instead of raising a clear error. This is a regression vs. the previous `numba.cuda.jit` path, which was dtype-generic via JIT specialization.

2. **`abtem/prism/s_matrix.py:4644`** — the ROCm workaround for the missing Thrust-backed `argsort` forces an unconditional device→host→device round trip on *every* backend, including native CUDA where `cp.argsort` works fine. Gating this on a HIP/CUDA check (e.g. `cupy.cuda.runtime.is_hip`) would avoid a needless sync + copy on CUDA, and centralizing it at a shared dispatch point would cover any other GPU sort call elsewhere in the codebase too.

3. The rewritten GPU kernel path (`numba.cuda.jit` → hand-written `cp.RawModule` CUDA/HIP kernel) has no test coverage — `TestStencilNumericalAccuracy` only runs `device="cpu"`. Given (1), a GPU-side numerical test against the scipy reference seems worth adding.

4. Minor: the lazy kernel-cache init (`_laplace_stencil_c64`/`_laplace_stencil_c128`) is two sequential non-atomic global assignments guarded only by a check on the first variable — a possible (if narrow) race under dask's threaded scheduler on first use.

5. Also worth double-checking: the raw kernel hardcodes NVRTC-style compiler options (`--std=c++14`) and includes `cupy/complex.cuh` — unclear from the diff whether ROCm's `hipRTC` build accepts the same flags/headers; the missing GPU test in (3) would catch this at import/first-use time.

pzeiger:
Thanks for the review! Point-by-point, with the fixes in fa3bb182 and d89655c1:

**1. Accepted — fixed in fa3bb182.** The dispatch now selects strictly on complex64/complex128 and raises `TypeError` for anything else. For context, the path is unreachable through abTEM's own callers (`LaplaceOperator` always builds the stencil with `get_dtype(complex=True)` and applies it to wave arrays), but you're right that the module-level function is dtype-generic in signature and silently reinterpreting a real buffer would have been nasty for any direct use — the old numba path was indeed dtype-generic via JIT specialization.

**2. Not taken, with numbers.** The sorted vector holds one scalar norm per retained mode (typically 10²–10³ entries), and `_compress` runs once per S-matrix build. Measured on this box, the entire round trip — norm reduction, device→host copy, `np.argsort`, upload — is 184 µs for 256 modes, sitting next to dense `eigh`/SVD calls on the full matrices that take milliseconds to seconds. Gating on `is_hip` would save an unmeasurable amount on CUDA while making the two platforms take different code paths through `_compress`, which is worse for testability than the uniform host sort. On centralizing: I audited the codebase for other GPU-reachable sorts — the only other `xp`-generic `.argsort` (`label_to_index` in `core/utils.py`) receives numpy arrays at every call site (measurements, slicing, atoms, visualize), so there is currently no second call site for a shared dispatch point to cover. Happy to revisit if one appears.

**3. Accepted — d89655c1.** `TestStencilNumericalAccuracy` now runs the scipy-reference comparison on both devices (the GPU leg is skipped where CuPy is absent), so the RawModule kernel is exercised through NVRTC on CUDA and hipRTC on ROCm, and a new test asserts the GPU stencil rejects non-complex arrays.

**4. Accepted — fa3bb182.** The guard variable is now assigned last, so a thread that observes it initialized also observes the second kernel; a comment notes that redundant compilation from two racing threads is harmless since CuPy caches the module.

**5. Resolved empirically.** `--std=c++14` + `cupy/complex.cuh` is the established combination from `abtem/core/_cuda.py` already on dev, and this branch was developed and validated on ROCm 7.2.0 — the full GPU suite (532 tests) runs through hipRTC, including this kernel at machine precision against the CPU stencil. CuPy translates NVRTC-style options for hipRTC and ships `complex.cuh` for both backends. With (3) in place, compile-and-run is now covered by CI on any GPU runner going forward.

Best,
Paul


🤖 Written by [Claude Code](https://claude.com/claude-code) on Paul's behalf.

TomaSusi:
Confirmed `fa3bb182` and `d89655c1` do what the writeup says: strict complex64/complex128 dispatch with a `TypeError` otherwise, the guard-variable reordering that actually fixes the init race, and `TestStencilNumericalAccuracy` now genuinely running on GPU (traced the `gpu`/`to_numpy` helpers — real, not new fixtures papering over a skip).

On the argsort call: fair tradeoff, and the `label_to_index` audit checks out — that's the only other GPU-reachable argsort and it only ever sees numpy arrays. No objection.

Resolving this thread.


TomaSusi (approved):

PR #378, opened 2026-09-08
Cache per-chunk GPU work: cuFFT plan entries and polar detector-bin geometry

pzeiger:
Hi Toma!

Second installment of the core-loss profiling series (after #375): two caches for work that is identical for every wave-function chunk but was rebuilt on each one. Both surfaced profiling core-loss scans but pay off for any chunked STEM workload — plain HAADF included.

## 1. cuFFT plan-cache entry limit (`cupy.fft-cache-entries`, default 64)

CuPy's plan cache keeps at most 16 plans. Workloads whose FFT batch dimension varies — a core-loss scan's scattering batches follow the number of sites passing the threshold — cycle through more than 16 distinct transform shapes within a few chunks and then rebuild plans continuously: profiling put ~30 % of a production scan's runtime in `_get_cufft_plan_nd`, and raising the limit made the same scan 18 % faster with identical results. The limit is now a config key applied alongside the existing `fft-cache-size` bound in `_configure_cufft_cache`: the entry count and the (#346) memsize bound compose, `0 MB` still disables the cache outright, and the applied-state token includes the entry count so runtime config changes take effect. The count is **raised, never lowered** — a cache someone tuned larger through CuPy's own API keeps its size, and `null` or a non-positive value opts out of touching the count entirely (invalid values never raise; this runs ahead of every GPU FFT dispatch). The applied-state token is thread-local, because CuPy's plan cache is per thread: the previous process-global token (from #346) configured only the first dispatching thread and left every other dask worker thread at CuPy's defaults — so this also makes the existing memsize bound actually apply on all threads. Frozen-phonon ensembles hit the same thrashing from ensemble batching, so plain imaging benefits too. (Known limitation, deliberate: the entry count treats the symptom — the root cause is that scattering batch sizes are data-dependent; bucketing them to repeating values, as `estimate_scan_batch_size` already does elsewhere, would bound the distinct shapes and is left as a follow-up.)

## 2. Polar detector-bin geometry cached instead of rebuilt per call

`_polar_detector_bins` builds full-grid polar coordinates on the host and, in `DiffractionPatterns._batch_integrate`, ships the resulting index arrays to the device — once per wave-function chunk, though the result depends only on the detector geometry and the grid. For a core-loss scan applying a `FlexibleAnnularDetector` per chunk, profiling put ~48 % of the runtime here. Two layers of caching fix it: the host-side geometry is `lru_cache`d (8 entries — they hold full-grid arrays, so the bound is deliberately small; returned containers are immutable tuples of read-only arrays), and the flat bin indices + separators are kept resident per (geometry, device) in a real LRU (8 entries, least-recently-used eviction), so the host-to-device copy happens once per geometry instead of once per chunk. Both caches key through one canonical `_polar_bins_key` helper, so the keys cannot drift apart, and the device cache keys on the device the diffraction array actually lives on (`array.device`), not the current-device context — safe under multi-GPU use.

One heads-up for the skewed-grids work (#282/#352): those branches add a `cell` argument to `_polar_detector_bins`, and an `ase.Cell` is not hashable, so the `lru_cache` key will need the cell flattened to a tuple when they rebase — happy to coordinate.

## Adversarial review

The branch went through the same adversarial review round as #375; the hardening commit addresses what it caught: `null`/`-1` for the new config key would have crashed every GPU FFT dispatch; the entry count clobbered externally tuned cache sizes; the state token being process-global left other threads unconfigured (a pre-existing #346 gap this branch now fixes); the device cache trusted callers to pass a matching key/indices pair, keyed on the current device instead of the array's, cleared wholesale at its bound, and retained more host memory than needed.

## Testing

New `test/test_fft.py` test: the entry limit follows the config (explicit values, grow-only, opt-out and invalid values, `0 MB` still wins), with try/finally restoration. New `test/test_detector_bin_cache.py`: cached geometry is identical to a fresh build and distinct geometries are not confused; repeat detects are served from the device cache; LRU eviction keeps hot entries; detector results are unchanged. Full local runs of the fft, measure, detect and ionization suites pass; an A/B benchmark on identical workloads gives bit-identical checksums with the caches on and off.

Best,
Paul

### DIFF
--- a/abtem/core/abtem.yaml
+++ b/abtem/core/abtem.yaml
@@ -42,6 +42,10 @@ cupy:
   #          which on Bluestein-fallback grid sizes can reach tens of GB).
   # null  — same as -1 (no bound).
   fft-cache-size: auto
+  # Number of cuFFT plans kept. CuPy's own default is 16, which thrashes for
+  # workloads with many distinct transform shapes (e.g. core-loss scattering,
+  # whose batch dimension follows the number of contributing sites).
+  fft-cache-entries: 64
 mkl:
   # The number of threads to use for mkl
   threads: 2
--- a/abtem/core/abtem.yaml
+++ b/abtem/core/abtem.yaml
@@ -42,9 +42,12 @@ cupy:
   #          which on Bluestein-fallback grid sizes can reach tens of GB).
   # null  — same as -1 (no bound).
   fft-cache-size: auto
-  # Number of cuFFT plans kept. CuPy's own default is 16, which thrashes for
-  # workloads with many distinct transform shapes (e.g. core-loss scattering,
-  # whose batch dimension follows the number of contributing sites).
+  # Minimum number of cuFFT plans kept. CuPy's own default is 16, which
+  # thrashes for workloads with many distinct transform shapes (e.g. core-loss
+  # scattering, whose batch dimension follows the number of contributing
+  # sites). The count is raised to this value, never lowered — a cache tuned
+  # larger through CuPy's own API is left alone. null or a non-positive value
+  # leaves the entry count untouched entirely.
   fft-cache-entries: 64
 mkl:
   # The number of threads to use for mkl
--- a/abtem/core/fft.py
+++ b/abtem/core/fft.py
@@ -366,8 +366,36 @@ U = TypeVar("U", np.ndarray, da.core.Array)
 
 # Cache the parsed cuFFT cache limit + the config value it was derived from.
 # Revalidated on every GPU FFT dispatch without reparsing when the config is
-# unchanged (the common case in a hot loop).
-_CUFFT_CACHE_STATE: tuple[object, int] | None = None
+# unchanged (the common case in a hot loop). CuPy's plan cache is per thread
+# (and per device), so the applied state must be thread-local as well: a
+# process-global slot would configure only the first dispatching thread and
+# leave every other dask worker thread's cache at CuPy's defaults.
+_CUFFT_CACHE_STATE = threading.local()
+
+
+def _reset_cufft_cache_state():
+    """Forget the applied plan-cache state for the calling thread (tests)."""
+    for attr in ("token", "limit"):
+        try:
+            delattr(_CUFFT_CACHE_STATE, attr)
+        except AttributeError:
+            pass
+
+
+def _parse_cufft_cache_entries() -> int:
+    """The configured plan-cache entry count, or 0 to leave the count alone.
+
+    Invalid values must not raise: this runs ahead of every GPU FFT, and an
+    exception here (e.g. ``int(None)``) would fail every dispatch. null and
+    non-positive values mean "do not touch the entry count", mirroring how a
+    user opts out of the sibling ``fft-cache-size`` bound.
+    """
+    raw = config.get("cupy.fft-cache-entries", 64)
+    try:
+        entries = int(raw)
+    except (TypeError, ValueError):
+        return 0
+    return entries if entries > 0 else 0
 
 
 def _configure_cufft_cache():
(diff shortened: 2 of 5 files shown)


TomaSusi:

Findings, most severe first:

1. **`abtem/measurements.py:400`** — the device-array cache evicts via `while len(cache) >= MAX: cache.popitem(last=False)` on a plain `OrderedDict`, mutated without a lock. Under dask's threaded scheduler processing several distinct detector geometries concurrently, multiple threads can each read a stale `len() >= 8` before any of them pops, then issue more `popitem(last=False)` calls than there are entries — the last one raises `KeyError` on an emptied dict, crashing a worker. `functools.lru_cache`, already used two lines below for the sibling host-side cache, has an internal lock and would avoid this (and simplify the code — see 3).

2. **`abtem/measurements.py:317`** — the docstring says the memory bound is per-device, but `_RADIAL_BINNING_DEVICE_CACHE_MAX = 8` is actually one global cap shared across *all* devices (device id is only part of the cache key, not a separate slot pool per device). On multi-GPU runs, activity on one GPU evicts another GPU's warm entries, giving each device less effective retention than documented.

3. **`abtem/measurements.py:330`** — this hand-rolls a bounded-LRU cache in the same module that already uses `functools.lru_cache(maxsize=8)` for the sibling cache a few lines below. Wrapping the same `cache_key` tuple with `lru_cache` would remove the manual bookkeeping and fix (1) as a side effect.

4. **`abtem/core/fft.py:428`** — the cuFFT cache-config short-circuit only detects drift caused by abTEM's own config; if something else in the process calls CuPy's `cache.set_size()`/`set_memsize()` directly between two abTEM-triggered configurations, that drift goes undetected until abTEM's own config value changes. Likely a narrow edge case, flagging for awareness.

5. Minor: `raw`/`entries` in `_configure_cufft_cache` are computed via `config.get()` before the cache-hit early return, so every GPU FFT dispatch pays the lookup cost even on the hot/unchanged path this mechanism is meant to short-circuit.

On the positive side: the cache key (`_polar_bins_key`) does include every geometry parameter that affects the cached output plus device id, and all external callers of `_polar_detector_bins` were checked and only read the (now-cached, read-only) result — no in-place mutation of a shared cached array found.

pzeiger:
Thanks — finding 1 was a real race.

1. \+ 3. **Eviction race — confirmed and fixed in 673e88bf, taking your suggestion.** Both check-then-act pairs were racy (`get`/`move_to_end` as well as the `len`/`popitem` eviction you called out). The hand-rolled `OrderedDict` is gone: the device arrays now come from a `functools.lru_cache`d builder keyed on the canonical geometry key plus the array's device — same locking and LRU semantics as the sibling host cache, and the manual bookkeeping deleted. A regression test hammers the cache from 8 threads with more distinct geometries than the bound.

2. **Per-device wording — fixed.** The docstring now states the bound is a single cap shared across devices. In the supported one-process-per-GPU layout each process only ever sees one device, so the distinction is immaterial there; a threaded multi-device process would share the 8 slots, which the docstring now says honestly.

3. (folded into 1.)

4. **External `set_size` drift going undetected — by design after the grow-only change.** Since abTEM now deliberately never lowers an externally tuned entry count, treating direct CuPy calls between abTEM configurations as user intent (rather than drift to be corrected) is the consistent behaviour: fighting it would reintroduce exactly the clobbering the grow-only semantics removed. The one case abTEM does re-assert is its own config changing, including `0 MB` disable, which re-applies on the next dispatch.

5. **Config lookups before the early return — measured, immaterial.** `config.get` is a ~1 µs dict walk, against an FFT dispatch costing milliseconds; and reading the current raw values is the mechanism by which runtime config changes are detected at all — the token can't be compared without them. Caching them harder would just move the staleness problem into a second cache layer.

TomaSusi:
Checked `673e88bf` — the `OrderedDict` is gone, `functools.lru_cache` now owns the locking and LRU semantics, and the 8-thread hammer test exercises exactly the race that used to `KeyError`. Docstring correction on the shared cap is honest about the tradeoff.

On the two `fft.py` points: verified the "raise, never lower" logic in `_configure_cufft_cache` — abTEM really does only ever grow an externally-tuned cache size, never clobber it, so treating outside `set_size` calls as intent rather than drift is consistent with that. The lookup-cost argument for point 5 is reasonable too.

Resolving — everything here checks out.


TomaSusi (approved):

PR #379, opened 2026-09-10
Fixing consistency between Bloch and multislice for 3DED

TomaSusi:
## Summary

Fixes two confirmed Bloch-wave/multislice consistency bugs surfaced by the
[py3DED project](https://github.com/3DED/py3DED)'s independent cross-checks
between the two simulation engines.

1. **Structure-factor sign convention** (`abtem/bloch/dynamical.py`).
   `calculate_structure_factors` computed the complex conjugate of the
   standard crystallographic convention (`exp(-2pi i g.r)` instead of
   `exp(+2pi i g.r)`). This was invisible to the existing potential
   round-trip test because `structure_factor_to_potential`'s `ifftn` happened
   to cancel the wrong sign when reconstructing a real-space potential — but
   `calculate_structure_matrix` consumes `F(g)` directly, with no such
   cancellation, so Bloch-wave diffraction was inconsistent with multislice.
   Fixed the sign and switched the paired transform from `ifftn` to `fftn`
   (dropping the now-unneeded `1/N` un-normalization) so potential
   reconstruction stays correct.

2. **Non-orthogonal-cell crash** (`abtem/bloch/utils.py`, cherry-picked from
   `da8a3220` on the unmerged `skewpixels` branch — same diagnosis, already
   fixed there). `calculate_structure_matrix` needs structure-factor values
   at *pairwise differences* of the selected beam set, but the array
   `reciprocal_space_gpts` sizes for them only (just barely) covered single
   reflections, using a Cartesian-bounding-box heuristic that ignores the
   reciprocal metric tensor. Non-orthogonal cells — hexagonal/rhombohedral
   angles near 120° are the worst case — hit this reliably:
   `ValueError: invalid entry in coordinates array` from `ravel_hkl`. Fixed
   by sizing the grid from the true per-axis bound, `g_max * |a_i|`.

## Tests

Two new regression tests in `test/test_bloch_waves.py`:

- `test_structure_factor_matches_standard_crystallographic_sign_convention`
  — checks `calculate_structure_factors` against an independent hand
  computation on an asymmetric two-atom cell (the wrong sign disagrees by
  far more than float noise, so this fails loudly on regression).
- `test_bloch_waves_on_skewed_cell_at_tilt_matches_orthogonalized_supercell`
  — reproduces the crash on a hexagonal cell at a general tilt, and checks
  the result quantitatively matches (R1 < 1e-4) an orthogonalized supercell
  of the same crystal.

All existing `test_bloch_waves.py` / `test_bloch_indexing.py` tests still
pass, including `test_potential_from_structure_factor` (confirms the
`ifftn`→`fftn` pairing still reconstructs the potential correctly).

## Note on a third, related report

The same 3DED cross-check also reported a thermal-sigma (Debye–Waller)
disagreement between the two engines. That turned out to already be fixed by
#271 (anisotropic DWF support), merged before this branch — no action needed
here. A fourth reported issue (rotation-axis orientation) was investigated at
length and could not be reproduced as a code-level bug; every
orientation/rotation-matrix convention shared between the two engines was
traced and found mutually consistent.

Thanks to @pbklar for reporting these!

### DIFF
--- a/abtem/bloch/dynamical.py
+++ b/abtem/bloch/dynamical.py
@@ -1815,6 +1815,12 @@ class BlochWaves:
         else:
             array = calculate_wave_functions(values, g_vec, extent, gpts, thicknesses)
 
+            if thicknesses.ndim == 0:
+                # a scalar thickness has no ensemble (thickness) axis; drop the spurious
+                # leading size-1 axis so the array matches the empty axis metadata (the
+                # lazy path already returns a 2D array in this case).
+                array = array[0]
+
         ensemble_axes_metadata: list[AxisMetadata] = []
 
         if isinstance(thicknesses, np.ndarray) and thicknesses.ndim > 0:
--- a/abtem/bloch/dynamical.py
+++ b/abtem/bloch/dynamical.py
@@ -225,7 +225,7 @@ def calculate_structure_factors(
 
     struct_factors = (
         xp.sum(
-            f_e * xp.exp(-2.0j * np.pi * positions @ hkl),
+            f_e * xp.exp(2.0j * np.pi * positions @ hkl),
             axis=0,
         )
         / atoms.cell.volume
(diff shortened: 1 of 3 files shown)


pbklar:
The link to the 3DED project points to an outdated repository. Here is the current one: https://github.com/3DED/py3DED

TomaSusi:
> The link to the 3DED project points to an outdated repository. Here is the current one: https://github.com/3DED/py3DED

Ah thank you, now corrected.

PR #381, opened 2026-09-10
Fix cupyx signal import guard

pzeiger:
Follow-up to #376 with two fixes: one commit that missed the merge window, and a portability bug that CUDA validation on Perlmutter caught after the merge.

**1. Don't block the abtem import when `cupyx.scipy.signal` fails to load** (47abc7b3, cherry-picked from the #376 branch — it landed there minutes after the merge). The guarded import asserted that an ImportError implies cupyx itself is missing, but `cupyx.scipy.signal` eagerly imports cuBLAS-backed submodules, so it can fail in an environment where cupy imports fine and the CUDA/ROCm libraries are simply not on the loader path (observed on Perlmutter with `cupy-cuda12x` and no cudatoolkit module loaded: `libcublas.so.12` not found → the assert killed every `import abtem`, including CPU-only use). Failure now degrades gracefully: GPU filters needing `scipy.signal` raise at use time, as before #376.

**2. Stop pinning a C++ dialect in the RawModule kernels** (2dc3f4e7 + 6b7e9542). Validating #376 on an A100 (Perlmutter, CuPy 14.2.0, CUDA 12.9) caught this — and confirms the concern raised in review about the hardcoded compiler options: CuPy 14 routes `cupy/complex.cuh` through its bundled CCCL headers, which reject anything below C++17, so the Laplacian stencil failed to compile under NVRTC with `--std=c++14` ("Thrust requires at least C++17"); the amd-cupy 13.5.1 ROCm build doesn't chain `complex.cuh` into CCCL, which is why hipRTC accepted it during development. Rather than repinning (first commit bumps to C++17, second removes the pin), the kernels now pass no `--std` at all: their own source needs nothing beyond C++11, the dialect requirement comes from CuPy's bundled headers, and each backend's default tracks those headers — CuPy pins no dialect for its own kernels either, so this matches the configuration CuPy tests everywhere. The two kernels in `core/_cuda.py` are unpinned as well (they compiled on CuPy 14 only because they avoid `complex.cuh`).

A final commit adds a docstring note that `LaplaceOperator` is specialized for complex input (the Laplacian itself is a real operator; the restriction comes from the stencil coefficients being cast to the configured complex dtype).

**Validation.** ROCm 7.2.0/hipRTC (gfx1151): all three RawModule kernels compile and run without a pinned dialect; the stencil matches the CPU reference to machine precision for both complex dtypes; 156 kernel-consumer tests green. CUDA (Perlmutter A100, CuPy 14.2.0, CUDA 12.9, at 6b7e9542): stencil reference tests `9 passed`; full GPU sweep (`-k gpu -m "not multigpu"`) `537 passed, 152 skipped, 0 failed` in 6:07. The #376 GPU stencil tests are what caught (2) on first contact with real CUDA.

### DIFF
--- a/abtem/core/_cuda.py
+++ b/abtem/core/_cuda.py
@@ -60,7 +60,7 @@ _sum_rle_f64 = None
 def _init_sum_rle_kernels():
     global _sum_rle_f32, _sum_rle_f64
     if _sum_rle_f32 is None:
-        mod = cp.RawModule(code=_SUM_RLE_KERNEL, options=("--std=c++14",))
+        mod = cp.RawModule(code=_SUM_RLE_KERNEL, options=("--std=c++17",))
         _sum_rle_f32 = mod.get_function("sum_rle_f32")
         _sum_rle_f64 = mod.get_function("sum_rle_f64")
 
@@ -227,7 +227,7 @@ _interpolate_radial_f64 = None
 def _init_interpolate_radial_kernels():
     global _interpolate_radial_f32, _interpolate_radial_f64
     if _interpolate_radial_f32 is None:
-        mod = cp.RawModule(code=_INTERPOLATE_RADIAL_KERNEL, options=("--std=c++14",))
+        mod = cp.RawModule(code=_INTERPOLATE_RADIAL_KERNEL, options=("--std=c++17",))
         _interpolate_radial_f32 = mod.get_function("interpolate_radial_f32")
         _interpolate_radial_f64 = mod.get_function("interpolate_radial_f64")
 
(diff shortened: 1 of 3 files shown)


TomaSusi:
Whoops sorry for being too fast :) Might make sense to mark PRs as drafts if you're still working on them!

pzeiger:
No, that was my bad. I thought it was ready and might have trusted Claude too much. I uncovered the issue fixed here when running the test suite on Nvidia GPUs (my testing was exclusively on my local AMD GPU) this morning and I then wanted to change the PR to a draft, but you merged it one minute before ^^. I will be more careful next time.

TomaSusi:
No harm done :) I'll review and merge this now!

PR #382, opened 2026-09-10
Raise a clear error before OOM in Bloch-wave eigendecomposition

TomaSusi:
## Summary
- Large unit cells at high `g_max` (e.g. triclinic structures like kyanite) can select tens of thousands of beams, making the dense structure matrix passed to `eigh()` in `calculate_dynamical_scattering` too large for available VRAM/RAM — this previously surfaced as a raw, confusing CUDA/numpy `OutOfMemoryError` deep inside the solver.
- Adds `check_eigh_memory()` in `abtem/bloch/dynamical.py`, called just before `eigh`. It estimates peak bytes needed (matrix + eigenvectors + solver workspace, ~2.5x the matrix size) from the beam count and dtype, and compares against live free memory (`cp.cuda.Device().mem_info()` on GPU, `psutil.virtual_memory().available` on CPU, softly skipped if `psutil` isn't installed).
- On insufficient memory, raises `MemoryError` naming the beam count and suggesting a lower `g_max`/`sg_max`, instead of letting the raw allocation failure propagate.

Found while cross-validating PR #379 against real production-scale 3DED simulations (py3DED project): a kyanite (Al2SiO5, triclinic) structure at `g_max=8.0`, `sg_max=1.0` selected 28,898 beams and crashed with a 13.36GB CUDA allocation failure inside `eigh` — matching `N**2 * 16 bytes` for `N=28898`. This is a real scaling wall for large unit cells / high `g_max`, not a correctness bug; this PR turns the crash into an actionable error.

## Test plan
- [x] `test/test_bloch_waves.py` (11 tests) passes unmodified on top of this change
- [x] Manually verified `check_eigh_memory` raises `MemoryError` when available memory is mocked low, and stays silent for normal-sized matrices with real available memory

### DIFF
--- a/abtem/bloch/dynamical.py
+++ b/abtem/bloch/dynamical.py
@@ -917,7 +917,18 @@ def calculate_structure_matrix(
 
     Mii = xp.asarray(Mii)
 
-    A = A * prefactor * Mii[None] * Mii[:, None]
+    # `A = A * prefactor * Mii[None] * Mii[:, None]` would keep the old A
+    # alive (still bound to the name) through all three multiplications,
+    # needing 2x A's footprint at once for the whole chain -- for a ~13GB
+    # structure matrix that is the difference between fitting on a 24GB GPU
+    # and not. The first multiplication can't be done in place (A may be
+    # read-only here, e.g. from the pandas round-trip in
+    # retrieve_structure_factor_values) but it produces a fresh, writable
+    # array as a side effect, so the old A is freed right after this line
+    # instead of being held through the rest of the chain.
+    A = A * prefactor
+    A *= Mii[None]
+    A *= Mii[:, None]
 
     sg = xp.asarray(excitation_errors(g, energy, use_wave_eq=use_wave_eq))
     diag = 2 * 1 / energy2wavelength(energy) * sg
--- a/abtem/bloch/dynamical.py
+++ b/abtem/bloch/dynamical.py
@@ -933,6 +933,52 @@ def plane_wave_coefficients(hkl: np.ndarray, xp) -> np.ndarray:
     return array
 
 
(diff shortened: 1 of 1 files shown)


TomaSusi:
Status update: **the changes in this PR do not actually fix the problem** and should not be merged as-is.

Real-GPU testing against the production-scale case that motivated this (kyanite, 28,898 beams, on a single 24GB GPU) shows the exact same raw `OutOfMemoryError` before and after every commit here — `check_eigh_memory` never fires:

- Confirmed via `inspect.getsource` (in the same running process, not a separate import) that the correct code — including the `check_eigh_memory(structure_matrix)` call site — is what's actually loaded.
- Confirmed via a debug write directly at the `if required_bytes > available_bytes:` line inside `check_eigh_memory` (writing to a file, to rule out terminal/tqdm output clobbering) that the function body is never reached: no file is ever written.
- Ruled out: multi-GPU distributed dask-cuda workers (only a single GPU is present, so that path never engages), and stale/shadowed `abtem` installs (file paths and source all confirmed correct).
- The `68f299b7` in-place-scaling commit was a real, tested improvement (removes a genuine unnecessary 2x-peak temporary during structure-matrix scaling — see its commit message), but it did not change the observed crash, which suggests either the scaling step wasn't the actual bottleneck for this case, or there's a second, still-unidentified allocation of the same size (an N×N complex128 array at N=28,898 is always exactly 13.36GB, so identical crash byte-counts don't distinguish between candidate causes).

Net: something is preventing `calculate_dynamical_scattering`/`check_eigh_memory` from ever being reached (or executing) for this real workload, and I haven't found the actual mechanism yet despite ruling out several plausible explanations. Picking this back up later rather than merging a guard that doesn't actually guard anything on the case it was written for.

PR #383, opened 2026-09-10
Add the GPU test runner and define GPU validation (tools/gpu)

pzeiger:
Adds `tools/gpu/` — a GPU test runner and a short document defining what "GPU-validated" means for abTEM.

Motivation: GitHub CI has no GPU runner, so GPU-only code paths escape it entirely. Two bugs entered the dev branch that way in September: the Thrust-backed sorts that segfault on some HIP builds (fixed in #376) and the raw-kernel compiler options that NVRTC rejects on CuPy 14 (caught by first-contact CUDA validation of #376, fixed in the follow-up). Both would have been caught by a routine GPU run of the suite — this PR turns that routine into a checked-in, documented procedure instead of tribal knowledge.

Contents:

- `tools/gpu/README.md` — defines GPU validation as two invocations (the stencil reference tests, which force-compile the raw kernels via NVRTC/hipRTC and compare against scipy; and the full `-k gpu` sweep), documents the dependency-group install, and notes that CUDA and ROCm have caught disjoint bugs so neither platform's pass implies the other's.
- `tools/gpu/run_gpu_tests.sh` — wraps both invocations with logging, a `status.tsv` history, and optional failure email. In-place mode tests the developer's current checkout with their active environment and installs nothing; managed mode (`ABTEM_CI_ROOT`) maintains its own clone and venv for unattended cron/scrontab use and self-heals after scratch purges. All site specifics (environment modules, cupy flavor, mail address, an external venv) are env-var knobs; the README includes a Perlmutter `scrontab` example.
- `.gitignore` entry for the in-place mode's log directory.

Validated by running both modes' logic on the two platforms currently doing scheduled runs: a Perlmutter A100 (CuPy 14.2.0, CUDA 12.9, managed mode via scrontab) and a ROCm 7.2 machine (amd-cupy 13.5.1, external-venv mode). Weekly scheduled runs on both have been active since 2026-09-10.

Deliberately out of scope: wiring GPU results into PR checks. The clean long-term solution is a GitHub Actions self-hosted runner on a designated GPU machine — flagging that as an open infrastructure question; this script is the bridge until then and remains useful for cluster runs after.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,3 +17,6 @@ dist
 # uv (no lock file: abTEM is a library, CI resolves fresh)
 uv.lock
 .venv/
+
+# tools/gpu/run_gpu_tests.sh in-place mode logs
+.gpu-test-logs/
--- a/tools/gpu/README.md
+++ b/tools/gpu/README.md
@@ -1,0 +1,61 @@
+# GPU testing
+
+GitHub CI has no GPU runner, so the GPU code paths — CuPy raw kernels and their runtime compilation (NVRTC on CUDA, hipRTC on ROCm), `cupyx` submodules, device-specific numerics — are exercised only on real hardware. Two bugs shipped in 2026-09 exactly this way: a kernel whose compiler options NVRTC rejected (invisible on the ROCm box it was developed on, [#376](https://github.com/abTEM/abTEM/pull/376) follow-up) and Thrust-backed sorts that segfault on some HIP builds. This directory holds the runner that would have caught both.
+
+## What "GPU-validated" means
+
+Two invocations, both green:
+
+```bash
+export OMP_NUM_THREADS=1
+python -m pytest test/test_realspace_multislice.py -q -k "StencilNumericalAccuracy or rejects"
+python -m pytest test/ -q -k gpu -m "not multigpu"
+```
+
+The first force-compiles the raw GPU kernels and compares them against the scipy reference — it fails fast on compiler/toolchain problems. The second runs every GPU-parameterized test (~540 as of 2026-09; the `--runslow`-gated ones stay skipped). Passing on one platform does not imply the other: CUDA/NVRTC and ROCm/hipRTC have caught disjoint bugs, so changes to GPU code should be validated on whichever platforms are available, ideally both.
+
+## Running
+
+`run_gpu_tests.sh` wraps both invocations with logging, a `status.tsv` history, and optional failure email.
+
+**Developer, testing the current checkout** (installs nothing, never touches git state; uses the active environment):
+
+```bash
+tools/gpu/run_gpu_tests.sh
+```
+
+The test dependencies come from the `test` dependency group (`uv pip install -e . --group test` or `pip install --group test .` with pip ≥ 25.1) plus a cupy matching your platform (`cupy-cuda12x`, or a ROCm build).
+
+**Unattended / cron** (`ABTEM_CI_ROOT` enables managed mode: the script maintains its own clone and venv there, hard-tracking `origin/dev`, and self-heals if a scratch purge deletes them):
+
+```bash
+ABTEM_CI_ROOT=$SCRATCH/abtem-gpu-ci \
+ABTEM_CI_MODULES="cudatoolkit/12.9" \
+ABTEM_CI_MAILTO=you@example.org \
+tools/gpu/run_gpu_tests.sh
+```
+
+Example: weekly on NERSC Perlmutter via `scrontab -e` (times are local; `-q shared --gpus 1 -c 32` charges ~¼ node):
+
+```text
+#SCRON -A <account>
+#SCRON -q shared
+#SCRON -C gpu
+#SCRON --gpus 1
+#SCRON -c 32
+#SCRON -t 00:40:00
+#SCRON -J abtem-gpu-tests
+#SCRON --mail-type=FAIL,TIMEOUT
+#SCRON --mail-user=<you@example.org>
+0 6 * * 1 ABTEM_CI_ROOT=$SCRATCH/abtem-gpu-ci ABTEM_CI_MODULES=cudatoolkit/12.9 ABTEM_CI_MAILTO=<you@example.org> $SCRATCH/abtem-gpu-ci/run_gpu_tests.sh
+```
+
+(Copy the script itself somewhere stable first — in managed mode it clones the repo, so it cannot run from inside the clone it manages. On Perlmutter, note that `/global/common` is read-only from compute nodes: keep `ABTEM_CI_ROOT` on `$SCRATCH`.)
+
+**Reusing an existing venv** (e.g. a dev machine where cupy is a custom ROCm build): `ABTEM_CI_VENV=<venv>` activates it and installs nothing; the checkout under test is put on `PYTHONPATH`, so an editable abtem install in that venv is never re-pointed.
+
+All knobs are documented in the header of `run_gpu_tests.sh`.
+
+## Toward CI integration
+
+The eventual proper solution is a GitHub Actions self-hosted runner on a GPU machine, wired into PR checks. Until such a machine is designated, this script is the bridge: scheduled runs on maintainers' hardware (a weekly Perlmutter A100 run and a weekly local ROCm run exist as of 2026-09), plus manual runs before merging GPU-touching changes.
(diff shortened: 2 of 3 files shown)


pzeiger:
@TomaSusi: I am not sure about the placement of the README.md introduced here. Maybe it would be best to add this information into the contributing/development section of the abTEM documentation. Also it might be good to add reading the contributor info to a CLAUDE.md or MODELS.md. In that way our AIs hopefully pick this up independently :).

TomaSusi:
Hmm... I'm not entirely sure why any extra code is needed?

The abTEM test suite already includes GPU tests, and whenever I do development that touches that path, I run them manually on our GPU workstation before I merge such changes in. In fact, my desktop iMac has an AMD GPU, I just never used it since this was not supported until recently. Enabling these for my routine development workflow would already address the biggest gap...

Periodic testing just won't supplement automatic CI integration or manually running the testsuite whenever related changes are made. But of course if you need this for running automated testing on clusters, and this is a useful step towards that direction, as stated, I'm not against facilitating that :) Including multi-GPU would be desirable if possible, currently it looks like that is not the case..?

TomaSusi:
Ah, we should check if the testsuite actually correctly detects and uses AMD, we only developed it for CuPy / CUDA.

pzeiger:
Fully agreed — this is a stop-gap until we have real GPU CI runners, not a substitute for them, and also not a substitute for running the suite manually before merging GPU-touching changes (the README frames exactly that manual run as the norm, the script just wraps it with logging).

The gap it covers is narrower but real: regressions that arrive without any abTEM change, and changes whose GPU impact isn't obvious. Both September bugs were in that class — the Thrust-backed sorts sat broken in dev for weeks because nothing prompted a GPU run, and the #376 kernel compiled fine on the ROCm box it was developed on but never compiled on CuPy 14/NVRTC, so even a conscientious single-platform manual run missed it. A weekly scheduled run is the cheapest detector for "nothing changed here, but the toolchain moved" — that's all it claims to be.

On the iMac: unfortunately that GPU can't help — ROCm is Linux-only and CuPy has no macOS backend at all, so the new AMD support doesn't reach macOS. A Linux box with a supported AMD or NVIDIA card is what it takes.

On AMD detection: confirmed working — this whole effort started from running the suite on a ROCm machine (amd-cupy 13.5.1, gfx1151), where all ~537 GPU-parameterized tests now pass. The suite's GPU gating is just "cupy imports", which the ROCm builds satisfy.

On multi-GPU: agreed, and now added (47d76dd0): `ABTEM_CI_MULTIGPU=1` runs the `multigpu`-marked tests as a third invocation. They skip themselves without ≥2 visible GPUs and dask-cuda, so it's safe anywhere; on a cluster it just means requesting a ≥2-GPU allocation for the scheduled run.

And toward the real solution: if a Linux GPU workstation can be designated, a GitHub Actions self-hosted runner on it, with a workflow that runs the same two invocations on PRs touching GPU paths, is the proper endgame — the script here then remains useful for cluster and periodic runs. Happy to draft that workflow whenever a machine is available.

TomaSusi:
Yep, I just also confirmed that ROCm never supported Apple so I cannot do that on my desktop.

However, I will look into making our Linux/CUDA workstation run the CI on GPU, I have pretty complete access to it so that should be doable :)

Let me know when you’re happy with this PR and I’ll merge it in!

pzeiger:
Marking this ready. Since the review: the multi-GPU opt-in landed (47d76dd0), managed mode now clones anonymously over HTTPS after the first Perlmutter run showed compute nodes can't use login-node SSH keys (f6974513), and the branch-tip script has since run green end-to-end on a Perlmutter A100 (9 stencil + 537 sweep tests) with weekly scheduled runs live on both that and a ROCm box. On the README placement I raised: I'd keep it next to the script for now and port the guidance into the docs' contributing section as a follow-up, since docs live in the separate repo. Great news about the Linux/CUDA workstation — happy to contribute the self-hosted-runner workflow once it's set up, which would make this the cluster-side complement rather than the main line of defense.

pzeiger:
Happy to merge from my side. Since your review: the multi-GPU opt-in (ABTEM_CI_MULTIGPU=1) landed, and testing it on 2×A100 caught two real bugs now fixed — the runner used to report "all green" when the multi-GPU tests silently skipped, so it's now a contract that fails unless they actually pass; and dask-cuda's cuda-bindings dependency needed NVIDIA's package index. Validated end-to-end on a fresh 2-GPU Perlmutter run: 9 stencil + 538 sweep + 6 multi-GPU tests all pass. Also added anonymous-HTTPS cloning after the first Perlmutter run showed compute nodes can't use login-node SSH keys. Scheduled runs are live: nightly single-GPU + multi-GPU on Perlmutter, weekly on a ROCm box.

TomaSusi (commented):
Reviewed `tools/gpu/run_gpu_tests.sh` and the README like a shellcheck-plus-logic pass. No blocking bugs — the exit-code plumbing is careful (`STENCIL_RC=$?`/`SWEEP_RC=$?` captured immediately, `PIPESTATUS[0]` used correctly for the piped multi-GPU run), the `-k "StencilNumericalAccuracy or rejects"` filter is precise (verified "rejects" only matches `test_gpu_stencil_rejects_non_complex_dtype` in that file — the other occurrence is in a docstring, which `-k` never sees), and the managed-mode git/venv logic and mail fallback chain all check out against what the README claims.

Four minor, non-blocking items:

1. No locking between overlapping runs sharing the same `ABTEM_CI_ROOT`. If a scheduled run is still executing (stuck in a SLURM queue, or a hung test) when the next trigger fires, the second run's `git reset --hard` / venv reinstall races the first run's in-flight checkout. Only matters if a run ever actually overlaps itself — `flock` around managed mode would close it if it does.
2. `ABTEM_CI_MULTIGPU`'s managed-mode install (`--extra-index-url https://pypi.nvidia.com dask-cuda cuda-bindings`) is CUDA-only, and neither the script header nor the README says so — worth a one-line note, since the PR itself names a ROCm test machine and a ROCm operator enabling this flag would just hit a confusing install failure. Not a bug this script introduces (dask-cuda/multi-GPU is already NVIDIA-only in abTEM), just an undocumented edge.
3. In-place mode's `REPO="$(cd "$(dirname "${BASH_SOURCE[0]}")/../.." && pwd)"` resolves a symlink's own location rather than its target — if the script is ever symlinked elsewhere for convenience, `REPO` points at the wrong directory. Weak, since the documented invocation is always the direct in-repo path.
4. `status.tsv`'s summary is scraped as the *last* "N passed/failed" match in the whole log, so with `ABTEM_CI_MULTIGPU=1` a passing run's recorded summary reflects only the small multi-GPU tally rather than the ~540-test main sweep — cosmetic, doesn't affect pass/fail correctness, just makes the tsv a bit less informative to skim.

Nice tool — good that this turns "run GPU tests before merging" from tribal knowledge into something checked in and reproducible.

pzeiger:
Thanks — all four addressed in the two commits since (cdc25382, 16c010f1), and I verified end-to-end on a fresh 2-GPU Perlmutter build:

Locking — managed mode now takes a non-blocking flock on ABTEM_CI_ROOT/.run.lock; a run that fires while another still holds the same root skips its trigger instead of racing the shared clone's reset/reinstall. In-place mode shares no state, so it's unlocked.
Multi-GPU is CUDA-only — documented in both the script header and the README (dask-cuda has no ROCm equivalent, so the flag doesn't apply on a ROCm machine).
In-place REPO — now resolves symlinks with readlink -f, so it points at the repo the script physically lives in.
status.tsv summary — now records a per-phase summary rather than the log's last tally, so a multi-GPU run no longer hides the ~540-test main sweep behind the small multi-GPU count.
While validating multi-GPU I also folded its dependency install into the single base resolve — it had been a second uv pip install that could downgrade packages the first pinned. Verified on 2×A100: one clean resolve, then stencil[9 passed] sweep[538 passed, 152 skipped] multigpu[6 passed], all green. Good to merge from my side whenever you're ready — thanks again for the careful read.

TomaSusi:
Checked `cdc25382`/`16c010f1`. All four addressed correctly:

- **Locking**: `exec 200>"${CI_ROOT}/.run.lock"` + non-blocking `flock -n 200`, skipping (exit 0) rather than racing when another run holds it. In-place mode correctly left unlocked since it shares no state.
- **CUDA-only**: now stated explicitly in both the script header and the README.
- **Symlink resolution**: `readlink -f` with a graceful fallback to the original `BASH_SOURCE` path if that's unavailable.
- **Per-phase summary**: each phase now tees to its own tally file and `status.tsv` records `stencil[...] sweep[...] multigpu[...]` instead of scraping the log's last tally line — the main sweep no longer disappears behind the multi-GPU count. Nice touch also fixing `STENCIL_RC`/`SWEEP_RC` to use `PIPESTATUS[0]` now that they're piped through `tee`, consistent with how `MULTI_RC` already worked.

The single-resolve-pass fix for the multi-GPU install (avoiding the two-install downgrade risk) is a good catch too, though it wasn't something I flagged — nice to see it found in your own validation pass.

Nothing further from me. Good to merge.

PR #386, opened 2026-09-13
Ship transition potentials as single graph nodes instead of per-task copies

pzeiger:
Hi Toma!

This is the last installment of the core-loss profiling series (#375, #378): distributed-execution memory. abTEM embeds the built transition potentials into every task that uses them, so a scan split into many tasks ships a copy of the multi-MB array with every task — serialized graph size, scheduler traffic and worker memory all grow with the task count, which is what exhausted workers on production core-loss scans long before the GPUs were busy (and what dask's "large graph" warning was pointing at).

## Root cause and fix

The payload is invisible to dask because it rides inside a `functools.partial`: `MultisliceTransform._from_partitioned_args` bakes `transition_potential` into the transform partial, and dask does not traverse partials. The ensemble machinery already has the right channel for a large constant — a zero-dimensional delayed-backed object array returned from `_partition_args`, which `DummyFrozenPhonons` and `CrystalPotential` use for exactly this — so the fix routes the payload through that rather than inventing a mechanism:

- `shared_constant_arg()` in `abtem.core.ensemble` packages any large constant that way, with one docstring stating the contract that the materialized value is shared by all tasks on a worker and must be treated as read-only.
- `MultisliceTransform` appends one such arg per kwarg named in the module-level `_GRAPH_NODE_KWARGS`, leaves those out of the baked partial, and restores them in `_multislice_transform_member`. Listing a name in that one tuple is the whole opt-in, so the strip and the re-add cannot drift apart, and `_from_partitioned_args` keeps its round-trip contract — the value travels *with* the partition, so `ensemble_blocks`/`generate_blocks` still rebuild a complete transform.
- The PRISM-EELS driver passes its transition potentials as a delayed `map_blocks` kwarg, which dask traverses the same way (verified against dask 2026.8).

Everything is scheduler-managed: one copy in the graph, one materialized copy per worker fetched on demand, freed when the graph's outputs are released, and the graph stays self-contained — it computes on any scheduler, including a cluster started only at compute time (`dask.multi-gpu`), and still computes locally after that cluster is gone. No new API, no config keys, no user-visible change.

## Shared-object safety

With one materialized payload per worker, tasks on that worker (and threads of the local scheduler) share the object — and the drivers used to `grid.match`/`accelerator.match` it in place before privatizing it with `copy_to_device`. Both drivers now match on a private view instead: `BaseTransitionPotential._task_local(match_to=waves)` returns a shallow copy with a private grid and accelerator, while the payload array and everything derived from it stay shared (so nothing is copied). Worth noting this closes a *pre-existing* latent race too: the local threaded scheduler never serialized the embedded partial, so concurrent CPU tasks already shared the object on dev — an energy-ensemble scan could race those in-place matches. A regression test compares threaded against synchronous execution bit for bit.

## Measured effect

hBN, B K edge, 60 kV, 32 mrad, ε = 10 eV, real `SubshellTransitions` potentials (4 transitions, complex64) — the configuration our production core-loss runs use. Measured on Perlmutter (A100 node, idrobolab container), `dev` @ `468e3bee` against this branch, the library swapped by `PYTHONPATH`. Total serialized task payload is what distributed actually ships:

| case | payload | tasks | dev | this branch | ratio |
|---|---|---|---|---|---|
| 1024² scan 32×32, `max_batch=1` | 33.6 MB | ~4230 | 1024 copies, **38.7 GB** | 1 copy, **54.7 MB** | 707× |
| 2048² scan 32×32, `max_batch=1` | 134.2 MB | ~4230 | 1024 copies, **154.6 GB** | 1 copy, **167.9 MB** | 921× |

The copy count is exactly `scan_positions / max_batch`, so the graph grows as payload × task count; on this branch it is always one, and the total is the payload plus a few tens of MB of real task structure. Graph *build* time is unchanged — the cost is paid later, in serialization and transfer.

Running the same scans on the 4-GPU cluster (`dask.multi-gpu`, 47.69 GiB per worker):

| case | dev | this branch |
|---|---|---|
| 1024² wall | 83.40 s | 82.63 s |
| 2048² wall | **183.85 s** | **109.71 s** |
| 1024² peak worker host RSS | 10.6 GB | **0.9 GB** |
| 2048² peak worker host RSS | 41.1 GB (dask: 38.25 GiB) | **1.3 GB** |
| 2048² worker memory events | pauses and resumes twice across dask's 80 % line | no memory warnings at all |
| VRAM pool per worker | 0.3 GB / ~1.2 GB | 0.3 GB / ~1.2 GB |
| checksums | 1.88651708e-03 and 1.88651665e-03 | identical to dev |

The same three cases run on a single GPU with the synchronous scheduler — no cluster, so no serialization at all — take 29.17 / 226.72 / 58.31 s on dev against 29.24 / 227.69 / 58.52 s here, with bit-identical checksums. Strip the transport away and the two refs are the same program, which is what makes the cluster difference below attributable to the copies rather than to the arithmetic.

So the copies cost real time once they are large enough: **1.68× at 2048²**, which is 1024 deserializations of a 151 MB object spread over four workers, plus the stall while a worker sat paused. At 1024² the effect is still invisible, which is why this needed a production-sized case to show up at all.

The memory pressure lands on **worker host RSS**, not on the GPU and not on the scheduler: each task unpickles its own copy and the allocator is slow to release it, so dev reaches 41 GB per worker — four fifths of the 47.69 GiB limit — and pauses, against 1.3 GB here. That memory is *unmanaged*, so it can be neither spilled nor freed; a worker that crosses the line has nothing to bring it back. On a smaller node, or a larger scan, that is an OOM rather than a pause. VRAM is identical between the two (the GPU only ever holds a per-task copy either way), and so are the results — bit-identical checksums at both grid sizes, on real hardware, at production scale.

On a real two-process cluster capped at 2 GB per worker (512², 12×12 scan), peak worker RSS drops **1293 MB → 621 MB** with the same wall time (549 s → 541 s) and an identical checksum. Dev is already at 65 % of that cap with an 8.4 MB payload and 144 tasks, which is the headroom that disappears at production scale.

And the control: single-process end-to-end time is unchanged (6.31 s → 6.22 s, 35.03 s → 34.86 s) with **bit-identical results**. This change is numerically inert and does not make the arithmetic faster — it removes copies.

One honest caveat: distributed's "large graph" warning can still appear, since the graph legitimately carries one copy of the payload. It now describes a bounded constant rather than a per-task multiplication.

## Deliberately not in scope

One interaction a reviewer may reasonably ask about: dask's memory accounting cannot see this payload — `sizeof` of an abTEM object wrapped for the graph is 8 bytes. That is not new, and not specific to this change: `CrystalPotential` already ships its whole `potential_unit` through the identical `_wrap_with_array` + `from_delayed` path, and every abTEM payload in an object array has always been invisible to `sizeof`. Registering `dask.sizeof` for `ArrayObject` would fix it library-wide and change spill behaviour everywhere, so it belongs in its own PR; filed separately.

Two measured follow-ups this change makes possible but does not take (filed separately so they do not enlarge the diff): the per-task `copy_to_device` is now a redundant host-to-device upload per scan chunk on GPU, which a per-worker memo could collapse to one — the shared identity that makes such a memo possible is exactly what this PR introduces; and `dask.delayed(..., pure=True)` content-hashes the payload on every graph build (~0.43 ms/MB, measured), which a cheap `__dask_tokenize__` or an identity memo would avoid. Neither is a regression: both costs exist on dev in another form.

## Testing

New `test/test_transition_potential_transport.py` (synthetic transition potentials, no gpaw, runs on CI): the graph holds exactly one payload-sized key and ~one copy of total task payload; eager, threaded, synchronous and distributed (two-process `LocalCluster`) execution agree bit for bit; the same graph computes locally after the cluster that existed at build time is gone; the PRISM driver agrees lazy vs eager; the task-local view isolates grid/accelerator while sharing the payload; and an end-to-end check that a lazy compute leaves the user's object exactly as built. The distributed test costs ~4 s and is deliberately left unmarked rather than gated behind `--runslow`, since it is the one that guards this PR's central claim.

Because `_partition_args`/`_from_partitioned_args` are shared ensemble machinery, the sweep was wider than the EELS suites: core-loss, ionization, site-filtering, multislice-input, multi-GPU logic, energy-ensemble, scan, detect, PRISM and potential-chunking all pass (326 tests).

One intentional behaviour change worth noting: a scan no longer leaves the caller's transition potential grid-matched as a side effect. That side effect used to be scheduler-dependent — eager and local-scheduler runs mutated the caller's object, distributed runs did not — so code that relied on `tp.build()` working only *after* a scan would now need an explicit `tp.grid.match(...)`. No in-repo caller or test depended on it.

Best,
Paul


TomaSusi (commented):
No correctness or race bugs found — traced the `_task_local()` shallow-copy isolation specifically (the core safety property this PR relies on) and couldn't construct a failing case; `copy_to_device()` always reconstructs a fresh instance before any mutation, so task-private state stays private. Nice work on the measurements too.

Two minor items worth a look before merge (not blocking):

1. `shared_constant_arg()` (`abtem/core/ensemble.py:31`) is essentially the existing pattern from `DummyFrozenPhonons._partition_args` (`abtem/inelastic/phonons.py:211-221`) extracted verbatim, but that call site wasn't refactored to use the new shared helper — duplicate logic left behind. Worth consolidating so a future fix to `shared_constant_arg` doesn't miss `DummyFrozenPhonons`.
2. `test_threaded_and_synchronous_schedulers_agree` only exercises the regular multislice driver, not the PRISM-EELS path (which shares the materialized transition potential via a different mechanism — a bare `dask.delayed(..., pure=True)` through `map_blocks`). Worth the same race coverage there — see #387, where the two drivers already turn out to have drifted apart on a related bug.

TomaSusi:
Checked the five new commits. Both review points are addressed:

- `DummyFrozenPhonons._partition_args` now calls `shared_constant_arg()` directly instead of duplicating it (`abtem/inelastic/phonons.py:214-217`).
- `test_prism_threaded_and_synchronous_schedulers_agree` (`test/test_transition_potential_transport.py:205-248`) now covers the PRISM race path through `_task_local(match_to=...)`. Confirmed it actually bites: reverting `_task_local` to mutate in place locally makes this test (and 3 others) fail.

Spot-checked the other three commits too (defensive raise in `_multislice_transform_member`, the tightened mutation-test fixture, the process-boundary tolerance loosening) — all correct, no new issues.

Looks good to merge from here.

TomaSusi (approved):

pzeiger:
Thanks — both points taken, both fixed.

**1. `shared_constant_arg` left `DummyFrozenPhonons` behind.** Correct, and it was a verbatim copy. `_partition_args` there is now a single call to the helper (`ccac1780`); I checked the equivalence explicitly — same types, shapes and dtypes for `lazy=True` and `lazy=False`, and the round trip through `_from_partitioned_args` is unchanged.

Worth recording which sites I deliberately left alone, since "consolidate the duplicates" could reasonably mean all of them: of the six `dask.delayed(_wrap_with_array)` call sites, only that one is `ndims=0`. The others (`phonons.py:451`, `:516`, `:607`, `:828`, `iam.py:2237`) build `shape=(1,)` args inside chunk loops, so folding them into a helper that returns a zero-dimensional array would change their shapes.

Consolidating did surface something, which I have written up rather than fixed here: `DummyFrozenPhonons` ships its `Atoms` as a shared graph node, and `Potential._prepare_atoms` mutates that object in place (`atoms.wrap(eps=0.0)` plus direct writes to `positions`), with `get_transformed_atoms()` and `DummyFrozenPhonons.randomize()` both returning it uncopied. So a scan rewrites the potential's own atoms, and under the threaded scheduler several tasks wrap one object concurrently. The writes are idempotent so the converged result is right; the exposure is a torn read feeding `SliceIndexedAtoms`. The user's own `Atoms` are safe — `Potential.__init__` copies them — and `FrozenPhonons.randomize` already copies, so this is the no-phonon path only. It predates this PR; happy to send it as its own change, with `FrozenPhonons.randomize` as the oracle.

**2. The race test only covered multislice.** Fixed in `ccac1780`. The PRISM version needs a frozen-phonon ensemble to mean anything: a plain potential gives PRISM exactly one block, so the test would have had no concurrency to expose — it asserts a multi-task graph up front. I verified it bites by reverting `_task_local` to matching in place.

That prompted a second adversarial pass over the whole branch (`26b1b21d`), which found that my *original* multislice race test did not guard what its docstring claimed: the grid/accelerator matching is idempotent, so concurrent tasks converge and a bit-for-bit scheduler comparison cannot see the race — reverting `_task_local` left that test green. The assertions that the caller's object is left unmatched are what actually bite. Both docstrings now say so. Also from that pass: the PRISM transport had no test of its own (deleting the `delayed()` wrapper would have failed nothing — now asserted), the reconstructor arity guard was untested, `_setup`'s `gpts` was silently dropped, and `_task_local`'s docstring now states the `copy_to_device` precondition its safety actually rests on.

**On #387 and driver drift**, since you raised it: the same pass found another instance. An unbuilt `TransitionPotential` from `SubshellTransitions.get_transition_potentials(energy=...)` works through `Probe.transition_potential_scan` and raises `GridUndefinedError` through `SMatrix.transition_potential_scan`, because `_prism_eels_common_setup` builds before matching while `multislice.py` matches before building. It predates this PR — that PR only replaced the two `match` calls — but since it is the same family you raised, I have fixed it here (`c8a3ac6d`): `build()` now runs below the match, matching the order `transition_potential_multislice_and_detect` and `TransitionPotential.scatter` already use, so those two are the oracle. The unbuilt input now agrees with the multislice result to 1.2e-07 relative, and an already-built transition potential is byte-for-byte unaffected.

Other call sites of that mechanism, since it is worth showing the grep rather than asserting it: `multislice.py:1011-1014` and `TransitionPotential.scatter` already match first; `core_loss.py:792` and `:876` (`integrated_intensities`, `show`) build off a grid the caller has already set and have no waves to match to. And a pre-built `TransitionPotentialArray` never entered the moved branch at all — it is not a `TransitionPotential` subclass — so its path is unchanged.

One thing the reorder does *not* fix, which I checked because a crash is usually guarding something: a built transition potential whose grid disagrees with the S-matrix completes **silently**, before and after. That is a separate gap and I have left it alone rather than widen this further, but it is worth knowing the raise was never catching it.

The test uses a recording stub rather than `SubshellTransitions`, so it runs in CI without GPAW; I checked it fails on the unfixed ordering.

Best,
Paul

TomaSusi:
Checked `26b1b21d` and `c8a3ac6d` too.

- `c8a3ac6d`: the match-before-build reorder in `_prism_eels_common_setup` is correct. Confirmed `TransitionPotentialArray` isn't a `TransitionPotential` subclass (`ArrayObject, BaseTransitionPotential`), so it never entered the moved branch, and confirmed `multislice.py` already matches before building — same order now on both drivers. The new `test_prism_matches_the_grid_before_building` (recording stub) is a good, targeted regression test for exactly this.
- `26b1b21d`: the `_setup()` fixture really was dropping `gpts` before (now passed through), and the docstring rewrites on the race tests are accurate — matching-idempotence does mean the scheduler-comparison assertion alone can't see the race, so the caller-object-unmutated assertions are correctly called out as the ones doing the real work.

The DummyFrozenPhonons torn-read note is accurately described too (spot-checked `randomize`/`get_transformed_atoms` — both do return the shared object uncopied on the no-phonon path) — agreed it's fine as a separate follow-up rather than folded in here.

No new issues. Nothing further from me.

PR #387, opened 2026-09-13
Index the potential-ensemble axis in the single-channel core-loss accumulation

pzeiger:
## What

The single-channel branch of `transition_potential_multislice_and_detect` built its accumulation index from the exit-plane axis alone and applied it to axis 0. Measurements are allocated with the potential's ensemble axes **before** the exit-plane axis (`_potential_ensemble_shape_and_metadata`), so as soon as the potential carries an ensemble that index addressed the wrong axis. One defect, two silent symptoms.

**Symptom 1 — eager results scaled by `num_configs`.** With a single exit plane the index was the empty tuple, so `array[()] += contribution` broadcast every configuration's contribution across every configuration slot. Each slot ended up holding the sum over all configurations, and the ensemble mean of that is `num_configs` times the correct answer. The lazy path chunks one configuration per block and was unaffected, so eager and lazy disagreed by exactly the configuration count:

```
num_configs=1: eager=1.046743e-09  lazy=1.046743e-09  ratio=1.0000
num_configs=2: eager=2.094608e-09  lazy=1.047304e-09  ratio=2.0000
num_configs=3: eager=3.142115e-09  lazy=1.047372e-09  ratio=3.0000
```

**Symptom 2 — thickness series silently all zero.** With several exit planes the plane slice landed on the configuration axis. On a length-1 configuration axis — which is what the lazy path produces for *every* configuration count — `slice(k, n_planes)` with `k >= 1` selects nothing, so contributions were dropped in silence. Identical inputs, the only difference being the presence of a length-1 frozen-phonon axis:

```
plain atoms,                  exit_planes=1: per-plane sums ['0.0000e+00', '1.0468e-09', '1.0468e-09']
FrozenPhonons(num_configs=1), exit_planes=1: per-plane sums ['0.0000e+00', '0.0000e+00', '0.0000e+00']
```

Eager and lazy agree here — both return zeros — so an eager/lazy cross-check does not catch it.

## Why this fix

`double_channel=True` was correct throughout: it goes through `_update_loss_measurements`, which builds the full index via `_validate_potential_ensemble_indices`. Routing the single-channel accumulation through the same helper is what fixes both symptoms, and it stops the two branches drifting apart again.

The helper needed no logic change, only widened type hints: it already wraps a non-tuple exit-plane index, so it accepts a `slice` as-is.

Worth noting for reviewers: dividing the eager result by `num_configs` would have corrected the *total* while leaving every per-configuration slot equal to the mean, so the per-configuration data would have stayed destroyed — invisibly. Indexing the axis fixes the data, and the existing `reduce_ensemble` mean is then already right.

## Verification

```
after: num_configs=3: eager=1.047372e-09  lazy=1.047372e-09  ratio=1.0000
after: FrozenPhonons(num_configs=1), exit_planes=1: ['0.0000e+00', '1.0467e-09', '1.0467e-09']
```

matching the plain-atoms series to within the frozen-phonon displacement.

## Tests

New `TestPotentialEnsembleAccumulation` in `test/test_core_loss_fixes.py`, covering eager-vs-lazy for `num_configs > 1`, the total not scaling with the configuration count, a thickness series over frozen phonons staying non-zero and matching the plain-atoms series, and the per-configuration slots being distinct before reduction. No existing test covered a potential ensemble together with exit planes on the core-loss path.

The `double_channel` parametrisation is kept deliberately: it passes before and after, documenting that that branch was always correct.

6 of the 9 new tests fail on `dev` without the source change. Full suite: 1952 passed, 318 skipped.

One assertion detail, in case it is useful elsewhere: core-loss intensities are ~1e-9, far below `np.allclose`'s default `atol` of 1e-8, which will happily call a factor-of-3 error "close". These tests compare against the data's own scale instead.

## Risk

Results change for exactly the configurations that are currently wrong. Anyone who normalised around the `num_configs` scaling will see a change.


TomaSusi (changes requested):
The fix is correct for the regular multislice path (`abtem/multislice.py`) — traced all the axis-ordering combinations (ensemble present/absent × single/multiple exit planes, including multi-axis ensembles) by hand and they check out.

But the PRISM-EELS driver has an independent, unfixed instance of the exact same bug class this PR sets out to fix:

1. `SMatrix._build_ensemble_shape_metadata` (`abtem/prism/s_matrix.py:5056`) builds the exit-plane axis metadata from `self.potential.base_axes_metadata[0]`, which is the per-*slice* `ThicknessAxis` (length = `num_slices`), not the per-exit-plane axis (`potential._get_exit_planes_axes_metadata()` — what the now-fixed `multislice.py:_potential_ensemble_shape_and_metadata` uses). This produces mismatched axis-metadata length whenever a `FrozenPhonons`-based potential is combined with more than one exit plane.

2. `prism_transition_potential_scan`'s `_reduce_and_record` (`abtem/inelastic/core_loss.py:1695`, indexing at `1762-1768` and `1804-1809`) indexes measurements from the exit-plane axis alone and never routes through `_validate_potential_ensemble_indices` the way the fixed `multislice.py` path now does. Since both drivers share `_potential_ensemble_shape_and_metadata`, which allocates the potential's ensemble axes *before* the exit-plane axis, this is exactly the num_configs-scaling / silently-all-zero bug this PR fixes — just left in place on the PRISM side, for both the single-channel branch and the `double_channel` int-index branch.

Since the PR description frames the bug as fixed, and PRISM-EELS is presumably the path production core-loss scans use at scale, could you fold in the equivalent fix here (or file an immediate follow-up) before merge?

Minor, non-blocking: the new tests don't cover multi-axis potential ensembles (frozen phonons + another `CrystalPotential` ensemble axis at once), and `n_plane_axes = 1 if len(potential.exit_planes) > 1 else 0` (`abtem/multislice.py:1180`) re-derives a check `_validate_potential_ensemble_indices` already makes internally.

pzeiger:
Thanks — the main point was right, and worse than it looked from outside. Pushed two commits: `bd3d486a` (the PRISM fix) and `d7f5fb57` (guards and test hardening).

**The two PRISM defects are real, and the first was hiding the second.** `_build_ensemble_shape_metadata` described the exit-plane axis with the per-*slice* `ThicknessAxis` (4 values for a length-5 axis) and raised. Fixing only that turns the loud `RuntimeError` into a silently all-zero thickness series, because `_reduce_and_record` then indexes the plane slice onto the configuration axis. They have to land together, so both are in `bd3d486a`:

```
PRISM thickness series, before   phonons: RuntimeError(ordinal axis 4 vs 5)
                  A1 fixed only  phonons: [0, 0, 0, 0, 0]
                  A1 + A2 fixed  phonons: [0, 4.678e-10, 8.743e-10, 8.743e-10, 8.743e-10]
                  plain-atoms reference: [0, 4.678e-10, 8.743e-10, 8.743e-10, 8.743e-10]
```

Both branches now route through `_validate_potential_ensemble_indices`, as you suggested.

**One half of the diagnosis did not reproduce.** The `num_configs`-scaling symptom does not occur on the PRISM path — the ratio against the multislice reference is constant across configuration counts:

```
single-channel  n=1: 0.4277   n=2: 0.4273   n=3: 0.4273
double-channel  n=1: 0.4299   n=2: 0.4293   n=3: 0.4294
```

because the S-matrix allocates the frozen-phonon axis at size 1 via `_ensemble_mean` and averages over it. Only the silently-all-zero half is live there. (The constant ~0.43 offset between PRISM and multislice is a separate question and unrelated to this PR — flagging it in case it is not expected.)

**Your minor points.** The broadcast count is now derived from the returned index tuple rather than re-deriving `len(exit_planes) > 1`. The multi-axis case is not reachable through the public API — `CrystalPotential(num_frozen_phonons=...)` collapses to a single `FrozenPhononsAxis` — so it is covered by a direct unit test of the index builder instead of end to end.

**A second review pass added two guards**, both in `d7f5fb57`:

- `_validate_potential_ensemble_indices` accepted a `potential_index` with too few entries, silently letting the exit-plane part land on an ensemble axis. That is exactly the mistake this entire family is made of, left armed for the next caller. It now raises.
- `prism_transition_potential_scan` indexes the ensemble axes with zeros because it handles one configuration per call. True for every `SMatrix` entry point, but unenforced, and the function is module-level and named in the public `SMatrix.transition_potential_scan` docstring. Called directly with a multi-configuration potential it returned exactly `1/num_configurations`. Note the direction of travel: before this PR that call produced obviously-broken output (flat series, non-zero at t=0); after it, a plausible monotone series that is silently N times too small. Harder to notice, so it now raises, as `prism_transition_potential_scan_beam_basis` already does.

**Test changes from that pass.** The per-configuration assertion only required the slots to differ, which a permuted index passes; with `ensemble_mean=False` the slots are public, so each is now compared bit-exactly against its own independently built single-configuration run. The PRISM tolerance was `rel=5e-2` against measured per-plane ratios within 1e-4 of unity — a 5 % systematic error passed — now `1e-3`.

**A correction to my own PR description.** "the plane slice selected nothing, giving an all-zero thickness series" is true for the lazy path and for eager with `num_configs == 1`, but not for eager with `num_configs > 1`, where the slice selects real configuration slots and you get a flat series carrying signal at zero thickness. Wrong either way, zeros only in the first case — worth knowing if anyone is checking old results.

**Not fixed here, filed instead.** There is a third instance of this family in the lazy PRISM path: `SMatrix.transition_potential_scan(lazy=True)` with several exit planes builds `chunks`/`new_axis`/`extra_axes_metadata` without the exit-plane axis, so the measurement comes back with one more array dimension than axes metadata (and with `ensemble_mean=False` it fails outright during compute). It is pre-existing on `dev`, it fails loudly, and it needs the `map_blocks` plumbing reworked rather than a metadata one-liner. Happy to do it as a follow-up — or in here if you would rather it not ship separately.

A 56-scenario before/after matrix (multislice × PRISM × elastic, ensemble none/1/3, exit planes 1/5, both channels, eager and lazy) is byte-identical everywhere except where this PR intends to change results.

**Merge order:** this conflicts textually with #391 in `test/test_core_loss_fixes.py` (both append test blocks). Trivial, but #391 is easier last.

TomaSusi:
Checked `bd3d486a`/`d7f5fb57`. Both PRISM defects are fixed, verified against the code rather than just the description:

- `_build_ensemble_shape_metadata` now uses `potential._get_exit_planes_axes_metadata()` instead of the per-slice `ThicknessAxis` (`abtem/prism/s_matrix.py:5058`).
- `_reduce_and_record` now routes through `_validate_potential_ensemble_indices` for both channels (`abtem/inelastic/core_loss.py:1790`).

Both new guards check out too (the short-`potential_index` raise, and the multi-config-potential raise in `prism_transition_potential_scan`), and the num_configs-scaling symptom not reproducing on PRISM makes sense given the S-matrix's `_ensemble_mean` collapse to a size-1 axis — it structurally can't happen there. No new issues found in the fix.

One more thing: I'd appreciate it if you could also fix the `SMatrix.transition_potential_scan(lazy=True)` with several exit planes case you flagged (the missing exit-plane axis in `chunks`/`new_axis`/`extra_axes_metadata` on the lazy PRISM path) rather than leaving it as a separate follow-up. It's the same family of bug this PR is about, and PRISM-EELS is the path production core-loss scans actually use, so I'd rather see it land together if it's not too much extra work. Thanks for the thorough second pass either way.

pzeiger:
Done — the lazy PRISM case is in, as `e753af23`, with a follow-up hardening commit `2ce2f61c`.

**The fix.** `SMatrix.transition_potential_scan`'s lazy branch declared `chunks`/`new_axis` for `map_blocks` from the S-matrix ensemble and the scan only, while each block's measurement also carries an exit-plane axis when the potential has more than one — so the declared block shape disagreed with the computed one. `extra_axes_metadata` came from `potential.ensemble_axes_metadata`, which has no exit-plane entry either. All four now include it.

Verified lazy against eager across 16 combinations — plain atoms and 1/3 frozen-phonon configurations × 1/5/explicit `[1,3]` exit planes × `ensemble_mean` True and False × both channels — all **bit-identical**, matching shapes and matching axes metadata. Six of those raised before. A separate 60-array sweep (1 exit plane, eager and lazy, grid/line/custom scans, elastic `SMatrix.scan`, `SMatrix.build`) is byte-identical to `dev`, so nothing that worked has moved.

**A review pass on that commit found two problems with it, neither in the behaviour**, both fixed in `2ce2f61c`:

- One of the four changes was **inert**. `_extract_measurement` only calls `.item()`, so reshaping the per-block object array does nothing — reverting that hunk, or adding two spurious axes, leaves shapes, metadata and values identical everywhere. I had described it as load-bearing; it wasn't. Removed.
- The new tests compared axes metadata by **type name**, which cannot see wrong axis *values* — and that is precisely the defect `bd3d486a` fixed earlier in this PR, where both sides were `ThicknessAxis` and only the values differed. Mutating the lazy branch back to `base_axes_metadata[0]` (right type, wrong values) left all 14 tests green. They now compare the axes themselves, and that mutant fails 11 of them. I've also added an assertion that the shape dask was told to expect equals what the blocks produce — with a single block a wrong `chunks` declaration is never honoured and goes unnoticed, which is the whole defect class.

**One case in the same area I have deliberately not fixed here.** A scan passed as a bare tuple rather than a `BaseScan` has its position axis squeezed, and the lazy branch declares `chunks` including it regardless:

```
GridScan      eager (5, 2, 2, 64, 1)  lazy declared (5, 2, 2, 64, 1)  computed (5, 2, 2, 64, 1)  to_cpu OK
tuple (x, y)  eager (5, 64, 1)        lazy declared (5, 1, 64, 1)     computed (5, 64, 1)        to_cpu -> RuntimeError
```

It is a *scan-axis* mechanism rather than the exit-plane one, it is pre-existing on `dev` (broken there too, with a different exception — and there `measurement[0]` silently returned a slice of the exit-plane axis while labelling it `PositionsAxis`), and the root cause is a different asymmetry: `Waves.transition_potential_multislice` ends in the module-level `reduce_ensemble`, which honours `_squeeze`, while the PRISM lazy branch only reaches `measurement.reduce_ensemble()`, which handles `_ensemble_mean` and nothing else. That is why the multislice path gets every one of these right.

I'd rather fix that as its own change than fold a fourth mechanism into this PR — three of the fixes in this series have each needed a correction found by the next review, and this one would be going in without the same scrutiny. Say the word if you want it here anyway and I'll do it.

The same sweep turned up two further sites with the identical omission, both pre-existing: `SMatrix.reduce`'s `disable_s_matrix_chunks=True` lazy branch (currently masked — elastic `exit_planes > 1` dies upstream on every path), and `_map_blocks`/`_single_rechunk_reduce`/`_no_chunks_reduce`, which is **reachable today** — `s_matrix.scan(exit_planes=1, lazy=True)` raises the identical "array dimensions does not match axis metadata items". All three are written up together so whoever takes the scan-axis fix can close the family in one go.

TomaSusi:
Checked `e753af23`/`2ce2f61c`. The lazy PRISM fix is correct: `chunks`/`new_axis`/`extra_axes_metadata` now include the exit-plane axis via the same `potential._get_exit_planes_axes_metadata()` helper the eager path uses, so the two can't diverge the way they did before.

The "inert hunk" claim checks out too — `_extract_measurement` (`abtem/prism/s_matrix.py:69-73`) really does only `array.item()[index].array`, so the per-block wrapper's shape was genuinely unobservable; good catch removing it rather than leaving dead complexity in the fix. And the strengthened test (`TestPrismLazyExitPlanes`) does compare `axes_metadata` by value, not type, and asserts dask's declared shape against the computed one — exactly what was missing before.

Appreciate you folding this in rather than filing it separately. The remaining out-of-scope items (tuple-scan squeeze, `SMatrix.reduce`'s masked lazy branch, the reachable `_map_blocks` one) sound like a reasonable separate PR to me — no need to pull those in here.

No new issues. Good to merge from here.

TomaSusi (approved):

PR #388, opened 2026-09-13
Key the scattering-factor cache on the grid and device, not the element alone

pzeiger:
## What

`ScatteringFactorProjectionIntegrals.get_scattering_factor` cached its result under the chemical symbol, while the cached array depends on `gpts`, `sampling` and the device it was allocated on as well. All three are passed in; none was in the key. Reusing one integrator therefore served the first caller's array to every later one:

```
requested gpts=(64, 64)   -> cached array shape (64, 64)
requested gpts=(128, 128) -> cached array shape (64, 64)     cache keys: ['Si']
```

Downstream that surfaces as

```
ValueError: operands could not be broadcast together with shapes (64,64) (128,128)
```

at `integrate_on_grid` (integrals.py:511), one frame from the cache that caused it — because `sinc(gpts, sampling, device)` on the same line *is* computed for the requested grid. Across devices it returns a numpy array for a `"gpu"` request, and the potential build then fails with `TypeError: Unsupported type <class 'numpy.ndarray'>`.

Calling the method directly is worse than either: it returns the stale array **silently**, with the wrong shape for the grid that was asked for.

## Why it matters

This is the default projection path — `projection="infinite"` (iam.py:1241) — so `ScatteringFactorProjectionIntegrals` is what a plain `abtem.Potential(...)` builds. Sharing an integrator is ordinary use: `integrator` is a documented `Potential` parameter and `Potential.integrator` is a public property. A `Potential` that does not pass one constructs a fresh integrator, which is why this has not been hit more often.

## The fix

Key on `(symbol, gpts, sampling, device)`, with the device resolved to a concrete one — `("gpu", <id>)` rather than the `"cpu"`/`"gpu"` string, which does not distinguish one GPU from another. This mirrors the keys merged in #375 (`_local_potential_on_device`) and #378 (`_radial_binning_device_arrays`).

**The cache is now bounded.** A full key admits one entry per (element, grid, device) rather than per element, so a sweep over grids would otherwise accumulate full-grid arrays — the correctness fix would have introduced a slow leak. Eviction is oldest-first. The bound is deliberately local and small; the natural follow-up is a shared bounded helper in `core/backend.py` unifying the several caches of this kind in the tree, and that follow-up should own it.

`scattering_factors` is a public property but has no consumers anywhere in `abtem/`, `test/` or `docs/`, so widening the key shape is safe; its type hint is updated.

## Tests

New `TestScatteringFactorCacheKey` in `test/test_atomic_potential.py`: a second grid is not served the first grid's array; two potentials on different grids may share an integrator and both match a fresh-integrator reference; the array lands on the requested device after the cache was warmed on another; a potential may share an integrator across devices; the cache stays bounded.

4 of the 7 fail without the source change (including both GPU variants). Full suite: 1952 passed, 318 skipped.

## Risk

Low — strictly widens a cache key, no call site changes. GPU behaviour was exercised on a single device here; the multi-GPU aliasing this also guards against needs a multi-GPU box to exercise properly.


TomaSusi (dismissed):
The cache-key widening itself is correct, but the bounded-eviction mechanism this PR adds is not thread-safe — reproduced the crash locally:

1. `abtem/integrals.py:505-507` — `while len(...) >= _MAX_SCATTERING_FACTOR_ENTRIES: self._scattering_factors.pop(next(iter(self._scattering_factors)))` races under concurrent access: `RuntimeError: dictionary changed size during iteration`, and an unguarded `KeyError` from `.pop()` when another thread has already evicted the chosen key. This isn't theoretical — abTEM's CPU path runs under dask's default *threaded* scheduler (GPU forces `synchronous` specifically because CuPy can't tolerate concurrent threads), and this codebase already hit this exact failure mode once: `abtem/measurements.py`'s `_radial_binning_device_arrays_cached` comment explicitly notes a hand-rolled dict cache "raced under the threaded scheduler" and was replaced with `functools.lru_cache` for that reason. This PR reintroduces the same anti-pattern instead of reusing that fix — and the new tests actively encourage the sharing pattern (one integrator across multiple potentials/grids) that triggers it under threads.

2. `_device_cache_key` (`abtem/integrals.py:414-426`) — docstring says it mirrors `_local_potential_on_device`/`_radial_binning_device_arrays`, but those deliberately key off an *already-allocated array's* device rather than trusting ambient CUDA context, specifically because (per their own comments) that context can be wrong when one process drives several GPUs. `_device_cache_key` has no array to anchor on and falls back to exactly the ambient-context read the siblings avoid — could key or serve the wrong GPU's array in that deployment. Not exercised by any test (PR description notes it needs a multi-GPU box to check properly).

Given the precedent already in this codebase, could you swap the hand-rolled dict + manual eviction for `functools.lru_cache` (or the bounded shared helper already flagged as a follow-up in the description) before merge?

Minor, non-blocking: eviction is FIFO-by-insertion rather than LRU (a cache hit doesn't refresh recency, so a sweep over more than 32 grid/element combos gets zero benefit from the cache) — the PR's own comment already flags this as a known interim gap. Also, `test/utils.py`'s `assert_array_matches_device` only checks numpy-vs-cupy, not GPU id, so it wouldn't catch #2 even run on multi-GPU hardware.

pzeiger:
Both points were right. Two commits: `c83943e6` took your suggestion, and `21e34fbc` had to back part of it out — details below, because the reason is worth knowing.

**The race is real.** It does not reproduce under a normal workload (6 trials × 8 threads × 400 iterations, clean), but with `sys.setswitchinterval` shortened it fails on the first trial: two threads pick the same victim from `next(iter(...))` and the second `.pop()` raises `KeyError`. Rare, not absent, and the fix is cheap — so, fixed.

**`functools.lru_cache` turned out to be the wrong tool here, in a way I did not anticipate.** A second review pass found that a per-instance `lru_cache` built around a bound method breaks four things, none of which any test covered:

```
                              lru_cache      OrderedDict
pickle / cloudpickle          FAILS          OK
dask scheduler=processes      FAILS          OK
dask.base.tokenize            random         deterministic
Potential == Potential        False          True
deepcopy                      shares cache   independent
freed without cyclic GC       no             yes
```

The wrapper resolves by `__qualname__`, and one built at runtime around a bound method no longer matches the class attribute of that name, so neither `pickle` nor `cloudpickle` will serialise it:

```
PicklingError: Can't pickle <functools._lru_cache_wrapper object ...>:
it's not the same object as ScatteringFactorProjectionIntegrals._cached_scattering_factor
```

That disables `dask.distributed` for every default `Potential` — which is abTEM's multi-GPU layout, one worker per GPU, i.e. the deployment this cache key exists to make correct. The change would have broken the configuration it was written to serve. `FieldIntegrator` also inherits `EqualityMixin`, which compares `__dict__`, so two fresh integrators stopped comparing equal; and `_lru_cache_wrapper.__deepcopy__` returns `self`, so `Potential.copy()` produced two potentials sharing one mutable cache.

I had also missed that the sibling I cited as precedent, `_local_potential_on_device`, carries a `__getstate__` for exactly this reason, with `test_local_potential_device_cache_survives_use_but_not_pickle` guarding it. Copying its key shape without its pickle discipline is what went wrong.

`21e34fbc` uses a bounded `OrderedDict` instead. Eviction is written so it cannot race — `popitem()` and `move_to_end()` are single C-level dict operations and the `KeyError` a losing thread sees is caught rather than escaping — which is the part the first attempt got wrong. It is LRU rather than FIFO, so your second point is fixed too, and the public `scattering_factors` property comes back with it, so the PR no longer removes API. If you would still prefer `lru_cache`, the module-level form (as in `measurements.py`) avoids all four problems; I went per-instance because these are full-grid arrays and holding 32 of them for the life of the process is not a rounding error at 1024².

**Your second point, on the device key.** You were right that the docstring overclaimed: the siblings anchor on an already-allocated array and I was reading ambient CUDA context. There *is* an array in scope — `integrate_on_grid` allocates it immediately before the call — so it is now passed as `like=`. And anchoring only the *key* was half a fix: the key named one device while the allocation still followed the ambient context, so under multi-GPU an array could be cached under a device it does not live on. The allocation now happens inside the keyed device's context, as `_local_potential_on_device` does with `with like.device:`.

Two further changes from the review pass:

- `device` is no longer part of the cache key. `get_array_module` accepts an array or a module as well as a string, and an array is unhashable, so keying on it raised `TypeError` for a call that worked on `dev` — a regression I had introduced. `"gpu"` and the `cupy` module also took two entries for one physical device. All spellings now share one canonical entry.
- The LRU test could not discriminate: it inserted exactly `maxsize` keys, so nothing was ever evicted and LRU was indistinguishable from FIFO. It now inserts past the bound.

Tests now cover pickling, cloudpickling, the `processes` scheduler, deterministic tokenization, equality, deepcopy independence, non-cyclic release and device spellings; 12 of them fail against `c83943e6`.

The cache-key widening itself is unchanged throughout and verified bit-identical against a freshly constructed integrator on every grid tried.

**Merge order:** this no longer conflicts with #391 — that PR has dropped its `integrals.py` changes — so it is independent of the other three.

TomaSusi:
Checked `c83943e6`/`21e34fbc`. Agreed that `lru_cache` was the wrong call here — reproduced the pickling, deepcopy and equality breakage you described locally; matches CPython's `functools`/`copy` internals exactly.

The `OrderedDict` + `popitem(last=False)`/`move_to_end()` approach in the final commit is race-free (single C-level dict ops, no TOCTOU window), fixes the device-key anchoring (now via `like=` at the actual allocation site, with the allocation itself done inside `with cp.cuda.Device(...):`, which closes the second gap — key naming one device while the array was allocated on the ambient one), and the LRU-vs-FIFO test now actually discriminates (inserts past the bound instead of exactly at it).

No new issues found. Looks good to merge.

TomaSusi (approved):

PR #389, opened 2026-09-13
Make wrapped_crop_2d total for any crop corner

pzeiger:
## What

`wrapped_slices` assumed `-n <= start < n` and `0 < stop <= 2n` and split on the raw values. Outside that range it returned a wrongly shaped array rather than failing: for a window lying entirely in negative indices, `slice(0, stop)` with `stop <= 0` is Python's *negative indexing* and selects `n + stop` elements instead of none, so the two pieces concatenated to `n + size` elements.

At `n=192, size=64`, **479 of 800 consecutive corners came back with the wrong shape**:

```
contiguous bad ranges: [(-400, -257), (-192, -65), (193, 399)]
  corner -80 -> shape (1, 256, 64)   wrapped_slices(-80, -16, 192) = (slice(112, None), slice(0, -16))
```

Nothing checked the result, so the mis-shaped crop reached the caller intact and surfaced as

```
ValueError: operands could not be broadcast together with shapes (4,1,64,64) (1,51,256,64)
```

in `prism_transition_potential_scan`, three frames from the cause. Any model with atoms far enough outside the cell reaches it once the interpolation factor makes the crop window small enough — found on a 190-atom hBN model with x down to −3.64 Å.

## The fix

Normalise before splitting — `start %= n; stop = start + size` — which leaves only the two cases two slices can express: the window fits, or it wraps once. Verified against modulo indexing for **every** corner across several periods.

**Windows wider than one period.** Two slices cannot cover more than one period, so `wrapped_slices` now says so explicitly and `wrapped_crop_2d` takes a general path that gathers by modulo index. This replaces the `xp.pad(..., mode="wrap")` fallback, which was not merely opaque but **wrong**: `n=16, corner=-23, size=24` came back `8×8`, and windows spanning several periods raised a bare `ValueError`. The gather copies through fancy indexing, so it is kept off the common path — but it is correct where the slice pair cannot be, and it makes the function total for any corner and any size.

**Shape check on the way out.** Across all 800 corners there was no case of a correct shape carrying wrong values, so a shape comparison is a *complete* detector for this class of defect — and it costs nothing on the hot path. It turns a broadcast error three frames downstream into a failure at the crop.

## Tests

New `TestWrappedCrop2D` in `test/test_prism.py`: every corner across several periods gives the right shape and values for three `(n, size)` combinations; both axes wrap independently; windows wider than one period are correct across a swept corner; `wrapped_slices` refuses more than one period; CPU and GPU agree with the modulo-index reference.

All 11 fail on `dev` without the source change. Full suite: 1952 passed, 318 skipped.

A note on picking test parameters here, since it caught me out: the old code was correct in a narrow band around zero, so a single hand-picked corner can pass against it by luck. `corner=(5, -7)` does. The multi-period test sweeps the corner instead — 287 of the swept cases fail on `dev`.

## Interaction with open PRs

#338 adds four new `wrapped_crop_2d` call sites and #365 adds two; neither touches `wrapped_slices`. Both inherit the fix, and the new shape check will fail loudly at the crop rather than at a broadcast downstream.

### DIFF
--- a/abtem/prism/utils.py
+++ b/abtem/prism/utils.py
@@ -60,24 +60,72 @@ def minimum_crop(positions: np.ndarray, shape):
 
 
 def wrapped_slices(start: int, stop: int, n: int) -> Tuple[slice, slice]:
-    if start < 0:
-        if stop > n:
-            raise RuntimeError(f"start = {start} stop = {stop}, n = {n}")
+    """Two slices covering ``[start, stop)`` on a periodic axis of length ``n``.
 
-        a = slice(start % n, None)
-        b = slice(0, stop)
+    ``start`` may be any integer. Normalising it into ``[0, n)`` first leaves
+    only two cases -- the window fits, or it wraps once -- which is what two
+    slices can express.
 
-    elif stop > n:
-        if start < 0:
-            raise RuntimeError(f"start = {start} stop = {stop}, n = {n}")
+    The previous implementation assumed ``-n <= start < n`` and ``0 < stop <= 2n``
+    and split on the raw values. For a window lying entirely in negative indices
+    (``stop <= 0``) it produced ``slice(0, stop)``, which is Python's *negative
+    indexing* and selects ``n + stop`` elements rather than none, so the two
+    pieces concatenated to ``n + size`` elements instead of ``size``. The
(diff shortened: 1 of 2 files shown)


TomaSusi:
Nice, thanks for catching this.

PR #390, opened 2026-09-14
Wrap atoms into the cell before binning them into slices

pzeiger:
## What

`SliceIndexedAtoms` bins atoms with `np.digitize` against bin edges nudged down by 1e-12. `digitize` returns `0` for any `z` below the first edge — **including arbitrarily negative `z`** — and `len(bin_edges)` for any `z` above the last, which `label_to_index` then discards. So an atom outside the cell was either assigned to slice 0 whatever its true wrapped depth, or dropped. Both in silence.

`Potential._prepare_atoms` wraps before slicing, so the potential's own atoms were fine. The other two construction sites are not:

**1. Explicit core-loss `sites=`.** `_extract_scattering_sites` passes an `Atoms` straight through, so `sites=<Atoms>` and `sites=None` disagreed about which atoms ionise and where:

```
z                 [1.0, 3.0, 5.0, 7.0, -0.5, 8.5]
correct per slice [2, 1, 1, 2]  total 6
sites=None        [2, 1, 1, 2]  total 6
sites=<Atoms>     [2, 1, 1, 1]  total 5
```

The `z = -0.5` atom belongs in slice 3 and is put in slice 0; the `z = 8.5` atom belongs in slice 0 and disappears.

**2. `CrystalPotential.get_sliced_atoms`.** It tiles the unit's transformed atoms and builds `SliceIndexedAtoms` from them without wrapping, so `sites=None` loses a site there too — 7 of 8 for the case in the tests. Worth calling out because "pass `sites=None` and let the potential supply them" is the natural workaround for symptom 1, and it does not hold here.

A dropped site loses signal, which shows up as a scale error. A mis-sliced site ionises at the wrong **depth**, which corrupts exactly what a depth-resolved or atom-counting measurement is trying to extract, while leaving the total roughly intact. Routine structures reach this: ASE builder output, rotated or re-celled models, imported CIFs.

## The fix

Factor the wrap-and-snap out of `_prepare_atoms` into `wrap_and_snap_atoms` and apply it in `SliceIndexedAtoms.__init__` — the single point all three construction sites pass through, so they cannot drift apart again.

It is **idempotent**: atoms already inside the cell are returned unchanged, so the potential path is unaffected and pays no copy. The caller's `Atoms` is never modified.

`SlicedAtoms` — the class used when the integrator is finite — is deliberately left alone: it legitimately receives `pad_atoms` output from beyond the cell. That distinction is what makes `SliceIndexedAtoms.__init__` a safe home for the wrap. (`_prepare_atoms` builds `SlicedAtoms` when `integrator.finite` and `SliceIndexedAtoms` otherwise, and in the latter branch `margins` is 0.0, so `SliceIndexedAtoms` never receives non-zero-margin padding atoms.)

Finally, raise when a label falls outside `[0, num_slices - 1]` rather than letting `label_to_index` discard it. After the wrap there is no such atom, so this is the guard that would have made the original defect loud.

## Tests

New `TestSliceIndexedAtomsWrapping` in `test/test_potentials.py`: out-of-cell atoms land in their wrapped slice; explicit `sites=` agrees with the potential's own atoms; the caller's `Atoms` is not modified; wrapping is idempotent; `CrystalPotential` keeps every site.

3 of the 5 fail on `dev` without the source change.

## Note for #365

That branch calls `_extract_scattering_sites` from four new entry points and inherits this fix for free.


pzeiger:
A second review pass on this branch found the first commit fixed the bug only for `pbc=True`, and made things worse for everything else. Pushed `d52ec745`.

`ase.Atoms.wrap` is a **no-op along any axis whose `pbc` is False** — which is ASE's default for a hand-built `Atoms`, and what every `build.*(vacuum=...)` slab carries. So the most common abTEM input shape was untouched. Worse, with `wrap()` doing nothing the boundary snap below it no longer saw near-boundary values but arbitrarily out-of-cell ones, and set them all to zero:

```
                          sites=<Atoms>     truth
pbc=True                  [2, 1, 1, 2]      [2, 1, 1, 2]
pbc=(True, True, False)   [3, 1, 1, 1]      [2, 1, 1, 2]
pbc=False                 [3, 1, 1, 1]      [2, 1, 1, 2]
```

— further from the truth than the unfixed `[2, 1, 1, 1]`. In plane it was a regression against `dev`: an explicitly passed site at x = 4.3 in a 4 Å cell came back at 0.0, where `dev` left it at 4.3. Those are not equivalent — 4.3 and 0.3 are (`fft_shift_kernel` and `fast_roll` are exactly periodic, agreeing to float32 noise), 0.0 is a different site 0.3 Å away.

`d52ec745` wraps with `np.mod` in all three directions instead, regardless of what the `Atoms` object declares. That is the right semantics here: `SliceIndexedAtoms` feeds an infinite-projection potential that is periodic in all three directions, and the z bin index is taken modulo the cell either way. The snap stays — positions land in `[0, L)` and the bin edges are nudged down by 1e-12, so a value in `(L - 1e-12, L)` would still fall outside every bin — but it now only ever sees values already inside the cell, which is the case it was written for.

**This also fixes `sites=None` for non-periodic atoms, which `dev` gets wrong too.** `_prepare_atoms` calls the same wrap, so the potential's own atoms were mis-binned whenever `pbc[2]` was False. The original issue was written up from an explicit-`sites` reproduction, so that half went unnoticed:

```
                          sites=None (before)   sites=None (after)
pbc=(True, True, False)   [2, 1, 1, 1]          [2, 1, 1, 2]
pbc=False                 [2, 1, 1, 1]          [2, 1, 1, 2]
```

Ordinary structures are unaffected — Si, MoS2, fcc111, a non-periodic slab and hcp Mg all build bit-identically to `dev`.

Other changes from the same pass: the test class is parametrised over `pbc` True / `(True, True, False)` / False (the gap that let this through); the idempotence check is bitwise rather than `np.allclose`; the `CrystalPotential` test compares the per-slice *distribution* against the tiled unit rather than a head count, which passed even with atoms in the wrong slices; the out-of-range guard drops its unreachable negative branch and now names the offending position; and a degenerate cell returns early so the caller's own validation reports it rather than a modulo by zero.

One incidental improvement worth knowing: this removes a pre-existing order dependence in `CrystalPotential.get_sliced_atoms`, whose result on `dev` depended on whether the unit potential had been built first.

TomaSusi (changes requested):
Good catch on the original bug, but the fix as written introduces a real regression of its own:

**`SliceIndexedAtoms.__init__` (`abtem/slicing.py:678`) wraps unconditionally, ignoring `Potential(periodic=False)`.**

`_prepare_atoms` (`abtem/potentials/iam.py:836-880`) deliberately only calls `wrap_and_snap_atoms` inside the `if self.periodic:` branch — when `periodic=False`, atoms are randomized *after* padding and are never wrapped, by design (per `Potential`'s own docstring: the transformed potential is "cut out of a larger repeated potential, which may not preserve periodicity"). `SliceIndexedAtoms` is used for both branches (it's selected by `integrator.finite`, not by `self.periodic`), so with the default infinite-projection integrator and `periodic=False`, atoms now get force-wrapped inside `SliceIndexedAtoms.__init__` regardless.

Concrete case: a 4-slice, 4 Å cell, `periodic=False`, an edge atom thermally displaced to `z=-0.15` (a normal frozen-phonon outcome for an atom near a face in a non-periodic/finite sample). On `dev` it stays at `z=-0.15` and lands in slice 0 — the physically sensible place for a non-periodic sample. On this branch, `wrap_and_snap_atoms` mod-wraps it to `z=3.85`, and it lands in slice 3, the *opposite* end of the box — the same class of depth corruption this PR is fixing, just reintroduced for the non-periodic path. None of the 13 new tests exercise `periodic=False`, so this doesn't get caught.

I think the wrap in `SliceIndexedAtoms.__init__` needs to be conditional (or the potential path needs to pass already-correctly-handled atoms and `SliceIndexedAtoms` needs a way to opt out), so that non-periodic potentials keep their current, deliberate no-wrap semantics.

Two smaller, non-blocking notes:
- `abtem/atoms.py` already has `wrap_with_tolerance` (used by `merge_close_atoms`/`shrink_cell`/`orthogonalize_cell`), which calls plain `atoms.wrap()` — a no-op on any axis where `pbc=False`, i.e. exactly the failure mode this PR's `wrap_and_snap_atoms` was written to avoid. Might be worth fixing that helper too (or consolidating), since a future caller of `wrap_with_tolerance` could hit the same silent-drop bug this PR just fixed elsewhere.
- `wrap_and_snap_atoms` runs before `BaseSlicedAtoms.__init__`'s orthogonal-cell check and reads `np.diag(atoms.cell)`, silently ignoring shear terms for a non-orthogonal cell before the `RuntimeError` for non-orthogonality fires one line later. Harmless today since the exception still raises before any wrapped-but-wrong positions get used, just a bit fragile — reordering would be cleaner.

pzeiger:
You're right, and thank you — that's a real regression I introduced. Fixed in `42cfb2b1`.

**Confirmed.** One correction to the reproduction, which doesn't change the conclusion: an atom *placed* at `z=-0.15` never reaches `SliceIndexedAtoms` — `pad_atoms(margins=0.0, directions="z")` drops it first, on both branches. The route that does reach it is the one you describe physically: the frozen-phonon randomize runs *after* padding on the `periodic=False` path, so an atom just inside the entrance face gets displaced out. With `sigmas=0.25, seed=1` on a 4 Å / 4-slice cell:

```
into SliceIndexedAtoms   z=[-0.106, 0.094, 2.138, 4.36]
dev                      per-slice [2, 0, 1, 0]   z unchanged
previous commit          per-slice [2, 0, 1, 1]   z=[0.094, 0.36, 2.138, 3.894]
```

An atom 0.1 Å below the entrance face moved to the exit face. Exactly the depth corruption this PR exists to prevent, for the other path.

**The fix.** `SliceIndexedAtoms` takes a `wrap` argument, defaulting to `True` since that is right for everything else that reaches it. `_prepare_atoms` passes `self.periodic`, and `_extract_scattering_sites` passes the potential's own value so that explicit `sites=` are sliced exactly as `sites=None` would be — which is the invariant this PR is about in the first place.

The out-of-range `RuntimeError` I added had to become conditional on the same flag: without a wrap, an atom outside the cell is *expected*, and `label_to_index` dropping it is pre-existing behaviour I must not change. (That silent drop on the non-periodic path is arguably its own defect — an atom just past the exit face vanishes rather than contributing to the last slice — but it isn't this PR's, and I've left it alone.)

All four `periodic=False` cases now match `dev` exactly, and the periodic fix is unchanged.

**Your second note is applied:** `super().__init__` now runs before the wrap, so a non-orthogonal cell raises `RuntimeError: atoms must have an orthogonal cell` rather than being silently mod-wrapped by `np.diag` first, and a zero-height cell again raises `ZeroDivisionError` as on `dev`.

**Your first note I have filed rather than fixed.** `wrap_with_tolerance` does call plain `atoms.wrap()`, and is therefore a no-op on non-periodic axes — the same failure mode. But its callers are `merge_close_atoms`, `shrink_cell` and `orthogonalize_cell`, so changing it could shift cell-construction results well outside this PR's blast radius, and I'd rather not fold that into a fix for a different path. Tracked, with your observation, for a PR of its own. Happy to do it next if you'd like it sooner.

**Test gap closed.** The class is now parametrised over `periodic` True/False — the gap that let this through — plus a test that the guard does not fire when wrapping is off, and one for the validation ordering. Two of the new tests fail against the previous commit.

TomaSusi:
Checked the fix. Confirmed:

- `SliceIndexedAtoms.__init__` now takes `wrap: bool = True`, and `super().__init__()` (the orthogonality check) runs before the wrap (`abtem/slicing.py:657-661`, `687-688`).
- `_prepare_atoms` passes `wrap=self.periodic` (`abtem/potentials/iam.py:897-902`); `_extract_scattering_sites` reads the same `.periodic` property rather than a separate proxy (`abtem/inelastic/core_loss.py:1304-1309`), so the two paths can't drift apart again.
- The out-of-range guard is correctly gated on `wrap`, and I checked the epsilon margins can't collide (wrap snaps at `1e-10`, bin edges nudge at `1e-12`, so a wrapped atom can never land past the last valid bin — the guard can't spuriously fire).
- Reproduced the exact `periodic=False`, `sigmas=0.25, seed=1` scenario from your comment on both branches — bit-identical positions, `z=-0.106` no longer gets wrapped to `z=3.894`. All 17 new tests pass.
- Also checked `CrystalPotential.get_sliced_atoms`, which doesn't pass `wrap=` explicitly — confirmed that's fine, since `CrystalPotential` hardcodes `periodic=True` and never goes through the pad-then-randomize path that caused the original regression.

No new issues. Good to merge from here.

TomaSusi (approved):

pzeiger:
One correction to my previous comment, on the `wrap_with_tolerance` point — I agreed with it without checking, and it does not hold.

I wrote that `wrap_with_tolerance` "does call plain `atoms.wrap()`, and is therefore a no-op on non-periodic axes — the same failure mode". The first half is true and the conclusion does not follow. The `atoms.wrap()` call on `abtem/atoms.py:520` is indeed a no-op when `pbc=False`, but the four lines after it do the wrapping unconditionally, through the scaled positions rather than through ASE:

```python
atoms.wrap()
d = np.linalg.norm(np.array(atoms.cell), axis=0)
tol = tol / d
scaled_positions = atoms.get_scaled_positions()
scaled_positions = ((tol + scaled_positions) % 1) - tol
atoms.positions[:] = atoms.cell.cartesian_positions(scaled_positions)
```

`get_scaled_positions()` leaves non-periodic axes unwrapped, but the `% 1` that follows applies to all three components regardless of `pbc`. So the function wraps on every axis either way:

```
                          z in                  ->  z out
wrap_with_tolerance, pbc=False  [ 5.0  -1.5  2.0 ]  ->  [1.0  2.5  2.0]
wrap_with_tolerance, pbc=True   [ 5.0  -1.5  2.0 ]  ->  [1.0  2.5  2.0]
plain ase Atoms.wrap(), pbc=False                    ->  [5.0 -1.5  2.0]   <- the no-op
plain ase Atoms.wrap(), pbc=True                     ->  [1.0  2.5  2.0]
```

Identical output at `pbc=False` and `pbc=True`, on `dev` @ `ce7d11f6`.

So there is no latent defect in `merge_close_atoms` / `shrink_cell` / `orthogonalize_cell` of the kind this PR fixes, and nothing for me to file — I had said I would track it, and what I should have done first was run it. Whether wrapping *unconditionally* on a non-periodic axis is the right behaviour for those three cell-construction helpers is a separate question and not one this PR raises; they are geometry utilities where wrapping looks intended.

The rest of the review stands unchanged.

PR #391, opened 2026-09-14
Remove dead code in the core-loss and integrals modules

pzeiger:
## What

Three unrelated pieces of dead code, one theme. Each is a separate commit and can be dropped independently if you disagree with one.

### 1. `_validate_transition_potentials` (core_loss.py) — delete

No callers anywhere in the package; its own definition was the only occurrence tree-wide. The wrapping it performed is done independently by an `isinstance` check in `Waves.transition_potential_multislice`.

The two are **not** equivalent spellings — the dead one was duck-typed on `hasattr(..., "scatter")` and they diverge on generators — so it is worth recording that it should not later be "restored" as a synonym.

### 2. `detectors_elastic` (multislice.py) — raise instead of ignoring

Declared in `transition_potential_multislice_and_detect`'s signature and never read. Auditing every parameter of that function for body references, it is the only one with zero:

```
waves 25 | potential 16 | transition_potential 12 | detectors 10 | algorithm 9
sites 5 | threshold 3 | double_channel 2 | scatter_max_batch 1 | pbar 1
detectors_elastic 0
```

so elastic detectors passed through `Probe.transition_potential_scan` were accepted in silence and only the inelastic measurement came back — a caller expecting a simultaneous elastic image got a short list and no explanation.

What makes this worse than an ordinary unused parameter: these kwargs arrive through `**multislice_func_kwargs`, so **every other keyword a caller might misspell already raises `TypeError`**. This one spelling suppressed the error the caller would otherwise have got.

It now raises `NotImplementedError` pointing at the separate-`Probe.scan()` workaround. Implementing it — detecting on the elastic waves at each exit plane and returning a second group of measurements — is a feature, and deliberately not attempted here. Happy to do that instead if you would rather have the capability than the guard.

### 3. `GaussianProjectionIntegrals` (integrals.py) — delete

Unreachable. `Potential.__init__` builds only `QuadratureProjectionIntegrals` (`projection="finite"`) or `ScatteringFactorProjectionIntegrals` (`projection="infinite"`) and raises `NotImplementedError` otherwise. The only other references in the tree were five commented-out lines in `test_atomic_potential.py`. Not exported at top level, absent from the docs.

Removing it also removes its `_sinc_cache`, which keyed device-resident arrays on the literal `"cpu"`/`"gpu"` string — one of the caches a follow-up on per-device caching would otherwise have had to fix or migrate.

Deleting the class orphaned `correction_projected_scattering_factors`, `gaussian_projected_scattering_factors` and `gaussian_projection_weights`, which had no other callers, and with them the `polar_spatial_frequencies` and scipy `erf` imports. Removed too — leaving code that this PR itself made dead would defeat the point.

## Tests

`detectors_elastic` now raises rather than being ignored, and the dead validator is gone. Both asserted in `test/test_core_loss_fixes.py`.

## Risk

`detectors_elastic` is the only behaviour change: code that passed it and silently got one measurement back now gets an exception. That code was not getting what it asked for.


pzeiger:
A review pass found the central justification in this PR's description is wrong, so I have narrowed it. Pushed `c90c7782`.

**`GaussianProjectionIntegrals` is reachable public API — please disregard the "unreachable" claim.** I argued that `Potential.__init__` only ever builds the two live integrators and raises `NotImplementedError` otherwise. That raise is inside `if integrator is None:`. `Potential` takes `integrator: FieldIntegrator | None = None`, documented as *"Provide a custom integrator for the projection integrals of the potential slicing"*, so passing an instance bypasses the string dispatch entirely.

It is not merely constructible — it builds a physically sensible potential, matching `projection="infinite"` to six significant figures in the total and sitting between the two live schemes in local deviation, exactly as a third slicing scheme should:

```
sums          infinite=3.105263e+05  finite=3.089394e+05  gaussian=3.105261e+05
max rel dev   gaussian vs finite   = 0.2632
              gaussian vs infinite = 0.5021
              finite   vs infinite = 0.6038
```

It also pickles and survives the dask path, and `benchmarks/benchmark_finite_projection.py` already uses the `Potential(integrator=<instance>)` pattern — so that is a live route, not a theoretical one.

**So I have restored it**, along with the three helpers, the two imports and the commented-out test blocks. Deleting it may well still be right: it is unfinished work (`afd2d110 begin reimplementing gaussian scattering factors`), it has no caller anywhere in the repo's history, and it carries a latent dead cache — `get_gaussians` computes a key, checks `self._gaussians` and never writes to it. But that is an API-removal decision for you and Jacob, not something to slip through a PR titled "remove dead code" on a false premise. **Happy to do it as an explicit deprecation or removal if you want it gone — just say so and I will redo it with a changelog entry.**

What remains in this PR is genuinely dead, and I re-verified both:

- `_validate_transition_potentials` — one occurrence tree-wide, its own `def`.
- `detectors_elastic` — declared in the signature, never read. Confirmed numerically as well as by inspection: on `dev`, passing a real `AnnularDetector` produced a result bit-identical to omitting it.

**Two fixes to the `detectors_elastic` guard**, both from the same pass:

- It fired on `detectors_elastic=[]`, which requests *no* elastic detectors and so was never having anything silently dropped. That turned a working call into a crash for no benefit; it now tests truthiness.
- It sat inside the per-chunk worker, and abTEM is lazy by default — so the caller still got a fully built measurement object back with no complaint and only found out at compute time, from inside a dask traceback. That defeats the point of the change. The check now runs in `Waves.transition_potential_multislice`, where the kwargs are first assembled, and is kept in the driver too since that is also a public entry point.

Tests are parametrised over `lazy` and cover the omitted / `None` / empty-list forms, none of which may raise. The validator test no longer just asserts that the deletion happened — it asserts the behaviour that made the dead function non-equivalent to the live `isinstance` check (they diverge on a generator).

The PR still changes no numbers: potential builds on CPU and GPU, for both live integrators, hash identically to `dev`.

**Merge order:** with the `integrals.py` changes gone this no longer conflicts with #388. It still conflicts textually with **#387** in `test/test_core_loss_fixes.py` (both append test blocks) — trivial, but this one is easier last.

**Unrelated, found while reviewing:** `projection="finite"` is not bit-reproducible on GPU — four consecutive runs on unmodified `dev` give four different hashes, consistent with float32 atomic-add ordering in the CUDA radial interpolation. Worth its own issue if that is news.

TomaSusi (commented):
Looks like this already went through a self-correction pass before I got to it — `c90c7782` reverted the `GaussianProjectionIntegrals` deletion (it's actually reachable via the public, documented `Potential(integrator=...)` parameter, used by `benchmarks/benchmark_finite_projection.py`) and fixed the `detectors_elastic` guard, which originally used `is not None` (would have raised on the legitimate no-op `detectors_elastic=[]`) and only checked eagerly inside the per-chunk worker (so under `lazy=True` the error would only surface at `.compute()` time, from inside a dask traceback, rather than at call time).

Reviewed the branch as it stands now (`c90c7782`): the `_validate_transition_potentials` deletion is confirmed safe (zero callers anywhere in the tree, including docs/examples), and the current `detectors_elastic` guard — truthy check, hoisted into `Waves.transition_potential_multislice` so it's eager regardless of `lazy=`, with the driver-level check kept as defense-in-depth since `transition_potential_multislice_and_detect` is also called directly elsewhere (e.g. `test/test_diffraction_pattern_cache.py`) — is correct. Full suite passes.

One trivial, non-blocking nit: `abtem/waves.py:1571` imports `_DETECTORS_ELASTIC_MESSAGE` from `abtem.multislice` inside the method body; `abtem/waves.py` already imports other names from `abtem.multislice` at module level and there's no circular-import barrier, so it could just move to the top-of-file import block.

Nothing blocking — good to merge.

pzeiger:
Thanks — and yes, that self-correction came out of a review pass on this branch, which caught the `integrator=` reachability before you did. Good to have it independently confirmed.

The nit is fixed in `0286dcaf`: `_DETECTORS_ELASTIC_MESSAGE` now comes from the module-level `from abtem.multislice import (...)` block in `abtem/waves.py`. Verified no circular import.

One thing worth putting on the record for whoever picks up `GaussianProjectionIntegrals` later: it is reachable and it *works*, but it has a latent defect of its own. `get_gaussians` and `get_corrections` each compute a cache key, check the dict, miss, recompute — and never write back, so both caches stay empty for the life of the object:

```
_gaussians   after a full potential build: {}
_corrections after a full potential build: {}
```

Its `_sinc_cache` also keys device-resident arrays on the literal `"cpu"`/`"gpu"` string, which is the same defect we have open against `QuadratureProjectionIntegrals._device_arrays` — and I'd previously written that one off as unreachable dead code on the strength of the same false premise this PR was built on, so that issue has been corrected too.

None of that belongs in this PR. Flagging it because "keep it and fix it" is only the better call if someone eventually does the second half — and if the answer is instead that it should go, I'm happy to redo the removal properly, as an explicit API removal with a changelog entry rather than as dead code.

**One practical warning for whoever merges this.** #387 and this PR both append test blocks to the end of `test/test_core_loss_fixes.py`, and git's conflict region does not line up with the definition boundaries — it straddles a function call. Resolving it hunk-by-hunk, or by naively keeping both sides, produces a file that will not parse:

```
SyntaxError: '(' was never closed
```

The clean resolution is to take one side's file wholesale (`git checkout --ours` / `--theirs`) and append the other side's four added blocks. From this PR those are `test_detectors_elastic_is_refused_rather_than_ignored`, `test_detectors_elastic_guard_has_no_false_positives`, `_detectors_elastic_setup` and `test_transition_potentials_are_wrapped_by_the_live_isinstance_check`; from #387 they are, in file order, `TestPotentialEnsembleAccumulation`, `TestPrismPotentialEnsembleAccumulation`, `test_ensemble_indices_handle_a_multi_axis_potential_ensemble`, `test_ensemble_index_length_is_enforced`, `test_prism_driver_refuses_a_multi_configuration_potential` and `TestPrismLazyExitPlanes`.

I verified the union: with all four PRs merged and that conflict resolved that way, the full suite is green at 2012 passed, 318 skipped. This is the only conflict left among the four — #388 and #390 are independent of everything.

TomaSusi:
Import nit confirmed fixed in `0286dcaf` — `_DETECTORS_ELASTIC_MESSAGE` now comes through the module-level import block, and confirmed there's no circular import (`abtem/multislice.py`'s only reference to `abtem.waves` is under `TYPE_CHECKING`).

Also spot-checked your two informational notes on `GaussianProjectionIntegrals`, since they're worth having on record even if out of scope here: both check out. `get_gaussians`/`get_corrections` (`abtem/integrals.py:226-238`) really do compute a key, miss, recompute, and return without ever writing back to `self._gaussians`/`self._corrections` — the caches are permanently empty. And `_sinc_cache` (`abtem/integrals.py:303-310`) does key on the literal `"cpu"`/`"gpu"` string rather than a resolved device id — same aliasing class #388 fixes for `ScatteringFactorProjectionIntegrals`, just not touched there since it wasn't the subject. Not asking for either to be fixed in this PR — just confirming they're accurately described for whoever picks it up.

Nothing further from me — good to merge.

TomaSusi (approved):

PR #392, opened 2026-09-15
Key the remaining integrals.py caches on everything their values depend on

pzeiger:
**Stacked on #388.** Please review that one first; the diff shown here against `dev` includes it. Base it on `scattering-factor-cache-key` or merge #388 first, either works.

## What

`abtem/integrals.py` holds seven caches. #388 fixed one of them. This fixes the other six — four repaired, two deleted — and in doing so turned up a defect worse than either of the two this pair of PRs set out to fix.

**`QuadratureProjectionIntegrals._tables` is keyed on the chemical symbol alone.** The table is not a function of the element alone — `_calculate_integral_table` takes `inner_limit` from `min(sampling)`, which sets the radial grid it is built on. Reusing one integrator across a sampling sweep, an ordinary convergence study, serves the first sampling's table to every later grid:

```
_tables keys: ['Si']
shared(coarse-then-fine) == fresh(fine): False
max relative difference: 5.431e-01
```

A 54 % wrong finite-projection potential, silently. It is the same defect as #388's, in the same file, on the other projection path — #388's crashes on a shape mismatch, which is how it was noticed; this one has no shape to mismatch. It turned up by sweeping the file for the same defect shape rather than by a report.

**Every cached array here is built through `get_dtype`**, directly or via `spatial_frequencies` — every one but the disk, which is the subject of the next paragraph — so `precision` belongs in their keys as well. Without it a `float32` entry is served to a `float64` request — a ~1e-6 relative error that `np.allclose` at default tolerances calls correct. That includes the scattering-factor cache #388 added.

**`_sinc_cache` and `QuadratureProjectionIntegrals._device_arrays` key device-resident arrays on the literal `"cpu"`/`"gpu"` string**, which does not distinguish one GPU from another, so in a single process driving several GPUs an array cached under one device can be served to work on another. Latent in the supported dask-cuda layout (one worker process per GPU), which is why it has not bitten.

**`_sorted_disks` was unbounded, and its key was incomplete in both components.** A sampling sweep grew it without limit while its device-resident twin was capped.

An earlier revision of this PR left `precision` out of that key, arguing that the disk is just integer pixel offsets and that a 400-sampling scan found no case where `int(ceil(cutoff / min(sampling)))` moves. **That was wrong, and the scan is why.** A linear scan cannot find these; they have to be solved for. Solving `cutoff32/m <= s < cutoff64/m` gives a counterexample for every element tried — at `s = 0.281458955844481` silicon wants radius 18 at float32 and 19 at float64 — and serving the float32 disk to a float64 build is wrong by **4.6e-05 on a peak of 312**. The last ring lands at `idx == n - 2` in `interpolate_radial_functions`, just inside the final radial gridpoint, so it still contributes rather than clamping to zero.

**The symbol component was already there** and is not part of this fix — a claim to the contrary in an earlier revision of this body was wrong. `disk_key` contains the symbol on `dev` and on this PR's base alike; what it had was no *test* covering that component, which `test_the_sorted_disk_is_not_served_across_elements` now provides. (The mistake: a mutation dropping the symbol survived the suite, and I wrote that up as a defect rather than as the missing coverage it actually was.)

**`GaussianProjectionIntegrals._gaussians` and `._corrections` are deleted.** Both computed a key, missed, recomputed and never wrote back — empty for the life of the object. They are deleted here rather than repaired because repairing them in place was wrong twice over: the key omitted precision, and the populated cache rode into every task graph (abTEM pickles the integrator into the graph), inflating a warm frozen-phonon graph 12.6×. The follow-up PR makes that class actually usable and reintroduces a correctly keyed, transient cache there, where its speed-up is the point. Nothing regresses in between: a cache that never stored had no speed-up to lose.

## The fix

One `_DeviceArrayCache` behind all remaining caches: bounded, least-recently-used, race-free eviction (`popitem` and `move_to_end` are single C-level dict operations, and the `KeyError` a losing thread sees is caught).

Deliberately **not** a `functools.lru_cache`. A per-instance one wrapping a bound method is not picklable — it resolves by `__qualname__` and no longer matches the class attribute of that name — which breaks dask's `processes` scheduler and `distributed`, i.e. the multi-GPU layout these device keys exist to serve. It also breaks `EqualityMixin` comparison and `deepcopy`. All three were discovered and undone inside #388; the container carries the comment so the next person does not retry it.

`__eq__` returns True for any two caches, because a cache is incidental state and never identity. Without it `safe_equality` compares the populated dicts, hits numpy arrays and returns False through its `ValueError` guard, so a used integrator stops comparing equal to an identical fresh one.

This fixes the comparison at the **integrator** level only. A built `Potential` still compares unequal to an identical unbuilt one, because `Potential.build()` populates `_sliced_atoms` lazily — the same defect shape one level up. That is pre-existing on `dev` (`True` before build, `False` after) and out of scope here; an earlier draft of this PR claimed the fix reached that far, which it does not.

`__getstate__` drops the caches — **new in this revision**, and it belongs here rather than in the stacked PR, where it originally sat. abTEM pickles integrators into every task, and on a warm four-configuration frozen-phonon graph at 512² the caches this PR *keeps* inflate it 126× (scattering factor), 532× (quadrature) and 39× (gaussian), while a sampling sweep takes a pickled quadrature integrator to 69 MB. Deleting two never-storing caches on graph-inflation grounds while leaving those un-dropped would have been incoherent.

A used integrator now pickles to exactly the same size as a fresh one — `11233 → 11233 B`, against 2.1 MB / 28.7 MB / 1.07 MB before. Two integrators that compare equal now also tokenize equal, since `tokenize` pickles.

`_DeviceArrayCache` is a `collections.abc.Mapping`. `QuadratureProjectionIntegrals.tables` and `ScatteringFactorProjectionIntegrals.scattering_factors` are public properties that returned a plain `dict` before this container existed; without `Mapping`, `.keys()`, `.items()` and `obj[key]` all disappeared and the declared `-> dict` return type became false.

`__setstate__` normalises what comes back, to an **empty** cache. `_sinc_cache` post-dates #269, so an older pickle lacks it entirely; one pickled while these were plain dicts comes back with an object that has `.get` but not `.put`, so the *miss* path raised `AttributeError`.

Contents are deliberately dropped rather than carried over. Only a pre-`__getstate__` pickle can have any, and its entries are keyed in the **old shape** — `_sorted_disks` without precision, `_tables` without sampling — so reusing them would serve a value found under an incomplete key, which is the whole thing this PR exists to stop. Recomputing costs one miss.

Device keys are read off an array already on the target device (`array` for the sinc, `disk_counts_device` for the disk copy) and the allocation happens in that device's context — keying on one device while allocating on the ambient one is the half-fix #388 had to correct. Note the anchor is a formality at both sites today: both arrays are allocated locally, so they cannot disagree with the ambient device. It is the shape the callers need if either ever receives an array from elsewhere.

## Call sites

Seven caches, all in `abtem/integrals.py`, all reached only from within it:

| cache | class | read at | was wrong how |
|---|---|---|---|
| `_scattering_factors` | `ScatteringFactorProjectionIntegrals` | `get_scattering_factor` | precision missing from the key #388 added |
| `_tables` | `QuadratureProjectionIntegrals` | `integrate_on_grid` | keyed on symbol alone; sampling and precision missing |
| `_sorted_disks` | `QuadratureProjectionIntegrals` | `integrate_on_grid` | unbounded; key dropped precision |
| `_device_arrays` | `QuadratureProjectionIntegrals` | `integrate_on_grid` | device key is the `"gpu"` string |
| `_sinc_cache` | `GaussianProjectionIntegrals` | `integrate_on_grid` | device key is the `"gpu"` string |
| `_gaussians` | `GaussianProjectionIntegrals` | `get_gaussians` | never stored — deleted |
| `_corrections` | `GaussianProjectionIntegrals` | `get_corrections` | never stored — deleted |

`scattering_factors` is a public property with no consumer in `abtem/`, `test/` or `docs/`; nothing outside the module touches any of the others.

## Results

Infinite, finite, gaussian and explicit-quadrature potentials all hash identically to `dev` (sha256 over a GaAs cell at 128², `fft: numpy` to keep FFTW's plan choice out of it).

The shared-integrator case cannot be compared directly, because `dev` cannot run it — it raises the `(64,64)` vs `(128,128)` broadcast error that is #388's symptom. Against the right oracle, a *fresh* integrator per grid, it matches exactly:

```
dev,    fresh integrator per grid:  ['4a4dc221e0c5f59e', '338ccd8972da1a77', 'ddc57ee95287c699']
branch, ONE integrator, all three:  ['4a4dc221e0c5f59e', '338ccd8972da1a77', 'ddc57ee95287c699']
```

i.e. sharing an integrator now gives exactly what not sharing one gives, which is the whole claim.

The one case where a number *does* change is the one that was wrong: a shared integrator across samplings now gets the table for the sampling it asked for. The precision component is not decorative either — `get_integral_table("Si", (0.05, 0.05))` returns `radial_gpts[-1] = 5.0662611059` under float32 and `5.0662612052` under float64, so a float32 entry served to a float64 request really is a different table.

## Tests

**The previous revision of these tests did not test what it said it did, and this is the largest change in the PR.**

The helper took a `sampling` argument and never passed it to `Potential`, and with a fixed cell sampling is just `extent/gpts` — so every case that meant to vary sampling independently of the grid varied neither. Ten key components had no coverage at all, two of them worth **13 %** and **20 %** silent error.

Worse, the level was wrong. A potential-level shared-vs-fresh comparison is *structurally blind* to a key that drops the chemical symbol: both sides serve the first element's array to the second, so they agree with each other and the test passes. Each component is now probed one at a time **at the level of the cached method**, which is where it becomes observable. The helper also gained a lattice-constant knob, which is what makes "independently" possible at all.

`TestIntegratorCaches` now covers: every scattering-factor key component (symbol, gpts, sampling, precision); every integral-table component (symbol, sampling, precision); every sinc component (gpts, sampling, precision); both disk components (symbol, precision — the latter via the solved-for sampling above, the former via the H/Au cell); `__getstate__`; bounded growth; results independent of cache state; an integrator restored from an older-shaped pickle; the dead gaussian caches being gone; picklable/comparable/copyable; and concurrent access serving correct *values*, not merely raising nothing.

**All fourteen mutations fail the suite.** Each reverts exactly one source line.

`test_cache_is_least_recently_used_not_first_in_first_out`, which came in with #388, could not tell LRU from FIFO: it asserted only that the touched entry was present at the end, and under FIFO the miss that follows eviction re-inserts it at the tail. Deleting `move_to_end` left all tests green. It now counts recomputations, which is the quantity that actually differs. I've fixed it here rather than pushing to #388, which you've already approved — say the word if you'd rather it went back there.

`_DeviceArrayCache.get` takes `(key, default=None)`, so the public `tables` / `scattering_factors` really do behave like the dicts they replaced — a one-argument override would have shadowed `Mapping.get` and broken the compatibility the `Mapping` exists to provide. `test_the_public_caches_still_behave_like_the_dicts_they_replaced` covers the two-argument form, `in`, `keys()`, `items()` and subscripting, and fails against the one-argument version.

Full suite: 1984 passed, 0 failed, 318 skipped.

## Risk

Low for correctness — the keys strictly widen and no call site changes. The one behavioural change beyond keys is `__getstate__`: an integrator now unpickles with empty caches. That is the intent (a populated cache is worker-local state, not graph payload), and every value is recomputed on first use.

The `_device_arrays` and `_sinc_cache` device keys need two GPUs to exercise and could not be tested here; they are verified only structurally, as in #388.


TomaSusi (commented):
Solid fix — verified all six cache-key repairs have matching key tuples at every put/get site, and independently reproduced two of the headline numbers: the float32/float64 disk-radius boundary for silicon at that sampling, and the ~54% error from `_tables` being keyed on symbol alone. Two things worth a look:

1. **`_DeviceArrayCache.get()` breaks the `Mapping`/`dict` two-argument contract.** It's `def get(self, key):` — a single-argument override that shadows `Mapping`'s inherited `get(key, default=None)`. Since the whole point of making this a `Mapping` was to keep `tables`/`scattering_factors` behaving like the plain dicts they replaced, this quietly breaks exactly that: `integrator.scattering_factors.get(key, "not cached")` (completely idiomatic dict usage) raises `TypeError` instead of returning the default. Nothing in-repo calls it that way today, so it's latent, but it undercuts the stated reason for using `Mapping` in the first place.
2. **The `_sorted_disks` "symbol-blind" story doesn't check out against this PR's own base.** I checked `origin/scattering-factor-cache-key` (i.e. #388) directly: `disk_key = (chemical_symbols[number], tuple(sampling))` already includes the symbol there. The precision-missing part of the fix is real and correctly done; the H-before-Au / 17-pixel-truncation narrative appears to describe a bug that wasn't actually present in the code being diffed against. Not a code problem — just flagging since your write-ups are usually airtight and this detail doesn't hold up on inspection.

Two very minor concurrency nits in `_DeviceArrayCache.put()` — a possible gratuitous extra eviction under a concurrent same-key miss race, and a soft rather than hard size bound under concurrent distinct-key writes. Neither corrupts data, both low severity, not worth blocking on.

pzeiger:
Both points addressed in `82e904cc`, and you were right about the second one — thank you for checking it rather than taking the write-up at face value.

## `get()` two-argument contract

Reproduced, then fixed:

```
before:  _DeviceArrayCache.get() takes 2 positional arguments but 3 were given
after:   two-arg get -> 'not cached'   one-arg miss -> None   hit -> the cached array
```

`get(self, key, default=None)`. New test `test_the_public_caches_still_behave_like_the_dicts_they_replaced` covers the two-argument form, the one-argument miss, `in`, `keys()`, `items()` and subscripting; it fails against the one-argument version, which I checked rather than assumed.

One design note, since it is the obvious follow-up question: the LRU touch stays on `.get()` and deliberately does **not** move into `__getitem__`. I built that variant to see, and it is not viable — `Mapping` routes `get`, `__contains__`, `values()` and `items()` through `__getitem__`, so refreshing recency there makes `{k: cache[k] for k in cache}` and `list(cache.values())` raise `RuntimeError: OrderedDict mutated during iteration`. Those are exactly the dict operations the public properties are meant to support. Side-effect-free subscript and membership are what keep them dict-like; refreshing on `.get` alone is the only split that works.

## The `_sorted_disks` symbol story

You are right and I was wrong. `disk_key` already contained the symbol, on both sides:

```
origin/dev                          disk_key = (chemical_symbols[number], tuple(sampling))
origin/scattering-factor-cache-key  disk_key = (chemical_symbols[number], tuple(sampling))
```

Only the precision component was missing, and only that was added.

What happened is worth stating plainly, because it is a failure mode I would rather not repeat: a mutation dropping the symbol from the key **survived** the test suite, and I wrote that up as a present defect. A surviving mutant means missing test *coverage*, not a defect in the code — I conflated the two and then invented the H/Au truncation scenario to explain a bug that was never there.

`82e904cc` retracts it in the commit message and in the comment at the key site, which now says the symbol component predates the commit and was never missing. `test_the_sorted_disk_is_not_served_across_elements` therefore covers that component for the first time rather than fixing anything — the H/Au case is a hypothesis justifying a test, and the docstring is phrased subjunctively for that reason. **The PR description above has been updated too**, since the retracted claim appeared there as well.

No count in the PR changes: "four repaired, two deleted" counts caches, not key components, and `_sorted_disks` was genuinely broken on `dev` twice over — an unbounded plain dict and a precision-blind key.

## The two concurrency nits

Left alone deliberately, and I should say why rather than just not doing them.

The extra eviction under a concurrent same-key miss costs one redundant recomputation of a value that is now bit-identical to the one it replaces — the keys are complete, so a recompute cannot differ. It buys nothing to serialise the hot path against it.

On the soft bound, I tried to measure the overshoot and could not produce one: 16 threads × 20 000 operations with `sys.setswitchinterval(1e-9)`, at `maxsize=4` forcing 100 % thrash and at `maxsize=32`, gave a maximum observed length of exactly `maxsize` in both cases, no exceptions, no wrong values. So the bound is soft by construction — `while len >= maxsize: popitem()` then `entries[key] = value` is unguarded — but I have no demonstration that it is observably exceeded, and I would rather report that honestly than either claim it is hard or leave the docstring silent. The docstring says soft.

Suite: 1984 passed, 0 failed, 318 skipped.

TomaSusi:
I don't see this being formally stacked on #388 – would be good to do that so the stacked merge goes cleanly.

TomaSusi (approved):

pzeiger:
Hi @TomaSusi!
I have just deleted the old stack and made a new one, so it should be all clean now.
Best,
Paul

PR #393, opened 2026-09-15
Make GaussianProjectionIntegrals usable: parameters, GPU, and caching

pzeiger:
**Third in a stack: #388 → #392 → this.** Review those first; only the last commit is new here. The diff against `dev` includes all three.

## What this class is for

`GaussianProjectionIntegrals` is the only integrator that is both **finite** (z-resolved within a slice) and **periodic**. `ScatteringFactorProjectionIntegrals` is periodic but z-unresolved; `QuadratureProjectionIntegrals` is z-resolved but needs padding. That combination is what you want for a finite-projection potential on a genuinely periodic cell, and it is reachable today through the documented `Potential(integrator=...)` parameter.

It was unfinished in three ways, and this PR finishes the wiring. **It deliberately stops short of a physics judgement** — see *What is still open* at the end. That call is better made once the machinery works, and separately from this diff.

What it was meant to be is still readable in `test/test_atomic_potential.py`, where three tests for it sit commented out: `test_gaussian_projection_integrals` (uncorrected vs Lobato-corrected, against an analytical projection at `rtol=1e-6`), `test_finite_gaussian_projection_integrals` (parametrized over `fourier_space`), and `test_finite_projections` (against `ProjectionQuadratureRule` for carbon). Those are the intent; none of them runs today. This PR does not resurrect them — they use an API that no longer exists — but the last one is exactly the comparison the open question at the end is about.

## 1. Dead parameters

`parametrization` and `gaussian_parametrization` were validated, stored, and never used: `get_gaussians`, `get_corrections` and `gaussian_projection_weights` all fell through to module defaults. Both are wired now. The defaults are the values the code hardcoded, so a default build is bit-identical — checked against `dev` rather than argued: same sha256 on a GaAs cell at 128², in both float32 and float64. (`cutoff_tolerance` was already live — two dead parameters, not three.)

`get_corrections` now validates too, not just `get_gaussians`: it takes `gaussian_parametrization` as its `long_range` term, so a non-Gaussian one returned a plausible-looking array rather than raising. It was safe only because `integrate_on_grid` happens to call `get_gaussians` first — an ordering accident, and `get_corrections` is public.

A `RuntimeError` guarding `len(weights) != len(gaussians)` is removed: both lengths come from the same `scaled_parameters` call on the same parametrization, so it could not fire. It only looked like a safety net.

`integrate_on_grid` also accepted a `fourier_space` flag it never read — a leftover of that commented-out test. No caller passes it (`iam.py:968` is the only caller of any integrator's `integrate_on_grid`) and the abstract signature does not declare it, so asking for a reciprocal-space result silently returned a real-space one. Removed, so the request is a loud `TypeError` instead.

The Gaussian count was hardcoded to `range(5)` while the source became configurable, so a six-term parametrization silently dropped its last term and a three-term one raised `IndexError` from inside the loop. The bound follows the parametrization now.

**Wiring `gaussian_parametrization` without validating it would turn a dead parameter into a silent physics error.** Lobato has the same `(2, n)` parameter shape as Peng but a different functional form; it runs, and gives a potential whose plane integral is 3.3× too large for carbon and 4.8× for silicon — finite everywhere, nothing to notice. So the form is checked *functionally* rather than by name: rebuild the scattering factor from the parameters as a sum of Gaussians and compare against the parametrization's own. Peng agrees to 1.8e-07 at float32 and 1.8e-16 at float64 across all 98 elements it covers; Lobato is off by 2.3 (C) and 3.8 (Si), and Kirkland — the nearest rejection — by 0.275. Three decades of headroom on either side of the 1e-4 tolerance.

**The comparison is made on the grid the call is about to use.** An earlier revision sampled a fixed `linspace(0.01, 4.0, 32)`, which is 2.3 % of the k² a 128² build evaluates and 0.1 % of a 512² one — and 32 points across that window sit 0.13 apart, so a narrower feature fell between them as well. Both gaps were reachable: a Peng subclass carrying a bump at k²=20, or one between two sample points, cleared the check and built a potential several per cent to tens of per cent wrong. **A check that validates 2 % of its domain is worse than no check, because it reads like one.** Using the real grid costs nothing — `gaussian_projected_scattering_factors` computes it one call later.

**Both sides are evaluated at the configured precision.** Mixing them made the verdict depend on `abtem.config`: `scaled_parameters` is float64 while `Parametrization._get_function` casts to `get_dtype`, so a five-term superposition whose last two terms nearly cancel was accepted at float64 and rejected at float32 — a script that worked would start raising on a precision switch.

The check runs **per element, on the element being used**, on cache miss — not once at construction on a fixed one. A first version validated carbon only, which rejected the shipped `peng_ionic.json` (all ions, no carbon) with an error claiming it had no parameters at all, and would have passed a parametrization whose carbon entry was sound and whose silicon entry was not — a ~10 % error, finite everywhere. That is exactly the failure the check exists to prevent, moved one element sideways.

One consequence worth flagging: the check also rejects a parametrization whose parameter table and whose own `scattering_factor_k2` disagree on the number of terms. That is wanted — the extra Gaussians would go into the field while `get_corrections` and `cutoff` still saw five — but it means `range(len(gaussians))` cannot be reached with any *shipped* parametrization, because Peng is abTEM's only Gaussian-form class and its `scattering_factor_k2` is hardcoded to five terms. Real n-term support has to fix that first. Until then the loop bound is correct in principle and exercised only by a custom parametrization.

## 2. It could not run on GPU

`TypeError: Unsupported type <class 'numpy.ndarray'>` — on `dev`, and equally on the two cache commits below this one, which change nothing here. Two boundaries:

- the parametrization helpers are host-only, so their results have to move across before meeting a device array;
- the deltas were superposed into a **complex** array, and cupy's scatter-add has no complex overload. They go into a real array now and are cast afterwards, the way the infinite projection already does.

Transfers cast to the configured precision on the way, since the parameters are float64 whatever the run is — to a **real** dtype for the corrections too, where an earlier complex cast saved nothing at float32, doubled the transfer at float64, and left cpu and gpu disagreeing on dtype.

**Verified on ROCm** (gfx1151): CPU and GPU agree to 1.1e-07 relative in float32 and 1.4e-16 to 2.1e-16 in float64, across six elements and both precisions.

**CUDA is unverified and two things want checking there.** There is no custom kernel, so no new HIP/NVRTC surface; the only platform-sensitive thing is cupy's scatter-add dtype support, which is a generic cupy restriction rather than a ROCm one. The second is the float64 path, which scatters into a float64 array and so leans on a double `atomicAdd` — cupy documents that less clearly than the float32 case. If someone has a CUDA box, those two are the whole check.

**Not deterministic on GPU.** `_integrate_gaussians` does not filter positions by slice, so every atom contributes to every slice; where two atoms share a pixel, `atomicAdd` reorders. Ordinary crystals come out bit-identical, but a contrived column of 2000 atoms gives a 7.4e-07 float32 spread. Last-bit, but not determinism — same family as the `projection="finite"` nondeterminism that already exists on `dev`, and not made worse here.

## 3. Caching, folded in

`integrate_on_grid` runs **once per slice per species** and recomputed both parametrization arrays every time — 92 calls for 2 distinct combinations on a 46-slice GaAs cell, i.e. 46× redundant, and 65–70 % of CPU build time (89 % on GPU at 1024², which is why the device could not help).

The preceding PR deletes these two caches, correctly: as written they never stored, their key omitted precision, and they rode into every task graph. They come back here with the key they always needed — element, grid, sampling, precision — bounded, and excluded from `__getstate__`, so they are per-worker state rather than graph payload. A used integrator pickles to exactly the same size as a fresh one.

```
gpts     before    after   speedup   (CPU, 45-slice GaAs, warm, min of 3)
64^2     0.116s   0.028s     4.1x
256^2    0.599s   0.167s     3.6x
512^2    2.558s   0.687s     3.7x
```

Measured twice on an idle machine; the repeat gives 4.3× / 3.5× / 3.9×. These already carry the form check, which now evaluates on the build's own grid and costs 15 ms at 512², four times per two-species build.

**These are warm numbers.** Cold one-shot timings are dominated by numba JIT, cupy kernel compilation and FFT planning, and invert the conclusion. They are folded into this PR rather than sent separately because a feature that is 46× redundant on its hot path is not really usable, and splitting them would mean reviewing the same function twice.

## Cutoff

The difference against the quadrature integrator is not mysterious: this class defaults to `cutoff_tolerance=1e-3` against quadrature's `1e-4`, and its `cutoff()` evaluates the *Peng* potential while the field it builds is `peng + (lobato - peng)`, i.e. Lobato at short range. Matching both reproduces quadrature's number exactly. Whether `cutoff()` should use the correction parametrization is a physics choice and is left alone here.

## Tests

`TestGaussianProjectionIntegralsUsable` in `test/test_atomic_potential.py` covers: builds on both devices agreeing; each of the three parametrization consumers following its configured parametrization (pinned by *value*); a non-Gaussian feature caught **wherever it sits** — outside the old window and between its sample points; `get_corrections` validating; the verdict not depending on precision; an element the parametrization lacks not rejecting it; a bad entry caught at the element that uses it; the Gaussian count following the parametrization; defaults unchanged by the wiring; the host path not casting; every key component of *both* caches; and `integrate_on_grid` accepting no parameter it ignores.

Three of these replace tests that could not fail:

- the Gaussian-count test asserted on `inspect.getsource`, so it would have failed a correct `zip(gaussians, weights)` refactor and passed a broken loop body. It is now behavioural, using a seven-term split of Peng's silicon entry. That needs an n-term-aware subclass, because a seven-term table on the shipped Peng is *correctly* rejected — its own function is hardcoded to five — which is exactly why the loop bound cannot be reached with anything shipped;
- the precision test asserted `served.dtype == correct.dtype`, which is float64 on both sides at either precision;
- the GPU tolerance was 5e-7 with a comment claiming it guarded the transfer cast. It cannot: dropping the cast leaves CPU–GPU agreement identical to seven significant figures. And 5e-7 was a property of the one cell tested — the worst of 100 element/grid/thickness combinations reaches 5.7e-07, so it sat 14 % from failing. It is now 2e-6 with an honest comment.

**All nine source mutations fail the suite.** The one that needed real work was the mixed-dtype validator: catching it needs a parametrization whose float32 evaluation error exceeds the tolerance while its exact deviation is zero — five near-cancelling terms at amplitude 1e5, which the pre-fix code rejects at float32 and accepts at float64, and the fix accepts at both with deviation exactly `0.00e+00`.

**One line is deliberately left uncovered, and the commit says so rather than pretending otherwise:** the device-transfer dtype cast changes bytes moved, not results. The byte counts are in §2.

Full suite: 2013 passed, 0 failed, 319 skipped (1984 / 318 on the branch below this one).

## What this PR deliberately does not fix — please look here hardest

**The short-range correction is not resolved along z.** This is pre-existing behaviour, unchanged by this PR, and worth a reviewer's attention precisely because this PR is what makes the class reachable in practice.

The Gaussian term *is* z-resolved: each atom contributes to every slice, weighted by the analytic z-integral of each Gaussian between that slice's limits. The correction term is not — `_integrate_corrections` filters atoms to the slice they sit in and adds the whole infinitely-projected difference at unit weight:

```
V_slice  =  ∫_slice Peng dz  +  [ ∫_-∞^+∞ (Lobato − Peng) dz ]
                                  ↑ should be ∫_slice
```

The total is right: each atom's correction is counted exactly once, so the slices sum to the full Lobato projection. That is why this has never surfaced as a visibly wrong answer.

**The reasoning behind it is sound in spirit.** Peng's five Gaussians reproduce `f(k²)` to within 0.3 % out to k² ≈ 4 Å⁻² and then fall off too fast — Gaussians decay as `exp(−bk²)`, the true factor as ~1/k² — so the difference against Lobato is dominated at large k², i.e. close to the nucleus. If it were *confined* there it would fit inside any reasonable slice, and adding it whole would be correct.

**It is not confined there.** The two parametrizations cross each other repeatedly: the ratio runs 0.925 at r = 0.05 Å, 1.034 at 0.1, 1.005 at 0.3, 1.011 at 0.5, 0.987 at 1.0, 0.906 at 2.0. A five-Gaussian fit wiggles around a different functional form rather than agreeing with it outside the core, and those sub-percent wiggles carry real weight out to ~3 Å:

| slice Δz | of the correction | of the atom's own potential |
|---|---|---|
| 0.5 Å | 11 % | 44 % |
| 1.0 Å | 20 % | 67 % |
| 2.0 Å | 31 % | 90 % |

The correction is **less localised in z than the atom it corrects**, so at ordinary slicing most of it is placed in the wrong slice.

Measured directly against `QuadratureProjectionIntegrals` at matched cutoff, single Si atom: the largest per-slice deviation is 0.16 % for one slice holding the whole atom, 0.62 % at Δz = 1 Å and 3.5 % at Δz = 0.125 Å — growing as 1/Δz — while the total stays within 1563–1643 ppm throughout. **Refining the slicing makes it worse**, which is the signature of a model choice rather than a discretisation error.

**A separate effect, not to be confused with it:** at the nuclear cusp the two integrators differ by a fixed ≈190 V. Both peaks *diverge* as sampling is refined (842 → 1948 V here, 670 → 1758 V for quadrature, over sampling 0.094 → 0.006 Å), so the same difference expressed as a percentage of the peak is just a readout of the sampling — 26 % down to 11 % across that range. Which of the two is closer to the true projected potential is not established here.

Ruled out as explanations, each checked: periodicity (the tail ratio is cell-independent across 8/16/24/32 Å cells), the radial cutoff and quadrature's taper (with matched `cutoff_tolerance` and one slice they agree to 1 % at r = 2 Å and 5 % at r = 3 Å, diverging only near quadrature's own cutoff), and grid convergence of the total (1.0109 → 1.0001 over gpts 64 → 1024).

**Why this PR leaves it alone.** Giving the correction a real z-profile means a numerical z-quadrature for that term, since `Lobato − Peng` has no closed-form z-integral — which would cost the analytic speed that makes this class worth having. That is a physics and performance trade-off, not a bug fix, so the behaviour is left exactly as it was and documented in the class docstring: prefer `QuadratureProjectionIntegrals` where the distribution of potential *within* a slice matters, and this one where periodicity does.

Two smaller questions in the same area, also left alone: `cutoff()` evaluates the *Peng* potential while the field built is Lobato at short range, and this class defaults to `cutoff_tolerance=1e-3` against quadrature's `1e-4`. Matching both reproduces quadrature's cutoff exactly.

## Risk

Contained. The class has no live caller anywhere in `abtem/`, `test/` or `docs/` — the only references are the commented-out tests above — so nothing existing can regress: a default-constructed integrator is bit-identical to before, and every behavioural change is on a path that previously either did nothing (the dead parameters) or raised (GPU).

One thing noticed and deliberately **not** touched: `FieldIntegrator.integrate_on_grid` still names its first argument `positions` and documents it as "2D array of xy-positions", while all three concrete integrators take an `Atoms`. That is a stale abstract signature, not a defect in this class, and renaming it belongs in its own change.


TomaSusi (commented):
Great work — GPU fix, dead-parameter wiring, and cache re-addition all check out. I specifically verified the "scatter into real then cast" claim is mathematically identical to the old crashing complex-scatter approach (traced `superpose_deltas`'s inputs — the deltas being scattered are always real-valued for both the Gaussian and correction terms, never complex, so there's no discrepancy). Your own description of the deliberately-unfixed z-resolution issue matches the code exactly too, for what that's worth.

One real finding: the class docstring's "exact for the total" claim is only true up to the `_GAUSSIAN_FORM_TOLERANCE` (1e-4) the functional validation check enforces, not unconditionally. Concrete counter-example: a `PengParametrization` subclass with a narrow bump near the edge of a 128² grid's k² range gets accepted by `_validate_gaussian_form`, but then produces a real, nonzero discrepancy (0.006 absolute / 5e-5 relative) between the Gaussian+correction sum and the true target — because the check bounds `max|reconstructed - own| / max|own|` globally, which tolerates large local relative error wherever `own` is small compared to the grid's peak (typically low-k). For the shipped Peng/Lobato default this is negligible (~1.8e-7), so it's not a live bug, but the docstring states exactness unconditionally rather than "exact up to the validation tolerance" — worth a one-line wording fix so a future custom `gaussian_parametrization` doesn't get an unpleasant surprise from a technically-true-sounding guarantee.

Nothing else — good to merge from here.

TomaSusi:
I would personally add that this was work in progress started by Jacob a long time ago and never finished – will be very curious to see where this might prove useful.

pzeiger:
Fixed in `adc23265`, and the point turned out sharper than a wording nit — thank you.

The reason the guarantee is conditional is structural, not incidental: the Gaussian term is built from `scaled_parameters` while the correction subtracts the parametrization's own `projected_scattering_factor`. The identity only holds where those two agree, and `_validate_gaussian_form` bounds their disagreement **globally**, as `max|reconstructed − own| / max|own|`. So a discrepancy parked where `own` is small relative to the grid's peak costs almost nothing in the quantity being checked while being large where it sits.

I scanned amplitude and position to find where that bites, rather than tuning until it agreed with your example — Si, 128² grid, bump width 40:

```
 amp  centre  global dev   verdict   local rel   total rel
0.30     150   3.302e-04  rejected   2.307e-01   1.913e-03
0.05     150   5.504e-05  ACCEPTED   4.761e-02   3.193e-04
0.01     150   1.101e-05  ACCEPTED   9.899e-03   6.400e-05
```

At k² = 150, `own` is 1.1e-03 of peak. So a Peng subclass with a bump there is **accepted** at a global deviation of 5.5e-05 while differing by **4.8 % locally**, leaving the total off by 3.2e-04 — the same effect as your 0.006 / 5e-5, arrived at independently.

The docstring now states the bound and carries that counter-example, rather than just softening the sentence. For the shipped Peng/Lobato default the disagreement is ~1.8e-07 and the total is exact to that, so this is not a live bug — it is a guarantee that reads stronger than it is, which is precisely what a custom `gaussian_parametrization` would lean on.

I kept it textually separate from the z-distribution caveat. The two are easy to conflate — both make the result "not quite exact" — but they have different causes and different fixes: this one is a tolerance on a functional check, that one is a model choice about where the correction is deposited.

The branch is rebased onto the updated #392 (which picks up the `get(key, default=None)` fix and the disk-key retraction). Bit-identity to `dev` re-verified after the rebase: same sha256 at both precisions on a GaAs cell at 128². Suite: 2013 passed, 0 failed, 319 skipped.

TomaSusi (approved):

Issue #398, opened 2026-09-15
PRISM lazy-scan axis bugs: exit_planes>1 crashes SMatrix.build, blocks the disable_s_matrix_chunks and no-chunks-reduce paths


TomaSusi:
Three related bugs in the PRISM lazy-scan path were grouped here as sharing one root cause, found while fixing #387 (the core-loss potential-ensemble accumulation bug). **Update 2026-09-15: item 1 is fixed by #404, and #404's investigation corrected item 3's attribution — see below.**

## Root cause

`Waves.transition_potential_multislice` ends in a module-level `reduce_ensemble` call that honours `_squeeze` — so a scan passed as a bare tuple `(x, y)` rather than a `BaseScan` correctly gets its position axis squeezed out of the result. The PRISM lazy branch (`SMatrix.transition_potential_scan(lazy=True)`) only reaches `measurement.reduce_ensemble()`, which handles `_ensemble_mean` but not `_squeeze` — so the two paths disagreed on axis bookkeeping whenever a bare-tuple scan was involved.

## 1. Tuple-scan squeeze in the lazy PRISM path — FIXED by #404

```
GridScan      eager (5, 2, 2, 64, 1)  lazy declared (5, 2, 2, 64, 1)  computed (5, 2, 2, 64, 1)  to_cpu OK
tuple (x, y)  eager (5, 64, 1)        lazy declared (5, 1, 64, 1)     computed (5, 64, 1)        to_cpu -> RuntimeError
```

[#404](https://github.com/abTEM/abTEM/pull/404) moved the squeeze to the two levels the multislice oracle uses — the per-block driver in `core_loss.py` now applies only the ensemble-mean half, and `SMatrix.transition_potential_scan` applies the module-level `reduce_ensemble` once, in both its eager and lazy return paths. Verified independently against the PR's own shape matrix and a full test run — this symptom is closed.

## 2. `SMatrix.reduce`'s `disable_s_matrix_chunks=True` lazy branch — still open, still masked

[...]

PR #404, opened 2026-09-15
Squeeze the PRISM scan axis once, not once per block

pzeiger:
**Stacked on #387.** Review that one first; the diff shown against `dev` includes it. Only the last commit is new here.

## What

`validate_scan` turns any non-`BaseScan` scan — `(x, y)`, `[(x, y)]`, `np.array([[x, y]])` — into a `CustomScan` whose position axis is tagged `_squeeze=True`. The only reader of that flag anywhere is the module-level `reduce_ensemble` in `waves.py`, and **where it is called from** decides whether the bookkeeping stays consistent.

`Waves.transition_potential_multislice` calls it **once, on the assembled result**, after the graph is built — so blocks and declared chunks both keep the axis, and only the finished object loses it.

`prism_transition_potential_scan` called it **inside every block**. It is the per-configuration driver: `SMatrix` invokes it once per dask block and once per potential-ensemble member, which is below the level that declares `chunks` and below the level that accumulates the ensemble. Two consumers were therefore handed a shape the blocks did not produce:

```
s_matrix.py   lazy branch    chunks += scan.shape
s_matrix.py   eager branch   allocate from dummy_probes(scan)
```

## Why it matters — the loud version is not the interesting one

With a single exit plane this does **not** crash. It returns a result whose *rank* depends on whether the potential is a `FrozenPhonons` ensemble and on whether the call was lazy:

```
scan          path                        before          after
GridScan      multislice (oracle)         (3,3,143,1)     unchanged
GridScan      PRISM eager / lazy          (3,3,143,1)     unchanged
tuple (x,y)   multislice (oracle)         (143,1)         unchanged
tuple (x,y)   PRISM eager, no ensemble    (143,1)         unchanged
tuple (x,y)   PRISM lazy,  no ensemble    (1,143,1)   ->  (143,1)
tuple (x,y)   PRISM eager, 3 phonons      (1,143,1)   ->  (143,1)
tuple (x,y)   PRISM lazy,  3 phonons      (1,143,1)   ->  (143,1)
```

Values are right; the axes contract is not, and downstream code keys off it (`_scan_shape`, `_scan_sampling`, dose application, `interpolate`, `to_cpu`). Three of four tuple-scan configurations disagreed with the oracle. The version that *raises* needs the exit-plane axis #387 restores — on `dev` it is masked, because every scan form is already broken there.

## The fix

At the two levels the oracle uses. The driver keeps only the ensemble-mean half (`measurement.reduce_ensemble()`), and `SMatrix.transition_potential_scan` applies the module-level `reduce_ensemble` once, in both its eager and lazy return paths.

## Verification

**Nothing that worked moves.** GridScan results are bit-identical before and after — sum `1.19625099e-07`, eager `==` lazy bitwise, with and without an ensemble. The single-position result now equals the same point taken from a 1×1 GridScan to `0.0e+00`.

**The residual PRISM-vs-multislice difference is the S-matrix approximation, not this change**: it measures `1.983e-02` on GridScan on *both* branches.

Checked against the ways this fix could be subtly wrong rather than only the case it targets:

- **the ensemble mean is not applied twice.** The driver's `m.reduce_ensemble()` consumes the `_ensemble_mean` axes, so the outer call finds none and returns `self`. Confirmed by value: `ensemble_mean=True` equals the manual mean of the `ensemble_mean=False` result to `0.0e+00`.
- **`FrozenPhonons(ensemble_mean=False)`**: the configuration axis survives and only the position axis is squeezed, matching the oracle's shape and axis types exactly, eager and lazy.
- **multi-detector calls** still return a `ComputableList`, identical shapes and metadata counts eager and lazy, since `reduce_ensemble` recurses over lists.

## Tests

`TestPrismScanAxisSqueeze` in `test/test_core_loss_fixes.py`: all three single-position scan forms × both ensembles × eager and lazy against the multislice oracle, plus rank stability across laziness, rank stability across ensemble-ness, a GridScan regression guard, and the 1×1-GridScan value equality.

It compares **axis counts and types, not full shapes**, because PRISM and multislice legitimately differ on detector bin count (32 vs 50) through the S-matrix's reciprocal sampling — that is algorithmic and pre-existing, and asserting on it would make the test fail for the wrong reason.

Restoring the per-block squeeze fails 11 of the 17; keeping the driver change but dropping the outer squeeze fails 12. Full suite: 1991 passed, 0 failed, 318 skipped.

## Deliberately not in this PR

**`SMatrix.reduce`'s `disable_s_matrix_chunks=True` lazy branch** has the identical exit-plane omission, but it is unreachable — elastic `exit_planes > 1` dies upstream in `SMatrix.build` on every path. Fixed alone it would be an unverifiable diff, and fixing `build` first turns it from an exception into a *silent* structural corruption, so the two belong together in their own PR.

**`s_matrix.py:629/755/812` are not defective**, contrary to an earlier description of this family. They relay `s_matrix_array.ensemble_axes_metadata`, which is the correct source. Verified rather than assumed: `_single_rechunk_reduce` has no live caller, and `_map_blocks` is reachable only from inside `_multiple_rechunk_reduce`, whose only call is commented out at `s_matrix.py:1361` (the `_single_rechunk_reduce` one at `:1366`, behind a `NotImplementedError`). Only `_no_chunks_reduce` is live, at `:1357`, and it is correct. The `RuntimeError` attributed to these sites comes from `SMatrix.build`'s lazy branch via `_lazy_reduce`.

## Known residual

With an `AnnularDetector` (base shape `()`) a single-position PRISM scan returns `(1,)` with a leftover `PositionsAxis` where the oracle returns `()`. **Pre-existing** — identical on `2ce2f61c`, checked against a base checkout — because `ArrayObject.squeeze` is a silent no-op for base-less measurements. Not closed here.


TomaSusi (commented):
Verified the fix thoroughly — ran the new tests plus the full `test_core_loss_fixes.py`/`test_prism.py`, and hand-checked the shape matrix from the PR description against the code directly. The core fix is correct: the per-block driver's `measurement.reduce_ensemble()` and the outer `reduce_ensemble` at the `SMatrix.transition_potential_scan` level don't double-reduce — the per-block call only ever consumes a length-1 potential-ensemble axis, and the real ensemble averaging happens exactly once.

I also independently verified both of the "corrections to the earlier bug-family description" hold up, which matters since they touch #398:
- `_single_rechunk_reduce`/`_map_blocks` really are dead code — zero live callers, confirmed by grep and by the fact that both are only reachable through commented-out branches.
- The `disable_s_matrix_chunks=True` lazy branch really is currently unreachable — reproduced directly: `SMatrix.build(lazy=False)` on a potential with `exit_planes > 1` crashes upstream with the same `ValueError` regardless of that flag, before its own exit-plane-omission bug could ever matter.

I've updated #398 to reflect both corrections and mark the tuple-scan-squeeze symptom as fixed here.

One new finding, not previously known: the "known residual" you flag (`AnnularDetector` on a single-position PRISM scan returning `(1,)` instead of `()`) has a concrete root cause — `ArrayObject.squeeze`'s `shape[: -len(self.base_shape)]` evaluates `-len(self.base_shape)` as `-0 == 0` when `base_shape` is empty, so `shape[:-0]` is `shape[:0]`, always empty. `squeeze()` silently no-ops for any base-less measurement. It's genuinely pre-existing and not something this PR should absorb — filed as [#407](https://github.com/abTEM/abTEM/issues/407) since it's a real, one-line-fixable bug independent of this PR's scope.

Nothing blocking here. Good to merge.

TomaSusi (approved):

PR #405, opened 2026-09-15
Gate GPU tests on a usable device, not on cupy importing

pzeiger:
## What

The test suite used **"cupy is importable"** as a proxy for **"a GPU is usable"**, in three idioms across eight places. cupy imports fine with no device, so on a machine where cupy is installed but no GPU is available every one of those tests is collected and run, and fails later at the first allocation inside the array module:

```
HIP_VISIBLE_DEVICES="" pytest test/test_array.py -k gpu
FAILED test/test_array.py::test_indexing[images-gpu-True] - cupy_backends.cud...
```

The predicate is simply wrong, not unlucky:

```
cp imported (what the skips used): True
_gpu_count() (the truth)         : 0
```

## Why it matters

This bites wherever cupy is present without a working device: a hidden device, a busy or reset GPU, a container without `/dev/kfd`, a CI image that pip-installs cupy on a CPU runner. It also removes the obvious way to isolate GPU-specific behaviour — hiding the device to compare against a CPU-only run turns the suite red instead of skipping the GPU half. That is how this was found.

## The fix

**The correct predicate already existed in the tree, in two places.** `test/utils.py:124` `_gpu_count()` calls `cp.cuda.runtime.getDeviceCount()` and returns 0 on any driver error, and `requires_multigpu` already used it. And in `test_fft.py`, `test_cufft_cache_auto_resolves_device_relative` does `importorskip` **and** a `getDeviceCount` check — while `test_oversized_plan_bypasses_cache`, twenty lines below it in the same file, does only the `importorskip`.

So this is not a new convention, it is the existing one applied consistently:

| file | change |
|---|---|
| `utils.py` | `gpu` param gated on `_gpu_count() < 1`; reusable `requires_gpu` marker added beside it |
| `test_multigpu_logic.py` | 4× `skipif(cp is None)` → `requires_gpu` |
| `test_ionization.py` | 2× `skipif(cp is None)` → `requires_gpu` |
| `test_prism.py` | 1× `skipif(cp is None)` → `requires_gpu` |
| `test_fft.py` | bare `importorskip` → + `requires_gpu` |
| `test_potential_chunking.py` | bare `importorskip` → + `requires_gpu` |

Only the first was in the original report; the other seven came out of sweeping for the predicate rather than fixing the reported instance.

## Verification

Both directions, since a gate can fail by skipping too much as easily as too little:

```
file                      GPU hidden                  GPU visible
test_fft                  23 passed,   4 skipped      27 passed
test_ionization           26 passed,  21 skipped      47 passed
test_multigpu_logic       28 passed,   4 skipped      32 passed
test_potential_chunking   58 passed,   2 skipped      60 passed
test_prism                31 passed, 183 skipped      54 passed, 160 skipped
```

The visible-device counts are unchanged from before the commit, so nothing that ran has stopped running; with the device hidden the suite now skips instead of failing.

## No test added, and why

These are pytest skip predicates evaluated at collection time. A test of them would either assert on source text or re-implement the predicate and compare it against itself — neither fails if a gate regresses. The table above is the verification, and the change is test-only.

## Risk

Test-only; no `abtem/` code is touched. The one way this could go wrong is over-skipping on a machine where a GPU *is* present, which the right-hand column above rules out.


TomaSusi (commented):
Clean fix, nothing to flag. Swept the whole `test/` tree myself for the same `importorskip("cupy")`/`skipif(cp is None)` pattern to check for a missed site — everything else either already had a real `getDeviceCount` check, never actually touches a device (pure-math or fully-mocked cupy), or is parametrized through the shared `gpu` fixture in `utils.py` and gets fixed for free by this PR's change there. `_gpu_count()`'s broad `except Exception` around `getDeviceCount()` correctly covers both "cupy absent" and "cupy present but broken driver" without swallowing anything it shouldn't.

Good to merge.

TomaSusi (approved):

PR #408, opened 2026-09-15
Refuse a built transition potential whose grid disagrees with the S-matrix

### LINKED ISSUE #400: PRISM-EELS: a grid-mismatched built TransitionPotential completes silently instead of raising
Noticed while fixing the `GridUndefinedError` build/match ordering bug in #386: a **built** `TransitionPotential` whose grid disagrees with the S-matrix's grid completes silently, both before and after that fix, rather than raising.

`_prism_eels_common_setup` now matches an *unbuilt* transition potential's grid before building it (the #386 fix), which correctly makes an unbuilt input agree with what the regular multislice driver produces. But if a caller hands in an *already-built* `TransitionPotential` (or a `TransitionPotentialArray`) whose grid doesn't match the S-matrix, there's no check catching the mismatch — the scan just runs and produces a result on the wrong grid, with no error or warning.

A crash is usually guarding something real; the concern here is the opposite — the absence of one. Worth adding an explicit grid-compatibility check (or at least a validation the way `_validate_potential_ensemble_indices` and similar guards elsewhere in this codebase have been added after review) so a grid mismatch fails loudly instead of silently producing a wrong result.

---

Original discussion: [#386](https://github.com/abTEM/abTEM/pull/386#issuecomment-5666438289)

pzeiger:
**Stacked on #386.** Review that one first; the diff shown against `dev` includes it. Only the last commit is new here.

Closes #400.

## What

`_prism_eels_common_setup` matches the transition potential's grid to the S-matrix waves. For an **unbuilt** potential that is #386's fix: it has no array yet, so the match sets the grid and `build()` then evaluates the form factors at the right gpts.

An **already-built** one cannot be matched. Its array shape is fixed, so the match overwrites the *grid* to agree with the waves and leaves the array alone:

```
built TP: gpts (32, 32)  array (4, 32, 32)
after _task_local(match_to=64²): grid gpts (64, 64)  array (4, 32, 32)
```

The object now lies about its own contents, and the scan completes on the wrong grid:

```
unbuilt  @64 (reference)   shape (12,12,95,1)  sum 9.085165e-07
prebuilt @64 (matched)     shape (12,12,95,1)  sum 9.085165e-07
prebuilt @32 (MISMATCH)    shape (12,12,95,1)  sum 7.437427e-07   <- 18.1 % low
```

**The output shape is identical in all three**, so nothing downstream can detect it, and the only console output is an unrelated `Overspecified grid` warning that the correct case emits too.

## The guard must precede the match — that is the whole fix

Checking after the match is useless: by then the grid reports the waves' gpts and `check_match` passes over a wrong-sized array. **The match is what destroys the evidence.** A mutation moving the guard below the match fails the tests exactly as removing it does, so the placement is pinned rather than merely commented.

## Why `RuntimeError`, and why here

`Grid.check_match` is the existing in-house guard for this — `iam.py:1948` already pairs it with `accelerator.check_match` for a potential against its waves. It compares extent as well as gpts, so a mismatched extent, equally un-matchable for a built array, is refused for free.

**This makes PRISM agree with its sibling rather than inventing an exception.** The multislice driver already refuses the same input, because `Waves.scatter` matches and then fails on the array:

```
MULTISLICE prebuilt @32   ValueError: operands could not be broadcast together
                          with shapes (1,4,32,32) (1,1,...
```

So today the same mismatched potential raises in multislice and returns an 18.1 %-wrong answer in PRISM. The new `RuntimeError` also names both grids, which that broadcast error does not.

## Behaviour change

Narrow, and only in the direction of loudness. A mismatched **built** potential goes from a silently wrong result to an error. A mismatched **unbuilt** potential is unaffected — still matched and built at the S-matrix's gpts, pinned by a regression test.

## Tests

`TestPrismEelsBuiltTransitionPotentialGrid` in `test/test_core_loss_fixes.py`: a built potential on the wrong gpts is refused; one on the right grid still runs (the guard must not fire on the case it is meant to allow); a mismatched extent is refused too; and an unbuilt potential is still re-gridded.

Removing the guard fails 2 of the 4. Moving it below the match fails the same 2 — which is the point. Full suite: 1948 passed, 0 failed, 318 skipped.


TomaSusi (changes requested):
Good fix for what it checks — verified the `isinstance` classification is exhaustive (only two `BaseTransitionPotential` subclasses exist), `Grid.check_match` correctly covers both gpts and extent (with float-roundoff tolerance on extent), the guard doesn't fire on the matching case, and the "placement is the whole fix" claim holds empirically (moving the guard after the match fails the same 2/4 tests as removing it).

But it's incomplete relative to the pattern it says it mirrors. `iam.py:1948` pairs `accelerator.check_match(waves)` with `grid.check_match(waves)` — this PR only added the grid half. `TransitionPotential.build()` bakes `energy` into the array's form factors (`k0`, `kn`, relativistic mass correction, interaction parameter all derive from it), so a built `TransitionPotentialArray` whose *grid* matches the S-matrix but whose *energy* doesn't is exactly as stale as a gpts/extent mismatch — `_task_local`'s `accelerator.match()` silently overwrites `.energy` to agree with the S-matrix while the array's baked-in form factors stay computed at the old energy.

Reproduced directly against this branch: built a `TransitionPotentialArray` at the S-matrix's exact grid but `energy=200e3` against an S-matrix at `energy=100e3` — `transition_potential_scan(...)` completes with no error, plausible-shaped result. Same silent-wrong-result failure mode #400 was filed for, just on the energy axis instead of the spatial one.

Suggested fix: add `transition_potential.accelerator.check_match(s_waves)` alongside the existing grid check, plus a fifth test for a built-but-energy-mismatched potential.

TomaSusi:
Confirmed — `accelerator.check_match(s_waves)` added right alongside the grid check, exactly matching `iam.py:1948`'s pattern, with a test (`test_a_mismatched_energy_is_refused_too`) reproducing the scenario I described. Good to merge from here.

pzeiger:

`_prism_eels_common_setup` now pairs `accelerator.check_match` with the existing `grid.check_match`, in the same place, for the same reason: `build()` bakes `self.energy` into the array's form factors — `k0`, `kn`, the relativistic mass correction, the interaction parameter — so a built array on the right grid but the wrong energy is exactly as stale as a gpts/extent mismatch, and `_task_local`'s match would otherwise silently overwrite `.energy` to agree with the S-matrix while those baked-in form factors stay computed at the old one.

`test_a_mismatched_energy_is_refused_too` asserts the `RuntimeError` directly rather than a value difference — the existing `_synthetic_transition_potential` fixture has no energy dependence, so comparing arrays would pass for the wrong reason. Placement matters here exactly as it did for the grid check: moving `accelerator.check_match` after `_task_local`'s match fails the new test, because the match is what destroys the evidence.

This broke two tests in `test_transition_potential_transport.py` (`tp-graph-nodes`, #386's own test file, inherited as this branch's base) — both built their fixtures around a mismatched, already-built transition potential completing successfully, which is exactly the case this guard now refuses. `test_prism_lazy_and_eager_agree` now builds its own energy-matched transition potential rather than reusing the shared fixture, whose mismatch exists to exercise the *multislice* driver's private-view matching, a concern that test never had. The concurrency test, renamed `test_prism_threaded_and_synchronous_schedulers_refuse_a_mismatch_alike`, now asserts what still needs guarding once the case is refused instead of corrected: every concurrent task hits the same `RuntimeError`, under both threaded and synchronous scheduling, and a refused scan still leaves the caller's object untouched.

One gap surfaced by that: the same pattern is unguarded on the multislice path (`abtem/multislice.py`, immediately before its own `_task_local` call) — a built, energy-mismatched transition potential runs `Probe.transition_potential_scan` to completion with no error. Out of scope here since it's a different driver entirely; recorded for its own fix.

`test_transition_potential_transport.py` (12), `test_core_loss_fixes.py` (39), `test_prism.py`, `test_waves.py` all pass; CI is green on 3.11–3.14.

TomaSusi:
Thanks for the thorough write-up and for catching the two dependent-test fixes.

On the multislice-side gap you flagged: since it's the identical defect (a built, energy-mismatched transition potential completing silently) with the identical fix (the same `accelerator.check_match` you just added here, placed before `_task_local`'s match for the same reason), I'd rather see it land in this PR than get filed separately — it's the same guard on the sibling driver, not a new investigation. Could you add it to `abtem/multislice.py` alongside its own `_task_local` call, with a matching test? Happy to review as soon as it's up.

pzeiger:
`transition_potential_multislice_and_detect` (`abtem/multislice.py`) now pairs `grid.check_match` with `accelerator.check_match` before its own `_task_local` call, mirroring `_prism_eels_common_setup` exactly — same guard, same reason, same place. Swept the codebase for every call site of `_task_local(match_to=...)`: there are exactly two, both now guarded.

The grid half isn't purely redundant here: a wrong-shaped array already raised inside `Waves.scatter`'s broadcast, but as an opaque `ValueError` that never named what disagreed. It now gets the same clear `RuntimeError` the PRISM path already gives.

`TestMultisliceBuiltTransitionPotentialEnergy` in `test/test_core_loss_fixes.py` mirrors `TestPrismEelsBuiltTransitionPotentialGrid`'s energy test directly, for side-by-side comparison. `test_transition_potential_transport.py`'s shared `_setup()` fixture was deliberately energy-mismatched (80 keV against a 60 keV probe) specifically to exercise this driver's private-view matching — exactly the case now refused rather than silently corrected, so it's matched instead. The two tests whose entire point was that mismatch (renamed `test_scan_refuses_a_mismatch_rather_than_rematching_in_place` and `test_scan_leaves_the_users_object_unmutated_even_when_refused`) now assert the new invariant: refused with a `RuntimeError`, under concurrent threaded execution, with the caller's object still untouched. The other six tests sharing that fixture — graph structure, lazy/eager and scheduler agreement, distributed-cluster survival — never depended on the mismatch and are unaffected by matching it.

Removing the new guard fails exactly the three tests built to catch it, and no others. `test_transition_potential_transport.py` (12), `test_core_loss_fixes.py` (40), `test_prism.py`, `test_waves.py`, `test_multislice_input.py` all pass; CI green on 3.11–3.14.

TomaSusi (approved):
Confirmed — the multislice-side guard mirrors the PRISM one exactly, and I checked there are only two `_task_local(match_to=...)` call sites in the whole codebase, both now guarded. Good to merge.

pzeiger:
This branch needed a rebase: `tp-graph-nodes` (#386) absorbed this branch's first commit ("Match the grid before building in the PRISM-EELS setup") into its own history under a new hash, so the automatic stack rebase collided trying to reapply an already-applied patch.

Rebased onto the current `tp-graph-nodes` tip, skipping the now-redundant commit and replaying only the four genuinely new ones. Two conflicts landed in `test/test_core_loss_fixes.py`, both from test classes inserted at the same point by sibling branches; resolved by computing the expected class union explicitly rather than trusting a line-based merge, and verified by parsing the result and diffing it against that computed union.

The two guard functions this branch actually changes — `_prism_eels_common_setup` and `transition_potential_multislice_and_detect` — are byte-identical before and after the rebase; the rebase only replayed them onto a tree carrying `tp-graph-nodes`'s own further, unrelated work (a detectors-elastic refusal and an exit-plane indexing fix in the same functions). Full suite reconfirmed: `test_core_loss_fixes.py` (97), `test_transition_potential_transport.py` (12), `test_prism.py`, `test_waves.py`, `test_multislice_input.py`. Removing all three guards at once — PRISM grid, PRISM energy, multislice energy — fails exactly the four tests built to catch them, and no others. CI green on 3.11–3.14.

TomaSusi:
Verified the rebase directly rather than just trusting the description: diffed `_prism_eels_common_setup` (core_loss.py) and the guard section of `transition_potential_multislice_and_detect` (multislice.py) line-for-line against what I approved before — both byte-identical, just shifted by the inherited commits around them. CI's green across 3.11-3.14. Still good to merge.

PR #410, opened 2026-09-15
Stop Potential._prepare_atoms writing into the atoms it was given

pzeiger:
**Stacked on #390.** Review that one first; the diff shown against `dev` includes it. Only the last commit is new here.

## What

`Potential._prepare_atoms` wrapped its atoms in place. For `DummyFrozenPhonons` — the wrapper every plain `Potential(atoms)` gets — `get_transformed_atoms()` and `randomize()` are both the identity, so the write landed on the object the potential stores and ships into the task graph as **one shared node**. Every task on a worker then wrapped the same `Atoms`.

#390 left this deliberately, with `copy=False` and a comment recording the aliasing as a separate defect. This is that defect.

## It reaches the caller's own objects, not just abTEM's copy

`_validate_frozen_phonons` copies a plain `Atoms`, but passes a **list** (which becomes an `AtomsEnsemble` holding references) and a **pre-built frozen-phonons object** straight through. Measured, two atoms in a 4 Å cell with one at x = 4.2:

```
Potential(atoms)                   caller safe      own atoms rewritten
Potential([a])                     CALLER REWRITTEN
Potential(DummyFrozenPhonons(a))   CALLER REWRITTEN
Potential(FrozenPhonons(...))      safe
```

`FrozenPhonons` is the oracle: its `randomize` already copies, which is why that path never had the defect, and the fix follows it rather than inventing a convention.

## The fix

One argument — `wrap_and_snap_atoms(atoms)` rather than `copy=False`.

**This layer and not a neighbouring one**, because it is the only writer in the mechanism. `get_transformed_atoms` has five consumers and only this one writes, so copying there would charge four read-only callers for a hazard they do not create; `randomize` copies unconditionally where this copies only when it must. `wrap_and_snap_atoms` already takes ownership as a parameter, so the fix is answering it correctly rather than adding machinery.

## Results and cost

**Bit-identical**, checked against the base across seven configurations — Si cubic, an out-of-cell atom, graphene and GaAs×2, both `periodic` settings where each is buildable, and a three-configuration `FrozenPhonons` ensemble. Every hash matches.

Cost is one `Atoms` copy per build task where wrapping is needed: 0.08 ms against 0.06 ms for 13 824 atoms, and nothing at all for a structure already inside its cell, since the copy is conditional.

## What this does not fix

The torn-read race the shared node makes possible under the threaded scheduler is removed *at this call site*. An automated sweep reported reproducing corruption more broadly and **I have not reproduced that myself**, so the claim here is confined to what is measured above.

## Tests

`TestPotentialDoesNotMutateItsAtoms` in `test/test_potentials.py`: all three entry points, the `FrozenPhonons` oracle, the cross-object ordering case, and that the wrap still reaches the slicing so copying cannot silently lose it.

**Two of these were rewritten after an adversarial review found they could not fail**, which is worth stating because both had been presented as the strongest evidence for the fix:

- *"building twice gives the same result"* asserted a consequence that does not occur. `get_sliced_atoms` memoises `_sliced_atoms`, so a second build never re-enters `_prepare_atoms`, and `wrap_and_snap_atoms` is idempotent regardless. It passed on the base. Replaced by the case that does discriminate — a **second potential** built from the same `Atoms` object, which on the unfixed code sees atoms the first build already wrapped and slices them differently.
- the `FrozenPhonons` oracle used `sigmas=0.0`, which makes an in-place randomize invisible: the displacement is `positions += 0 * r`. Deleting `randomize`'s own copy left it green. At `sigmas=0.1` it catches that, which is what "pinned so the fix cannot be simplified by moving the copy there" was supposed to mean.

Restoring `copy=False` now fails 4 of the 6. Deleting the copy in `FrozenPhonons.randomize` — a different mutant, on the path this fix treats as its oracle — fails 1. Full suite: 1966 passed, 0 failed, 318 skipped.


TomaSusi (changes requested):
Everything actually in the diff is correct — traced `_prepare_atoms` end to end, confirmed `wrap_and_snap_atoms` never mutates its input (copies when it needs to wrap, no-ops read-only when already in-cell), and confirmed every downstream step (`pad_atoms`, `SliceIndexedAtoms.__init__`'s own wrap) copies unconditionally, so nothing after the fixed call site can reach the caller's object.

But there's a second, unguarded writer this doesn't touch, and it undermines the PR's central claim. `get_transformed_atoms()` — the very *first* line of `_prepare_atoms`, running before the newly-fixed `wrap_and_snap_atoms` call — calls `orthogonalize_cell()` whenever the cell isn't already orthogonal and box-matching. `orthogonalize_cell` mutates `atoms` in place with no copy: `cell = atoms.cell; cell[...] = 0.0; atoms.set_cell(cell); atoms.wrap()`, straight on the object it's handed.

Reproduced with a graphene cell (non-orthogonal — the common case for 2D materials, and for anything not axis-aligned): `Potential([atoms])` and `Potential(DummyFrozenPhonons(atoms))` still mutate the caller's original position by a full lattice vector on this branch. More importantly: **`Potential(FrozenPhonons(atoms, num_configs=2, sigmas=0.1, seed=1))` mutates the caller's object too** — directly contradicting "`Potential(FrozenPhonons(...))` ... was always safe because `FrozenPhonons.randomize` already copies." That's only true when `get_transformed_atoms` takes the identity path; here `orthogonalize_cell` writes into `FrozenPhonons._atoms` (itself uncopied at construction) before `randomize` ever gets a chance to copy.

The test fixture (`TestPotentialDoesNotMutateItsAtoms._atoms()`) hardcodes an orthogonal `cell=(4,4,4)` with the default `plane="xy"`, so none of the 6 tests — including the one explicitly commented as pinning the `FrozenPhonons` oracle — ever reach `orthogonalize_cell`, which is how this got through.

Given the PR title's claim ("stop `_prepare_atoms` writing into the atoms it was given"), I think this needs the same treatment applied to `orthogonalize_cell` (or its call site) before merge, plus a non-orthogonal-cell case added to the test fixture so it can't regress silently again.

TomaSusi:
Confirmed, and this went further than what I flagged — `orthogonalize_cell` now copies once at the top rather than patching just the call site I found, which correctly also covers the two other in-place ASE calls inside that function (the `origin`-translate branch and the snap-before-cut branch) I hadn't traced. And catching the second, independent `orthogonalize_cell` call site in `Potential.__init__` via `sampling="auto"` — reached before `build()` is ever called — is a real find; I only looked at the `get_transformed_atoms()` path. Good to merge from here.

pzeiger:

`orthogonalize_cell` copies its argument once, at entry, rather than at one of its three internal mutation points — `set_cell`/`wrap` unconditionally, `translate`/`wrap` for a non-default `origin`, and `_snap_scaled_positions_to_cell_boundary` right before `cut()` in the repeat-and-cut branch. A fix at any single call site would have left the others live; every later `atoms = ...` reassignment in the function now operates on that private copy, and the parameter is never touched again after the first line.

`TestPotentialDoesNotMutateItsAtoms`'s fixture hardcodes an orthogonal, box-matching cell, so none of its tests take the `orthogonalize_cell` branch at all — only the `wrap_and_snap_atoms` one this PR already fixed. `TestNonOrthogonalCellDoesNotMutateItsAtoms` adds a sheared cell with an atom placed outside `[0, 1)` in fractional coordinates along the sheared vector, so wrapping moves it by a whole lattice vector and an in-place write is unambiguous. Parametrised over the same three aliasing constructions as the existing class — a list, `DummyFrozenPhonons`, `FrozenPhonons` — since a plain `Potential(atoms)` is already copied earlier, before `orthogonalize_cell` is ever reached.

Adversarial review of that fix found one real gap, not a live defect: `orthogonalize_cell` has a second call site, in `Potential.__init__` itself, reached through `sampling="auto"` when the cell needs a transform — running synchronously in the constructor, before `build()` is ever called. The single copy at entry already covers it, since both call sites go through the same function, but nothing pinned it: every test in the new class builds before checking, which only exercises the build-time call site. `test_sampling_auto_does_not_rewrite_the_callers_atoms_either` mirrors the existing parametrized test without calling `build()` at all, closing that gap.

Removing the copy fails all six tests in the class — the original three plus the new one, three constructions each — and only those six; `test_potentials` (73) and `test_atomic_potential` pass otherwise.

TomaSusi (approved):
The `orthogonalize_cell` gap is fixed at the root, plus the second call site I'd missed — good to merge.

PR #412, opened 2026-09-16
Make Potential equality answer the question it is asked

pzeiger:
**Stacked on #393**, which this targets rather than `dev`, so the diff here is the single commit on top of it.

The real dependency is #392, two below: without its `_DeviceArrayCache.__eq__`, `_integrator` leaks into the comparison as well, and two of the tests added here fail for that second reason.

## What

`safe_equality` is wrong in **two independent directions at once**, and the suite cannot see either, because its only equality test compares an object with itself:

```python
def test_copy_equals(data, copyable):
    original = data.draw(copyable())
    assert original.copy() == original      # cannot fail for an over-permissive __eq__
```

### False positive

The loop ends in `if equal is False: return False` — an identity test against the `False` singleton. `ase.Atoms.__eq__` returns `numpy.bool_`, and `np.False_ is False` is `False`, so a genuine difference is silently dropped:

```
Potential(atoms) == Potential(atoms with one atom moved 1.234 Å)   ->   True
```

Different chemical species happen to be caught, but only because `_atomic_numbers` is an ndarray `np.allclose` can handle.

Now `np.all(equal)`, which treats `bool`, `numpy.bool_` and array alike — the array case matters too, since a `==` that `np.allclose` then fails to reduce leaves an array here.

### False negative

`_sliced_atoms` is populated lazily by `get_sliced_atoms()`, so a **built** `Potential` stops comparing equal to an identical **unbuilt** one: equality depends on whether a result has been computed from the object. It is derived state, so it carries no identity of its own.

`EqualityMixin` now honours an `_eq_exclude` class attribute, which reaches `safe_equality`'s existing `exclude` parameter — until now only `OrdinalAxis.concatenate` passed one. Same reasoning as `_DeviceArrayCache.__eq__` in #392, one level up.

Two things make that exclusion reach the cases that occur:

- **It is declared where the attribute is created** — on `_FieldBuilderFromAtoms`, so `Potential`, `MagneticField` and `VectorPotential` all inherit it, and separately on `CrystalPotential`, which builds its own cache and descends from `_PotentialBuilder` instead. Declared on `Potential` alone it would not reach either sibling.
- **`safe_equality` forwards it through its own recursion.** That recursion bypasses the nested object's `__eq__`, which is what would otherwise apply `_eq_exclude`, so without forwarding the invariant holds only for a top-level operand and `SMatrix(built potential)` still compares unequal to `SMatrix(unbuilt potential)`. The core-loss driver calls `get_sliced_atoms()` on the caller's potential to extract EELS sites, so a potential handed to one and then compared is the ordinary case, not a corner.

**The two directions have to ship together.** Fixing the false negative alone makes `==` look authoritative while it still reports physically different potentials as identical — the negative was partly masking the positive.

## A third, in the same function

```python
if isinstance(value, EmptyEnsemble) and isinstance(b.__dict__[key], EmptyEnsemble):
    return True          # ← for the WHOLE object
```

Two `EmptyEnsemble`s make *that attribute* equal, not the object; the return skips every attribute declared after it. Now `continue`.

**Latent**: nothing in `abtem/` instantiates `EmptyEnsemble` or its subclass `EmptyTransform`, and no live object carries one. It is demonstrated on a local class in the test, which is also the only way to pin it against reintroduction.

## Deferred values are left exactly as they were

`np.all` on a dask value returns another dask value, and the `not` in front of it calls `__bool__`, which executes the graph. Unguarded, this change would make `==` on two lazy `Waves` run both multislice simulations, discard the results (the operands stay lazy) and let any error raised mid-graph escape `==`. On a 2×2×8 Si multislice at 256×256 gpts, warm: 0 dask computes before and 1 after, the comparison costing 0.25 s against 0.12 s to compute one operand outright — about twice, since it runs both operands' graphs.

A dask value is therefore skipped, which is what the old identity test did anyway: such a value could never be the `False` singleton. That leaves lazy arrays comparing equal regardless of content, which is wrong, and a larger class of false positive than the `ase.Atoms` one above. It is pre-existing, and what `==` should mean for a lazy object is a question about the whole library rather than about `Potential`, so it is raised separately rather than settled here.

The exclusion is blunt, and a comment next to it says so. `get_sliced_atoms()` returns the cache rather than a copy, so a caller who mutates what it returned changes what the potential builds while `==` still reports equal.

What that costs depends on the projection, because the two cache classes differ. For `projection="infinite"` the cache is a `SliceIndexedAtoms`, whose `_slice_index` is a `list` of arrays `safe_equality` cannot compare — `==` on it raises `ValueError`, which becomes "unequal" — so that comparison is constant-`False` and excluding it gives up nothing. For `projection="finite"` the cache is a `SlicedAtoms`, which has no `_slice_index` and compares correctly; there the exclusion does give up a real check that the `np.all` fix would otherwise have made effective.

Neither path regresses against `dev`, which misses that mutation on both:

| `p == q`, two identical potentials | `dev` | this PR |
|---|---|---|
| `infinite`, built vs unbuilt | False ✗ | True ✓ |
| `infinite`, built vs built | False ✗ | True ✓ |
| `infinite`, one mutated through the alias | False ✓ (constant-`False`) | True ✗ |
| `finite`, built vs unbuilt | False ✗ | True ✓ |
| `finite`, built vs built | True ✓ | True ✓ |
| `finite`, one mutated through the alias | True ✗ | True ✗ |

Comparing derived state only when both operands have it would get every row right, including the two mutation rows. That is the same rule the lazy question needs, so both go in the follow-up issue together rather than here.

## Blast radius

This function backs `==` for every `EqualityMixin` in the tree, so that is the point rather than a footnote.

Swept across `Potential`, `CrystalPotential`, `SliceIndexedAtoms`, `Probe`, `PlaneWave`, `SMatrix`, `FrozenPhonons`, `GridScan`, `CTF`, `AnnularDetector`, `MagneticField` and `VectorPotential`, three still cannot compare equal to an identical instance of themselves, all for pre-existing reasons outside this diff:

- `SliceIndexedAtoms`, for the `_slice_index` reason above.
- `MagneticField` and `VectorPotential`, because `_integrator` is a `QuasiDipoleProjections` subclass, which is not an `EqualityMixin` — so `==` falls back to object identity and returns the `False` singleton, which the old test caught just as the new one does. They do now inherit the exclusion, but it changes nothing for them until that is fixed.

All three go in the follow-up issue.

Equality still discriminates on gpts, slice thickness and species, and identical potentials still compare equal.

## Tests

`TestEqualityDiscriminates` in `test/test_copy.py` compares objects that genuinely differ, which the existing self-comparison never did.

Six mutations were run against it, and each fails a test:

| mutation | test that fails |
|---|---|
| restore `equal is False` | `test_potentials_with_different_atomic_positions_are_not_equal` |
| restore the `EmptyEnsemble` `return True` | `test_an_empty_ensemble_attribute_does_not_short_circuit_the_rest` |
| empty `_eq_exclude` on `_FieldBuilderFromAtoms` | `test_a_built_potential_equals_an_identical_unbuilt_one` **and** `..._inside_a_container` |
| drop the forwarding in the recursion | `test_a_built_potential_equals_an_unbuilt_one_inside_a_container` |
| drop `_eq_exclude` on `CrystalPotential` | `test_a_built_crystal_potential_equals_an_identical_unbuilt_one` |
| drop the deferred-value guard | `test_equality_does_not_execute_a_dask_graph` |

Five fail exactly one test; emptying the base `_eq_exclude` fails both potential-exclusion tests, since the container case rests on the same declaration. So every fix has its own guard rather than one test covering several by accident.

The `CrystalPotential` test uses `get_sliced_atoms()` rather than `build()`, because `build()` populates the *unit* potential's cache: a test written against it passes through the forwarding fix and never exercises the declaration it is meant to pin.


TomaSusi (changes requested):
All three documented bugs and their fixes are correct — independently verified each: `np.all(equal)` correctly handles the `numpy.bool_ is False` false-positive (confirmed `np.False_ is False` really is `False` in Python, and confirmed the empty-array vacuous-truth edge case never actually reaches `np.all` since a shape mismatch raises and is caught first); the `_eq_exclude` recursion-forwarding fix does reach nested comparisons (`SMatrix(built) == SMatrix(unbuilt)` now correctly compares equal); and the `EmptyEnsemble` `return True` → `continue` fix is right. The "deliberately not fixed" claims (mutated-through-alias false positive, `SliceIndexedAtoms`/`MagneticField`/`VectorPotential` still not self-equal) all check out against the code as described.

One gap: the PR frames itself as a sweep for lazily-populated cache attributes needing `_eq_exclude`, listing a dozen classes it checked — but `TransitionPotentialArray._local_potential_device_cache` (`abtem/inelastic/core_loss.py:914`) has the identical pattern (populated lazily by `_local_potential_on_device()`, a normal side effect of core-loss/EELS multislice) and has no `_eq_exclude`. Confirmed no such declaration exists on that class. Same bug class the PR fixes elsewhere, just missed in the sweep — cheap to close (`_eq_exclude = ("_local_potential_device_cache",)` on the right class).

Also worth a note, lower severity, pre-existing and not introduced here: nested `EqualityMixin` comparisons get computed twice per level (once implicitly at the `==`, again explicitly for the exclude-forwarding), so comparison cost grows exponentially with nesting depth — not blocking, but worth a follow-up issue since this PR is the one now maintaining that function.

TomaSusi:
Confirmed — `_eq_exclude = ("_local_potential_device_cache",)` on `TransitionPotentialArray`, same pattern as `Potential._sliced_atoms`. Good to merge from here.

pzeiger:

The sweep missed one class with the identical pattern: `TransitionPotentialArray._local_potential_device_cache` (`abtem/inelastic/core_loss.py`), populated lazily by `_local_potential_on_device()` — a normal side effect of core-loss multislice, the same shape of derived state `_sliced_atoms` has for `Potential`. `__getstate__` already drops it for pickling, for the same reason it doesn't belong in equality: it's a per-process convenience, not part of the object's identity. `_eq_exclude` now says the same thing to `safe_equality`.

`test_transition_potential_array_ignores_its_device_cache` pins it in `test/test_copy.py`, next to the other `_eq_exclude` tests: two transition potentials built from the same array, one used on a device and one not, compare equal. Emptying the new declaration fails exactly this test and no other.

Separately raised as its own follow-up: nested `EqualityMixin` comparisons are computed twice per level, once implicitly at `==` and again explicitly for the exclude-forwarding, so cost grows with nesting depth. Pre-existing, not introduced here, and worth its own issue rather than folding into this one.

`test_copy` (30), `test_core_loss_fixes` and `test_potentials` all pass.

TomaSusi (approved):
The `_eq_exclude` gap is fixed and confirmed — good to merge.

PR #414, opened 2026-09-16
Add experimental Metal (MPS) backend for Apple silicon

TomaSusi:
## Summary

abTEM's docs have long advertised [experimental Metal support](https://abtem.github.io/doc/dev/getting_started/install.html#metal-on-apple-silicon-experimental), but the implementation only ever existed on the stale `metal` branch (May 2024, 547 commits behind `dev`). The later `torch` branch started a larger Array-API rewrite and was also abandoned (Jan 2026, 307 commits behind). Neither rebases cleanly, so this ports the `enable_mps` approach forward against current `dev`.

`mps` becomes a device alongside `cpu` and `gpu`. New `abtem/core/_torch.py` adapts PyTorch's Metal tensors to the numpy-like array-module interface abTEM dispatches on — the role CuPy plays for CUDA.

```python
# enable_mps must be set before abTEM is imported (see Known limitations)
potential = abtem.Potential(atoms, gpts=512, device="mps")
probe = abtem.Probe(energy=200e3, semiangle_cutoff=20, device="mps")
measurement = probe.scan(potential, scan=scan, detectors=detector)
```

## Results

**Failures on Metal: 435 → 0 of 2478. Process aborts: several per run → 0.**

Now merged up to current `dev`: its 185 new tests pass on Metal as they stand, and its new per-device caches have been taught about Metal (see below).

Correctness against the CPU path, at float32 tolerances:

| quantity | max relative difference |
|---|---|
| potential | 7.9e-07 |
| exit wave | 5.1e-05 (overlap 0.9999996) |
| HAADF / flexible annular | 4.9e-05 / 8.3e-05 |

Performance on an M2 (8-core, 16 GB), FFT-friendly grids:

| workload | CPU 1-thread | CPU 4-thread | MPS | vs 4-thread |
|---|---|---|---|---|
| PlaneWave multislice, 4x4x8 cell | 0.69s | 0.31s | 0.18s | 1.7x |
| STEM 3x3x6, 256² grid, 64 probes | 5.13s | 1.70s | 0.21s | **8.0x** |
| STEM 4x4x8, 512² grid, 64 probes | 40.3s | 13.7s | 0.97s | **14.1x** |
| PRISM upsampled build, 512² grid | 1.71s | — | 0.37s | **4.6x** |

Metal degrades badly on FFT sizes with prime factors above 7 — a 272² grid (2⁴×17) gave no speedup at all.

## Four constraints that shaped the design

Each found empirically and documented at the code that depends on it.

**1. PyTorch must be imported before pyfftw.** Both ship their own `libomp.dylib`, and in a process that loaded pyfftw's first, ordinary torch operations *segfault* (exit 139, no traceback). The torch import is therefore eager and gated on `enable_mps` at `abtem.core.backend` import time — a later `config.set` cannot retroactively fix library load order.

**2. PyTorch's MPS backend is not thread-safe.** It keeps a process-wide Metal shader-library cache whose hash table two threads corrupt, surfacing as a process spinning forever inside `MetalShaderLibrary::exec_unary_kernel` or as `Fatal Python error: Aborted`. Being a race it struck intermittently, wherever a collision landed, so it could not be chased call site by call site. Entry into torch is serialized behind a reentrant lock in the backend, which also covers callers who thread on their own account.

**3. NumPy has two dispatch protocols, and both are needed.** `__array_function__` keeps dask's `np.concatenate` from silently copying results to host; `__array_ufunc__` is what operators resolve to, without which `host_array @ device_array` fails outright. Unregistered entries return `NotImplemented` so NumPy raises rather than quietly falling back to the host — the behaviour CuPy has. Turning the silent fallback off immediately exposed four functions that had been copying arrays back and forth unnoticed.

**4. Metal is single precision.** Selecting float64 with the Metal device is now refused where it is set, and a refused configuration rolls back rather than leaving half of itself applied. `cpu` and `gpu` are untouched — both support double precision.

## Fixes that also apply to CUDA

- **`Waves.tile(renormalize=True)` silently doubled the precision of its result.** It divided by `xp.asarray(np.prod(repetitions))`; under NEP 50 both a NumPy scalar and a 0-d array participate in promotion, so a `complex64` wave came back `complex128` — twice the memory, against the configured precision — on CPU and CUDA alike. Metal only exposed it by being unable to perform the mixed operation at all.

- **Per-device cache keys exclude any non-CuPy accelerator.** `integrals.py` (three sites) and `core_loss.py` key their caches on the concrete device an array lives on, spelled as `int(like.device.id)` / `with cp.cuda.Device(...)` / an unconditional `import cupy` for any non-CPU device. A torch device has neither an id nor a context, so all four raised on Metal. This is a recurring shape rather than a one-off, and would be worth a backend helper naming the current device generically.

- Two computations called `.compute()` on a **bare dask array** rather than the `ArrayObject` around it (`s_matrix.py`, `iam.py`), bypassing scheduler resolution. `_resolve_gpu_scheduler`'s own docstring says the threaded scheduler cannot be used with CuPy — Metal just fails loudly enough to expose it.
- `fft_shift_kernel` forced positions to NumPy whenever CuPy was absent, collapsing the entire probe onto the host.
- zarr writes and JSON metadata now bring device arrays to the host rather than handing zarr something it cannot encode.

## Where the host is used deliberately

Two operations genuinely cannot run on Metal, and fall back to the host rather than silently losing accuracy or becoming unavailable:

- **The PRISM Gram matrix**, accumulated in double precision on purpose (its eigenvalues are the *squares* of the singular values, so float32 would resolve the spectrum only to ~1e-3). Measured cost: **5.6%** of a 256² build, **8.8%** of a 512² build — against which the build still runs 3.4–4.6x faster than CPU.
- **`scipy.ndimage`** (`map_coordinates`, `gaussian_filter`), on measurement-space arrays far smaller than the wave functions they came from.

## Test coverage

The device-parametrized tests only ever ran their accelerator half on CuPy, so on Apple silicon the whole `["cpu", gpu]` surface skipped. The `gpu` parameter now resolves to whichever accelerator the machine has — Metal here, CUDA elsewhere — with **no change at the 121 parametrization sites**.

## Dependencies

torch only (on macOS it pulls just `networkx` + `setuptools`; the CUDA/triton deps are all `platform_system == "Linux"`). An opt-in extra, excluded from `all` and restricted to macOS/arm64 so `pip install abtem[mps]` on Linux cannot drag in a multi-GB CUDA stack for a device that cannot exist there.

## Known limitations

None outstanding in the test suite. Two areas run deliberately on the host, because Metal cannot do the arithmetic they need rather than because of any gap here:

- **The PRISM Gram matrix and probe-position grouping**, both documented in the code as requiring double precision (the Gram eigenvalues are squared singular values; the positions are grouped to 1e-4 pixels). Measured cost of the round trip: 5.6% of a 256² build, 8.8% of a 512², against which the build still runs 3.4–4.6× faster than the CPU.
- **`scipy.ndimage` / `scipy.signal`**, which have no Metal counterpart, on measurement-space arrays far smaller than the wave functions they came from.

Both raise rather than narrow silently if the configuration asks for something Metal cannot honour.

## Test plan

- [x] `test/test_mps.py` (11 tests): device round-trip, 0-d scalar shape, precision narrowing/rejection at both config and use time, unsupported-operation error, FFT round-trip, scatter-add, and potential / plane-wave / STEM agreement with CPU
- [x] Exhaustive NumPy-equivalence checks of the hand-written `pad` (constant/wrap/reflect, 1-D to 3-D, asymmetric and scalar widths), negative-step slicing across 8 forms, `squeeze`, `diff`, `linspace(endpoint=)`, `RandomState`
- [x] Full suite, MPS disabled: 1603 passed, 0 failures (unchanged from `dev`)
- [x] Full suite, MPS enabled: 2116 passed, 0 failed, 0 aborts
- [x] ruff/black clean on all changed files; mypy adds no new errors
- [ ] Verify on Apple silicon other than M2, and on a CUDA machine that the CuPy path is unaffected


TomaSusi:
Heads-up on a merge overlap with [abTEM/abTEM#426](https://github.com/abTEM/abTEM/pull/426): both this PR and #426 modify the same `gpu = pytest.param(...)` definition in `test/utils.py`.

- This PR replaces the old `cp is None` skip check with `_accelerator_device() is None`, so `gpu` resolves to `"mps"` on Apple Silicon, `"gpu"`/CUDA elsewhere.
- #426 (a test-suite dedup/cleanup pass) keeps the pre-existing `_gpu_count() < 1` check, and adds `devices`/`lazy_params` parametrize decorators plus a few other `test/utils.py` helpers that just reference the `gpu` variable.

These are compatible in intent, not a real conflict — whichever PR merges second will need to manually resolve the textual conflict on that one line, keeping this PR's `_accelerator_device()`-based resolution. #426's new `devices`/`lazy_params` decorators will automatically pick up MPS support once this PR's version of `gpu` is in place, with no further changes needed.

Verified this by running #426's branch on a CUDA GPU workstation (2508 passed, 7 skipped, 0 failed) and comparing against a durations log from this PR's sync branch on Apple Silicon (2403 passed, 84 skipped, 0 failed) — same underlying `gpu` sentinel mechanism, different resolved accelerator. Also noticed while comparing: `Images.gaussian_filter`'s CUDA path (`cupyx.scipy.ndimage.gaussian_filter`) is disproportionately slow relative to CPU and relative to `lorentzian_filter`'s FFT-based approach — this PR's Metal fallback to the CPU host for `scipy.ndimage` sidesteps that, which is a useful independent data point if it's ever worth revisiting the CUDA path.

EugenKozyrau:
I've confirmed GPU usage and ~3x speed-up in a 'large' crystal potential scenario on apple silicon.
I also encountered 2 bugs: One when invoking `build()` of a potential before forming the `CrystalPotential()` of it `[TypeError can't assign a Array to a torch.mps.FloatTensor]`;
and the second when trying to plot the phase of the probe wave function: `probe.build().phase().show()` `[TypeError: angle() takes 1 positional argument but 2 were given]`.

TomaSusi:
Thanks for testing — both reproduced and fixed in 68aa72d5.

The `CrystalPotential` one turned out not to be Metal-specific. A unit potential you build yourself is used as-is, and `build()` is lazy by default, so its array is a dask array rather than a device array. Metal raised on it; NumPy quietly accepted it and recomputed the entire unit potential once per slice, so this was costing CPU users too. Both slice generators now materialize the unit, as the other branch already did.

`phase()` was a signature mismatch: `numpy.angle(z, deg)` takes `deg` positionally and dask passes it that way, but `torch.angle` has no such argument. Fixed, along with `numpy.round(a, decimals)`, which had the same problem. Regression tests added for both.

EugenKozyrau:
Invoking the `build()` method of a potential that is then used as argument for a `CrystalPotential()` now doesn't break `.build().compute()` of said `CrystalPotential()`, but shows two instantaneously finished loading bars instead of one. Hopefully nothing is calculated twice?

A new bug as appeared, as calculating the probe's FWHM via `intensity.interpolate_line_at_position(center=(10, 10), angle=0, extent=10).width(height=0.5)` now raises a `TypeError: where() missing 2 required positional arguments: 'x' and 'y'`.

TomaSusi:
Nothing is calculated twice — the two bars are two genuinely separate computations, and that is the fix working rather than a symptom.

A unit potential you build yourself is used as-is, so it has to be materialized before it can be tiled. That is now one dask computation (its own bar), and the crystal build is the second. I measured it by tapping the unit's dask graph with a counter: for a 4-unit-slice unit at `repetitions=(2, 2, 4)` — 24 crystal slices — the unit is evaluated **1** time. Before the last commit it was evaluated **24** times, once per slice, and on every subsequent use of the same object as well. `compute()` also materializes the object in place, so a unit reused across several `CrystalPotential`s is built once in total. There is a test asserting the count now.

The `width()` failure is fixed in fda95310. `numpy.where(condition)` with no `x` and `y` is `numpy.nonzero(condition)`, and `_calculate_widths` reaches it that way when it scans for sign changes; `torch.where` has the same one-argument form but the wrapper declared all three arguments as required. The FWHM now agrees with the CPU to better than 1e-4 relative. Thanks for the sharp reports — please keep them coming.

PR #295, opened 2026-06-20
Feature: Fast plasmon and phonon phase-scramble scattering (Mendis 2023) for multislice/PRISM

TomaSusi:
## Summary

This PR follows B. G. Mendis, *Microsc. Microanal.* **29** (2023) 1111, "A 'Phase Scrambling' Algorithm for Parallel Multislice Simulation of Multiple Phonon and Plasmon Scattering Configurations", which covers two related but methodologically distinct threads:

1. **Frozen-phonon (thermal-diffuse) disorder** — `CrystalPotential(mixing="phase_scramble")`. This is the paper's own novel contribution (eq. 7c): a *coherent* combination of every frozen-phonon pool configuration's transmission function, with independent random relative phases at every sub-slice, all within a single multislice pass — an alternative TDS-averaging mode for small, reused frozen-phonon pools.
2. **Plasmon energy loss** — `PhaseScramblePlasmons`. For plasmons, the 2023 paper does not introduce an independent method: it explicitly cites and reuses the earlier plasmon-scattering model of B. G. Mendis, *Ultramicroscopy* **206** (2019) 112816 (and its 2020 corrigendum). `PhaseScramblePlasmons` implements that cited 2019 plasmon model (the per-slice Lorentzian angular scattering probability and random-order-Bessel tilted-beam operator), applied *inline* at the bottom of every slice so a single pass covers all plasmon-loss orders simultaneously, with statistical convergence via *incoherent* ensemble averaging over repetitions (reusing the potential's frozen-phonon configuration axis) — i.e. the same kind of ensemble average as ordinary TDS/frozen-phonon averaging, not the eq. 7c coherent combination used for phonons in Part 1.

So Part 1's coherent multi-configuration combination and Part 2's inline-but-ensemble-averaged plasmon operator are not literally the same mechanism — they're two different techniques from the same paper, sharing only the "apply a correction inline at the bottom of every slice" architecture. See the note near the end on why they're nonetheless kept in one PR (they do interact when composed on the same potential).

For contrast, `MonteCarloPlasmons` (already on `dev`, *not* part of this PR) implements the *2019* paper's original approach directly — sampling discrete scattering events and running a full separate multislice per sample, rather than an inline per-slice operator.

## Part 1 — Plasmon energy loss: `PhaseScramblePlasmons`

[...]

PR #415, opened 2026-09-16
Fix the FFTW plan cache in CachedFFTWConvolution

TomaSusi:
## The bug

`CachedFFTWConvolution` (`abtem/core/fft.py`) never hit its cache. `self._shape` was set to `None` in `__init__` and **never assigned anywhere else**, so the guard `if array.shape != self._shape` was always true and both pyfftw plans were rebuilt on every call — each allocating and zeroing a scratch array the size of the input.

This sits on the CPU multislice hot path: `FresnelPropagator.propagate` routes here whenever `config.get("fft") == "fftw"` (the default) and the wave array is numpy. The GPU and `numpy`/`mkl` FFT paths are untouched.

Not a recent regression: the `self._shape` bug traces to `6cb7735a` (2023-11-06, itself titled "fixed issue with cached FFTW when the shape of the transform is changed") — the commit meant to fix a different shape issue is the one that introduced this one. It has been live for close to three years; every later commit touching the file left the two lines untouched.

Measured on a 64-probe / 22-slice CPU STEM scan, both variants instrumented in the same process under identical settings:

| | dev | this PR |
|---|---|---|
| convolution calls | 44 | 44 |
| plans built | **92** | **4** |
| plans per call | 2.09 | 0.09 |

The 44 calls are 22 slices × 2 probe chunks; the scan sees two batch shapes, `(3, 512, 512)` and `(61, 512, 512)`, giving **42 cache hits and 2 misses** — one unavoidable miss per shape.

## Why the one-line fix would have been wrong

Just assigning `self._shape = array.shape` makes the cache hit and **silently returns wrong results from the second call onward**:

```
one-line fix:  call 0: 2.4e-07 OK   call 1: 1.201e+00 WRONG   call 2: 1.182e+00 WRONG
this PR:       call 0: 2.0e-07 OK   call 1: 2.2e-07 OK        call 2: 2.3e-07 OK
```

A pyfftw plan is bound to one specific *buffer*. The `update_arrays` calls that re-point the cached plans sat inside the `if not overwrite_x` branch — and multislice propagates with `in_place=True`, so that branch is skipped. The old code got away with it only because `_new_fftw_object` re-points at build time and it rebuilt every call. This PR re-points on every call instead.

## What is in the cache key, and why

```python
key = (array.shape, array.dtype, array.strides, _fftw_plan_config())
```

- **shape, dtype, strides** — a plan is bound to one buffer *layout*. Probed against pyfftw 0.15 directly: `update_arrays` raises `ValueError` on a mismatch in any of the three.
- **`_fftw_plan_config()`** — `fftw.threads`, `fftw.planning_effort`, `fftw.planning_timelimit`, i.e. everything `_new_fftw_object` bakes into a plan. See below.
- **Deliberately *not* alignment.** See below.

**Cache size and eviction:** a single entry per thread, replaced whenever the key changes. It cannot grow with the number of shapes seen. Alternating between two shapes re-plans on each switch — the same work `dev` does on *every* call, so never worse than the status quo.

**Retention:** a plan holds a reference to the buffer it was last pointed at, so one wave array stays reachable per thread until the propagator is freed (64 MiB at 64×512² complex64). This is not new — `dev`'s `self._fftw_objects` retained the same buffer between calls — but it is now per thread rather than per instance. Propagators are built per `multislice_and_detect` call, so the retention ends with the task.

## Numerics: plan flags must never depend on the input buffer

My first attempt derived an `FFTW_UNALIGNED` planner flag from the incoming array's alignment, to stop `update_arrays` rejecting an under-aligned buffer. **CI caught that as a real numerics regression**, and it was right to:

- An `FFTW_UNALIGNED` plan selects different codelets and returns different numbers — measured on the x86 runners at **~2.3e-07 relative**, the order of float32 epsilon.
- glibc's malloc only guarantees 16-byte alignment while FFTW's `simd_alignment` is 32 on x86, so the flag flipped from one allocation to the next. Observed on CI across four consecutive calls: `flags=('FFTW_UNALIGNED',), ('FFTW_UNALIGNED',), (), ('FFTW_UNALIGNED',)`.
- So the result depended on where malloc happened to place the buffer. Two runs that must agree bitwise stopped agreeing — `test_finite_multislice_chunked` (same multislice, different potential chunk size) and an exact-equality check in `test_ionization`, both at ~2e-7.
- It also thrashed the very cache this PR exists to fix, since the flag was part of the key.

Plans are now **always built with the same flags, exactly as on `dev`**, so the result cannot depend on the buffer. A buffer a plan will not accept is aligned by *copying* — a copy changes no arithmetic, whereas re-planning for it does.

**Verified bitwise identical to `dev`** across shapes 32–256, both `overwrite_x` paths, and deliberately offset buffers. No test tolerance was loosened.

`test_cached_fftw_convolution_result_is_independent_of_buffer_alignment` guards the invariant.

**Coverage caveat on the `byte_align` fallback:** it does not fire on arm64 macOS, where `simd_alignment` is 4 and nothing is ever under-aligned — I confirmed it stays unexercised there. On the x86 runners a plan built this way reported `input_alignment=16`, and an offset buffer raised exactly the `ValueError` the fallback catches (`Invalid input alignment: The original arrays were 16-byte aligned`), so the offset-buffer test should exercise it there. I inferred that from a separate diagnostic rather than instrumenting the fallback inside CI, so treat it as strongly indicated rather than directly observed.

## Caching must not freeze the FFTW config

`_new_fftw_object` reads `fftw.threads`, `fftw.planning_effort` and `fftw.planning_timelimit` when it builds a plan. On `dev` every call rebuilt the plan, so these were re-read each time and a `config.set` block took effect immediately. Keying only on the buffer layout pinned the plans to whatever configuration the *first* call happened to see:

```
dev        : plans rebuilt under changed config = 2  [(4, 'FFTW_ESTIMATE'), (4, 'FFTW_ESTIMATE')]
before fix : plans rebuilt under changed config = 0  []
```

A user writing `with config.set({"fftw.threads": 8})` would silently have kept getting single-threaded plans, and since thread count and planning effort can change which algorithm FFTW selects, a stale plan is a numerics concern as well as a performance one.

Those three settings are now in the cache key, mirroring `_configure_cufft_cache`, which is applied on every GPU dispatch for exactly this reason. Costs 0.85 µs per call against a ~1.2 s propagate step at 64×512².

`test_cached_fftw_convolution_respects_config_changes` covers all three settings and fails without the fix.

## Thread safety

The cache is thread-local. The old code was genuinely racy, not theoretically: 8 threads × 60 calls on a shared instance gave 28 wrong results *and* `TypeError: 'NoneType' object is not subscriptable` crashes.

**This was latent on `dev`**: `multislice_and_detect` builds a fresh `FresnelPropagator` per call, and a real scan showed 2 instances used by 1 thread each. It bites only if an instance is shared — which the documented `propagator=` reuse argument invites.

## Performance — modest, and honestly measured

The win is **2–6% of a propagate step**, not the dramatic number "2 plans per call" suggests. FFTW wisdom accumulates process-globally, so re-planning after the first call is only ~20 ms per plan at 64×512² — mostly the `np.zeros_like` scratch — against ~1.2 s of real FFT work.

| config | dev | this PR | saved |
|---|---|---|---|
| 64×512² | 1249 ms | 1208 ms | 41 ms (3.3%) |
| 64×256² | 189 ms | 178 ms | 11 ms (5.8%) |
| 1×512² | 19.5 ms | 18.4 ms | 1.1 ms (5.5%) |

Also removes **256 MiB of transient scratch allocation per call** at 64×512², which may matter more than the time.

**Caveat on the numbers:** I could not get a usable whole-scan wall-clock measurement — the benchmark machine was at load average 26 and paired scan timings scattered from 20 s to 80 s. The figures above are min-of-N, with `dev` and this PR interleaved in one process so machine drift cancels; that is robust to load in a way the scan-level number is not, so I am deliberately not quoting a scan-level speedup. Someone on a quiet machine may want to confirm.

Measured on macOS arm64, Python 3.13, pyfftw 0.15, numpy 2.2.5, `FFTW_MEASURE`, `fftw.threads=1`.

## Scope

- No public API change. The touched names are all private: `CachedFFTWConvolution._fftw_objects` / `._shape` (replaced by `._local`), and the single call site `FresnelPropagator._cached_fftw_convolution` in `abtem/multislice.py`. `_fftw_alignment_flags` was added and removed within this PR and never existed on `dev`.
- `_new_fftw_object` and `get_fftw_object` are unchanged.
- CPU only by construction — the GPU path never reaches this class.

## Picklability

Storing the plan cache in `threading.local()` made every `CachedFFTWConvolution` instance unpicklable — including one that has never been used:

```python
>>> cloudpickle.dumps(FresnelPropagator())
TypeError: cannot pickle '_thread._local' object
```

On `dev`, a fresh, unused instance pickled fine (`_fftw_objects`/`_shape` are `None`); pickling only ever failed once a plan had actually been built (`pyfftw.FFTW`'s C extension has no `__reduce__`), so this was a strictly wider failure than before. `conventional_multislice_step`'s `propagator` parameter is documented specifically for this kind of reuse ("the main reason for using this argument is to reuse a previously calculated propagator"), so a caller following that documentation with `dask.distributed`'s process-based scheduler — sending a freshly constructed, not-yet-used propagator to a remote worker — would have regressed from working to an immediate `TypeError`.

Fixed with `__getstate__`/`__setstate__`: the cache is a rebuildable performance artifact, not state that needs to survive a pickle round trip, so it is dropped on pickling and a restored instance transparently rebuilds its own cache rather than shipping a stale one across (a plan cache built on one process/thread is not valid on another regardless). Verified `cloudpickle.dumps(FresnelPropagator())` round-trips before first use, and that a pickled-and-restored instance which *had* built a plan still produces correct output.

## Not fixed (pre-existing, out of scope)

A non-contiguous input still cannot be planned for, since `_new_fftw_object` plans on a contiguous `np.zeros_like`. Unreachable from `propagate`, which passes a contiguous `waves._array`. The strides key means such a case raises a clear pyfftw error rather than silently reusing a mismatched plan.

## Tests

New tests in `test/test_fft.py`:

| test | guards | fails on `dev`? |
|---|---|---|
| `..._reuses_plans` | the cache actually hits | yes |
| `..._correct_on_a_cache_hit` (×2: `overwrite_x`) | plans re-pointed at the current buffer | **no** — see below |
| `..._replans_on_layout_change` (×2: shape, dtype) | shape and dtype invalidate the key | **no** — see below |
| `..._is_thread_safe` | no cross-thread buffer sharing | yes |
| `..._result_is_independent_of_buffer_alignment` | numerics do not depend on allocation | yes |
| `..._respects_config_changes` (×3: threads, planning_effort, planning_timelimit) | `config.set` still takes effect | yes |
| `..._is_picklable_before_first_use` | an unused instance survives `cloudpickle` | yes |
| `..._is_picklable_after_use` | a used instance drops its cache rather than failing to pickle | yes |

**Correction:** an earlier version of this description claimed all tests fail without their fix. That does not hold for `correct_on_a_cache_hit` and `replans_on_layout_change` (4 of the 12 parametrized cases): both pass unmodified against `dev`, because "rebuild from scratch on every call" trivially satisfies "plan matches the current buffer" and "shape/dtype change triggers a rebuild" — there is no cache on `dev` to get wrong. They still guard real invariants of the fixed implementation and are kept; their docstrings now say so instead of implying they discriminate. The other 8 do fail on `dev` as claimed.

Local suite: **1462 passed, 859 skipped** — `pytest test/ -k "not gpaw"`.


TomaSusi:
@pzeiger: I discovered this when doing Apple MPS related work on my MacBook M2, and it seems this is too serious of an omission to have slipped by... so approach with skepticism. Could be a regression introduced by recent changes, not sure.

Currently working to make the linux-based CI not fail.

pzeiger:
## Review

The `self._shape` bug this fixes traces to `6cb7735a` (2023-11-06, "fixed issue with cached FFTW when the shape of the transform is changed") — the commit meant to fix a different shape issue is the one that introduced this one. `git log -L256,261:abtem/core/fft.py` on `dev` shows every later commit touching the file (mypy/typing/docstring passes through 2024) left the two lines untouched. Not a recent regression — it has been live for close to three years.

### Verified by running the branch

- All 10 new tests in `test/test_fft.py` pass on `ee72d0f8`. `mypy`/`flake8` on `fft.py` are clean; the same 6 pre-existing overload/import errors reproduce on `dev` at shifted line numbers, unrelated to this change.
- The claim "each [new test] confirmed to fail without the corresponding fix" does not hold for 4 of the 10 parametrized cases. Running this PR's `test/test_fft.py` verbatim against unfixed `dev`: both parametrizations of `test_cached_fftw_convolution_correct_on_a_cache_hit` and both of `test_cached_fftw_convolution_replans_on_layout_change` pass unmodified, because `dev`'s "rebuild from scratch on every call" behavior trivially satisfies "plan matches the current buffer" and "shape/dtype change triggers a rebuild" — there's no cache there to get wrong. The other 6 (`reuses_plans`, `is_thread_safe`, `result_is_independent_of_buffer_alignment`, all three `respects_config_changes` parametrizations) do fail on `dev` as claimed.
- A plausible-looking concern was checked and ruled out. FFTW plans built without `FFTW_UNALIGNED` require the alignment recorded at plan-build time (confirmed at 16 bytes on x86_64 Linux by forcing the `ValueError` directly), which glibc's malloc does not unconditionally guarantee — so the new `except ValueError: pyfftw.byte_align(array)` fallback looked like it could end up firing on every call in the multislice hot path, turning the fast path into an exception-driven one. A real `Probe.multislice(Potential(...))` run (33 slices, `fft: fftw`) on x86_64 Linux was instrumented directly: 0 fallback triggers across 33 calls, 5 plan builds. Ordinary numpy-allocated wave arrays satisfy the alignment the plan actually requires; only the deliberately offset synthetic buffer in the new test exercises the fallback. Not a defect.
- `mergeStateStatus` is currently `CONFLICTING` against `dev`, but only because an unrelated 3-line addition landed in `test/test_fft.py` on `dev` after this branch was cut — a trivial rebase, not a functional conflict.

### Verdict

The fix itself is correct, and more complete than a minimal `self._shape = array.shape` patch would be — it additionally closes the thread-safety and config-staleness gaps a minimal fix leaves open. Before merge: correct the "confirmed to fail without the fix" claim for the two test names that don't actually discriminate against `dev` (they're still worth keeping as invariants against a future partial fix, just not as regression tests against the current bug).

pzeiger:
Hi @TomaSusi,
this bug was also turned up by my Claude instances a few days ago. It is fixed somewhat in PRs #295 and #385, probably because the agents recognized it also and allied fixes. I think the fix here seems to be more thorough, though, so it might be worth to land this PR before #295 and #385.
Best,
Paul

pzeiger (changes requested):
## Review

The `self._shape` bug this fixes traces to `6cb7735a` (2023-11-06, "fixed issue with cached FFTW when the shape of the transform is changed") — the commit meant to fix a different shape issue is the one that introduced this one. `git log -L256,261:abtem/core/fft.py` on `dev` shows every later commit touching the file (mypy/typing/docstring passes through 2024) left the two lines untouched. Not a recent regression — it has been live for close to three years.

### Verified by running the branch

- All 10 new tests in `test/test_fft.py` pass on `ee72d0f8`. `mypy`/`flake8` on `fft.py` are clean; the same 6 pre-existing overload/import errors reproduce on `dev` at shifted line numbers, unrelated to this change.
- The claim "each [new test] confirmed to fail without the corresponding fix" does not hold for 4 of the 10 parametrized cases. Running this PR's `test/test_fft.py` verbatim against unfixed `dev`: both parametrizations of `test_cached_fftw_convolution_correct_on_a_cache_hit` and both of `test_cached_fftw_convolution_replans_on_layout_change` pass unmodified, because `dev`'s "rebuild from scratch on every call" behavior trivially satisfies "plan matches the current buffer" and "shape/dtype change triggers a rebuild" — there's no cache there to get wrong. The other 6 (`reuses_plans`, `is_thread_safe`, `result_is_independent_of_buffer_alignment`, all three `respects_config_changes` parametrizations) do fail on `dev` as claimed.
- A plausible-looking concern was checked and ruled out. FFTW plans built without `FFTW_UNALIGNED` require the alignment recorded at plan-build time (confirmed at 16 bytes on x86_64 Linux by forcing the `ValueError` directly), which glibc's malloc does not unconditionally guarantee — so the new `except ValueError: pyfftw.byte_align(array)` fallback looked like it could end up firing on every call in the multislice hot path, turning the fast path into an exception-driven one. A real `Probe.multislice(Potential(...))` run (33 slices, `fft: fftw`) on x86_64 Linux was instrumented directly: 0 fallback triggers across 33 calls, 5 plan builds. Ordinary numpy-allocated wave arrays satisfy the alignment the plan actually requires; only the deliberately offset synthetic buffer in the new test exercises the fallback. Not a defect.
- `mergeStateStatus` is currently `CONFLICTING` against `dev`, but only because an unrelated 3-line addition landed in `test/test_fft.py` on `dev` after this branch was cut — a trivial rebase, not a functional conflict.

### Further stress-testing

Beyond the PR's own tests: a 200-trial fuzz sweep over random shape/dtype/`overwrite_x` combinations, a non-contiguous array fed directly to `CachedFFTWConvolution` (sliced with a stride, and separately with negative strides), 2000 calls alternating between two shapes while watching RSS, and a 5-second stress run racing `config.set()` from two threads against six threads doing convolutions concurrently (67,628 calls total). All correct, no crashes, no memory growth. The fix holds up well past what its own test suite exercises.

### A pickling regression

Storing the plan cache in `threading.local()` makes every `CachedFFTWConvolution` instance unpicklable — including one that has never been used:

```python
>>> import cloudpickle
>>> from abtem.multislice import FresnelPropagator
>>> cloudpickle.dumps(FresnelPropagator())
TypeError: cannot pickle '_thread._local' object
```

On `dev`, a fresh, unused instance pickles fine (`_fftw_objects`/`_shape` are `None`); pickling only ever failed once a plan had actually been built (`pyfftw.FFTW`'s C extension has no `__reduce__`), so this is a strictly wider failure than before, not a pre-existing one. `conventional_multislice_step`'s `propagator` parameter is documented specifically for this kind of reuse ("the main reason for using this argument is to reuse a previously calculated propagator"), so a caller following that documentation with `dask.distributed`'s process-based scheduler — sending a freshly constructed, not-yet-used propagator to a remote worker — would regress from working to an immediate `TypeError`.

### Suggested patch

The cache is a rebuildable performance artifact, not state that needs to survive a pickle round trip — dropping it and rebuilding fresh on the other side is the correct behavior, not just a workaround:

```python
def __init__(self):
    self._local = threading.local()

def __getstate__(self) -> dict:
    # threading.local is not picklable, and a plan cache built on one
    # process/thread is not valid on another -- drop it rather than the
    # object as a whole failing to pickle. FresnelPropagator's documented
    # `propagator=` reuse argument invites sending an unused instance to
    # a dask.distributed worker before it has cached anything.
    state = self.__dict__.copy()
    del state["_local"]
    return state

def __setstate__(self, state: dict) -> None:
    self.__dict__.update(state)
    self._local = threading.local()
```

Verified: `cloudpickle.dumps(FresnelPropagator())` round-trips before first use; a pickled-and-restored instance that *had* built a plan still produces correct output and transparently rebuilds its own cache rather than shipping a stale one across. Two new tests (`test_cached_fftw_convolution_is_picklable_before_first_use`, `test_cached_fftw_convolution_is_picklable_after_use`) both fail on `ee72d0f8` and pass with this patch; the existing 10 tests and the stress tests above are unaffected; `mypy`/`flake8` stay clean.

### Verdict

The fix itself is correct, and more complete than a minimal `self._shape = array.shape` patch would be — it additionally closes the thread-safety and config-staleness gaps a minimal fix leaves open. Before merge: correct the "confirmed to fail without the fix" claim for the two test names that don't actually discriminate against `dev`, and apply the `__getstate__`/`__setstate__` patch above so a not-yet-used propagator stays picklable.

TomaSusi:
Thanks for the thorough review — the pickling regression especially, that's a real defect I introduced and would not have caught myself.

- **Pickling**: applied the `__getstate__`/`__setstate__` patch verbatim (`0f22e23d`/rebased as `d81f848d`). Reproduced the regression myself first (`cloudpickle.dumps(FresnelPropagator())` on `dev` succeeds, on `ee72d0f8` raises `TypeError`), then confirmed the patch fixes it and your two new test names now pass. Also confirmed they fail on `ee72d0f8`, both before and after the rebase.
- **The "confirmed to fail without the fix" claim**: verified your finding independently — `correct_on_a_cache_hit` and `replans_on_layout_change` (4 parametrized cases) do pass unmodified against `dev`. That claim in the PR description was wrong and has been corrected; the two test docstrings now say why they don't discriminate rather than implying they do. Kept both tests since they guard real invariants of the fixed implementation.
- **Merge conflict**: rebased onto current `dev` (`ab6bf9c0`). It was exactly the trivial import-ordering conflict you diagnosed — a `@requires_gpu` marker addition in `test_fft.py`, no functional overlap.
- **Byte-align fallback / mypy / flake8**: took your findings as reported rather than re-verifying independently — I don't have mypy/flake8 installed in this environment, and can't reproduce the x86 alignment behavior on this arm64 machine. Flagged that in the updated PR description rather than silently asserting coverage I can't confirm firsthand.
- **Fuzz/stress testing and the bug's age**: added the `6cb7735a` provenance note to the description. Appreciate the extra stress testing — didn't reproduce it myself, but nothing in the report is inconsistent with what I found.

Local suite now at 1462 passed (rebase picked up new tests on `dev` plus your two picklability tests). CI re-running on the rebased branch.

pzeiger (approved):
The pickling fix, thread-safety re-verification, and alignment behavior all check out on x86_64 Linux — re-verified independently rather than taken on trust, since the environment that produced the fix couldn't check either.

- `__getstate__`/`__setstate__` matches the suggested patch. `cloudpickle.dumps(FresnelPropagator())` round-trips before first use, and a restored instance that had already built a plan rebuilds its own cache and produces correct output rather than shipping a stale one across.
- Pushed further than before: 5 seconds of `config.set()` racing against four worker threads doing convolutions *and* two threads concurrently pickling/unpickling the same shared instance — 62,904 convolution calls, zero errors, zero mismatches.
- `mypy`/`flake8` on `fft.py` are clean (the same 6 pre-existing, unrelated overload/import errors reproduce on `dev`).
- The alignment-independence test passes on x86_64 Linux, confirming the numerics claim on the platform it's actually about.
- All 39 tests in `test/test_fft.py` pass, including the two new picklability tests, which fail on the pre-fix commit and pass here. CI is green across all four Python versions.

Approving.

PR #416, opened 2026-09-16
Read configuration from abTEM's own location, not dask's

TomaSusi:
## The bug

`abtem/core/config.py` is a fork of dask's config module in which the config directory was carried over but the `collect` call was not adapted. `PATH` was computed at module level:

```python
if "ABTEM_CONFIG" in os.environ:
    PATH = os.environ["ABTEM_CONFIG"]
else:
    PATH = os.path.join(os.path.expanduser("~"), ".config", "abtem")
```

…and then never referenced again anywhere in the file, while `refresh()` called `dask.config.collect()` with no arguments. abTEM therefore only ever saw dask's config paths and dask's `DASK_`-prefixed environment variables.

Verified on `dev` before the fix, with a directory containing `abtem.yaml`:

| | `abtem.config.get("device")` |
|---|---|
| `ABTEM_CONFIG=<dir>` | `cpu` — **silently ignored** |
| `DASK_CONFIG=<dir>` | `gpu` — honored |
| `DASK_DEVICE=gpu` | `gpu` — honored |

So a user following the natural expectation (`~/.config/abtem/abtem.yaml`, or `ABTEM_CONFIG`) got nothing, and abTEM's settings collided with dask's namespace. Two further symptoms: dask's *entire* user config was merged into abTEM's namespace, and `DASK_CONFIG=/tmp/x` itself landed in abTEM's config as a key `config: "/tmp/x"`.

## The fix

abTEM now owns its whole config-source layer:

- **`_get_paths()` / `paths`** — `/etc/abtem` (overridable via `ABTEM_ROOT_CONFIG`), `<sys.prefix>/etc/abtem`, `site.PREFIXES/etc/abtem`, `~/.config/abtem`, then `$ABTEM_CONFIG`, in increasing priority. Mirrors dask's own discovery so the mental model transfers.
- **`collect_env()`** — reads the `ABTEM_` prefix (`ABTEM_DEVICE=gpu`, `ABTEM_DASK__CHUNK_SIZE="256 MB"`). dask's `collect_env` hardcodes `DASK_` and takes no prefix argument, so this is reimplemented on top of its `interpret_value` and abTEM's own `set` class — which means env keys also pass through `check_deprecations`.
- **`collect()`** — `collect_yaml(abtem paths)`, then legacy env, then `ABTEM_` env.

Precedence is `abtem.yaml` defaults < yaml files (later paths win) < `DASK_` env < `ABTEM_` env, so user config overrides the shipped defaults rather than the reverse.

## Two judgement calls

**Dask's yaml paths are no longer read.** This was pure accident, it polluted abTEM's namespace with dask's whole config tree, and no `config.get()` call site in abTEM reads a dask-owned key (every one of the 25 keys read is defined in `abtem.yaml`). A test asserts the two path sets are now disjoint. The one break: anyone who worked around the bug by dropping `abtem.yaml` into `~/.config/dask/` must move it to `~/.config/abtem/` — worth a changelog line.

**`DASK_`-prefixed variables keep working, with a warning.** `collect_legacy_env` honors them but only for keys abTEM itself defines, checked against the defaults tree — so `DASK_DEVICE=gpu` still works and emits a `FutureWarning` naming `ABTEM_DEVICE`, while `DASK_DISTRIBUTED__WORKER__MEMORY__TARGET` is silently left to dask. Warning unconditionally on every `DASK_` variable would have fired on genuine dask settings, which is the whole reason for the filter.

That filter is also why the bootstrap at the bottom of the module is now `_initialize()` before `refresh()` (previously the reverse): the defaults must be registered before collection runs. Precedence is unaffected by the swap, since `update_defaults` applies defaults with `priority="old"`.

Runtime access is untouched — `config.set()`, the context manager, and mutating `abtem.config.config` directly all behave exactly as before. `TestRuntimeAccess` pins that down so this layer can't regress it.

`abtem/core/abtem.yaml` also gains a header pointing users at the override location, since the file is easy to mistake for the place to edit.

## Tests

New `test/test_config.py`, 26 tests: path discovery, yaml files actually being read from the abTEM path (including that overriding one nested key leaves its siblings at their defaults), path and env precedence, underscore↔hyphen canonicalisation, the legacy transition and its filter, and an end-to-end subprocess check that `import abtem` under `ABTEM_CONFIG` picks the file up while `DASK_CONFIG` no longer does.

## Verification

| Check | Result |
|---|---|
| `pytest test/ -k "not gpaw"` | `1419 passed, 859 skipped, 9 deselected` |
| `ruff check abtem/core/config.py` | clean |
| `mypy abtem/core/config.py` | clean |
| `black --check` / `isort --check-only` | clean |
| `pre-commit run --files ...` | check-yaml, end-of-file-fixer, trailing-whitespace, black, flake8 — all pass |

Project-wide `ruff` (245) and `mypy` (367) errors are pre-existing, mostly in `benchmarks/` and `abtem/waves.py`; both checks were scoped to the changed files. `test/` is excluded from ruff and mypy by config, so `test_config.py` is covered by black and flake8.


pzeiger:
## Review

The original bug (config collected from dask's namespace instead of abTEM's) also predates this PR by a long margin — the `refresh()`/`collect()` mismatch traces to the 2024-05-24 "Refactor config module for better organization and readability" commit, not a recent change.

### Verified by running the branch

- All 26 new tests in `test/test_config.py` pass on `a6cb2fbf`. `mypy`/`flake8` on `config.py` are clean. The branch merges cleanly against current `dev` (no conflict). Path discovery and deduplication in `_get_paths()` were checked directly and behave correctly, including the case where `sys.prefix` and `site.PREFIXES[0]` resolve to the same path inside a venv.
- The dead `PATH` computation this PR's own description identifies as the bug's smoking gun is still in the file, unremoved. Lines 48-51 of `abtem/core/config.py` still compute `PATH` from `ABTEM_CONFIG`/`~/.config/abtem` exactly as before, immediately below the new `paths = _get_paths()`. A repo-wide grep confirms nothing reads it — it's exactly as dead after this PR as before. Worth deleting as part of this change rather than leaving two near-identical, only-one-of-them-real variables side by side.
- `ABTEM_CONFIG`/`ABTEM_ROOT_CONFIG` leak into the merged config dict as stray keys. `collect_legacy_env()` correctly filters `DASK_`-prefixed variables through `_defines_key()`, excluding ones like `DASK_CONFIG` that aren't real abTEM settings. `collect_env()`, which reads the `ABTEM_` namespace, applies no equivalent filter — so `ABTEM_CONFIG` and `ABTEM_ROOT_CONFIG`, the two control variables this PR itself defines for path discovery, get swept into the config dict as `config: <path>` / `root_config: <path>`. Reproduction:

  ```python
  from abtem.core import config
  config.collect_env({"ABTEM_CONFIG": "/some/dir"})
  # {'config': '/some/dir'}
  ```

  This reproduces, in the new `ABTEM_` namespace, essentially the symptom this PR fixes in the old `DASK_` one, and isn't covered by the new test suite (`test_genuine_dask_variables_are_left_to_dask` only parametrizes over `DASK_`-prefixed names). Currently inert — nothing reads a `config`/`root_config` key back out — but it would silently collide if `abtem.yaml` ever gains a real top-level key by either name. Suggest either excluding `{"ABTEM_CONFIG", "ABTEM_ROOT_CONFIG"}` explicitly in `collect_env()`, or applying the same `_defines_key()` filter there that `collect_legacy_env()` already uses.

### Verdict

The core fix is correct and the precedence design is sound. Two loose ends before merge: delete the now-genuinely-dead `PATH` block, and close the `ABTEM_CONFIG`/`ABTEM_ROOT_CONFIG` stray-key gap (with a test analogous to the existing `DASK_CONFIG` one).

pzeiger (changes requested):
## Review

The original bug (config collected from dask's namespace instead of abTEM's) also predates this PR by a long margin — the `refresh()`/`collect()` mismatch traces to the 2024-05-24 "Refactor config module for better organization and readability" commit, not a recent change.

### Verified by running the branch

- All 26 new tests in `test/test_config.py` pass on `a6cb2fbf`. `mypy`/`flake8` on `config.py` are clean. The branch merges cleanly against current `dev` (no conflict). Path discovery and deduplication in `_get_paths()` were checked directly and behave correctly, including the case where `sys.prefix` and `site.PREFIXES[0]` resolve to the same path inside a venv.
- The dead `PATH` computation this PR's own description identifies as the bug's smoking gun is still in the file, unremoved. Lines 48-51 of `abtem/core/config.py` still compute `PATH` from `ABTEM_CONFIG`/`~/.config/abtem` exactly as before, immediately below the new `paths = _get_paths()`. A repo-wide grep confirms nothing reads it — it's exactly as dead after this PR as before. Worth deleting as part of this change rather than leaving two near-identical, only-one-of-them-real variables side by side.
- `ABTEM_CONFIG`/`ABTEM_ROOT_CONFIG` leak into the merged config dict as stray keys. `collect_legacy_env()` correctly filters `DASK_`-prefixed variables through `_defines_key()`, excluding ones like `DASK_CONFIG` that aren't real abTEM settings. `collect_env()`, which reads the `ABTEM_` namespace, applies no equivalent filter — so `ABTEM_CONFIG` and `ABTEM_ROOT_CONFIG`, the two control variables this PR itself defines for path discovery, get swept into the config dict as `config: <path>` / `root_config: <path>`. Reproduction:

  ```python
  from abtem.core import config
  config.collect_env({"ABTEM_CONFIG": "/some/dir"})
  # {'config': '/some/dir'}
  ```

  This reproduces, in the new `ABTEM_` namespace, essentially the symptom this PR fixes in the old `DASK_` one, and isn't covered by the new test suite (`test_genuine_dask_variables_are_left_to_dask` only parametrizes over `DASK_`-prefixed names). Currently inert — nothing reads a `config`/`root_config` key back out — but it would silently collide if `abtem.yaml` ever gains a real top-level key by either name.

### A crash found under further stress-testing

Config parsing must never raise on a hot path, and `import abtem` is about as hot as paths get — but a malformed yaml file under `ABTEM_CONFIG` crashes the import outright:

```python
# tmp_path/abtem.yaml: "device: [unterminated\n  - this is not valid yaml: :\n"
ValueError: A dask config file at '.../abtem.yaml' is malformed, original error message: ...
```

This is inherited from `dask.config`'s own `collect_yaml` (a malformed file at `DASK_CONFIG` crashes `dev` today too, confirmed directly), so it isn't new in the strict sense — but this PR is what makes `~/.config/abtem/`/`ABTEM_CONFIG` the newly-documented, blessed place to put that file. Before this PR nothing was ever read from there, so a typo there was harmless; after it, the exact file this PR tells users to write can take down their next `import abtem` with a raw parser traceback instead of a clear message.

### Suggested patch

```python
def collect(
    paths: list[str] = paths,
    env: Mapping[str, str] | None = None,
    defaults: list[Mapping] = defaults,
) -> dict:
    ...
    try:
        yaml_configs = list(collect_yaml(paths=paths))
    except ValueError as e:
        # A malformed yaml file must not take down `import abtem` -- config
        # discovery now points users at exactly the files this can fail on.
        warnings.warn(
            f"Failed to read abTEM's yaml configuration files under {paths}; "
            f"configuration from these files is being skipped: {e}",
            RuntimeWarning,
            stacklevel=2,
        )
        yaml_configs = []

    configs = [
        *yaml_configs,
        collect_legacy_env(env=env, defaults=defaults),
        collect_env(env=env),
    ]
```

Together with excluding `{"ABTEM_CONFIG", "ABTEM_ROOT_CONFIG"}` in `collect_env()` (for the stray-key finding above) and deleting the dead `PATH` block. `dask.config`'s own `_load_config_file` already wraps the underlying `yaml.YAMLError` as a `ValueError` naming the offending file, so the warning message needs no extra work to be actionable. This degrades all yaml sources for that `collect()` call together rather than isolating just the one bad file among several — real per-file isolation would mean replicating `collect_yaml`'s own path-discovery loop, which seemed like more surface than this fix should take on; flagging the coarser granularity in case a finer version is wanted.

Verified: `mypy`/`flake8` stay clean, all 26 original tests still pass, and three new tests — the two leak checks in `TestControlEnvVarsDoNotLeak` plus `test_malformed_yaml_warns_and_falls_back_instead_of_crashing_import` (a fourth, `test_real_keys_still_pass_through`, is a sanity check that passes on both branches by design) — all fail on `a6cb2fbf` and pass with this patch.

### Verdict

The core fix is correct and the precedence design is sound. Before merge: delete the now-genuinely-dead `PATH` block, close the `ABTEM_CONFIG`/`ABTEM_ROOT_CONFIG` stray-key gap, and stop a malformed user config file from crashing `import abtem` — patches for all three above.

TomaSusi:
All three findings confirmed and fixed in d42b6a45:

- Deleted the dead `PATH` block (lines 48-51) — confirmed nothing read it before or after.
- `collect_env()` now excludes `ABTEM_CONFIG`/`ABTEM_ROOT_CONFIG` via `_CONTROL_ENV_VARS`, so they no longer leak into the config dict as `config`/`root_config`.
- `collect()` now catches the `ValueError` a malformed yaml file raises (via dask's `_load_config_file`), warns with `RuntimeWarning`, and falls back to no yaml config for that call instead of crashing `import abtem`. Went with your suggested approach and its stated caveat: this degrades all yaml sources for the call together rather than isolating the one bad file, since real per-file isolation would mean replicating `collect_yaml`'s own path-discovery loop.

Added four tests (`TestControlEnvVarsDoNotLeak` x3, `TestMalformedYaml`) — confirmed all four fail on a6cb2fbf and pass on d42b6a45, existing 26 unchanged. Full suite: `1423 passed, 859 skipped` on `pytest test/ -k "not gpaw"`. `ruff`/`mypy`/`black`/`isort` clean on both changed files.

pzeiger (approved):
All three fixes verified independently, plus a few more edge cases.

- The dead `PATH` block is gone; `ABTEM_CONFIG`/`ABTEM_ROOT_CONFIG` no longer leak into the config dict, confirmed via `collect_env()` directly.
- `import abtem` now survives a malformed `abtem.yaml` under `ABTEM_CONFIG`, with a `RuntimeWarning` naming the file and falling back to shipped defaults — checked end to end through a real subprocess import, not just the unit-level `collect()` call.
- Checked the documented coarse-grained caveat directly: a valid sibling yaml file in the same directory as a malformed one also gets skipped for that `collect()` call, exactly as described, not silently picked up regardless.
- A missing `ABTEM_CONFIG` path and an empty (but syntactically valid) yaml file both degrade cleanly to the shipped defaults.
- `mypy`/`flake8` on `config.py` are fully clean. All 30 tests in `test/test_config.py` pass, including the four new ones, which fail on the pre-fix commit and pass here. CI is green across all four Python versions.

Approving.

PR #417, opened 2026-09-16
Fix center_of_mass() to normalize by total captured intensity

TomaSusi:
## Summary

`DiffractionPatterns.center_of_mass()` computed only the intensity-weighted first moment (`sum(I*k)`) without dividing by the total intensity (`sum(I)`). That's only equal to the true center of mass when the diffraction pattern sums to exactly 1 — it silently breaks whenever intensity is lost to cropping (`.crop()`) or to scattering outside the detector footprint, since the result gets scaled down by whatever fraction of the total intensity is actually captured.

Reported in [discussion #402](https://github.com/abTEM/abTEM/discussions/402): a GaAs crystal simulation cropped to a 7 mrad window around the direct beam only captured ~48.6% of the total intensity, and the computed center of mass was off by almost exactly `1/0.486 ≈ 2.06x` — matching the reporter's observed factor of ~2 discrepancy against an independent reference implementation.

This also fixes `integrated_center_of_mass()` (iCOM/DPC), which builds on `center_of_mass()` and had the same systematic error whenever detected intensity varies across scan positions (e.g. from channeling or absorption).

## Fix

Normalize both `com_x` and `com_y` by the per-pattern total intensity in `DiffractionPatterns._com`.

## Test plan

- [x] Added `test_diffraction_patterns_center_of_mass_is_normalized`, which asserts the computed center of mass is invariant to the fraction of total intensity captured (1.0, 0.5, 0.1) — verified it fails on the pre-fix code for fractions < 1 and passes with the fix.
- [x] Full `test/test_measure.py` suite passes (150 passed, 94 skipped).


TomaSusi:
[andreas-beyer](https://github.com/andreas-beyer) any further changes needed or does this fix survive your testing?

andreas-beyer:
It survived. Sorry for the delay. Our new neoARM is currently set up, this takes lots of my time :)

TomaSusi:
No worries! I'll go ahead and merge this fix then, let me know if you run into anything else and thanks again for reporting.

PR #420, opened 2026-09-17
Fix -0 slices for base-less ArrayObjects: squeeze, chunk bookkeeping, partitioning, and the same bug in fft_crop/fft_interpolate

### LINKED ISSUE #407: ArrayObject.squeeze() is a silent no-op for base-less measurements (shape[:-0] bug)
`ArrayObject.squeeze` silently does nothing whenever a measurement has an empty base shape (e.g. any `AnnularDetector`, or other scalar-per-position detector), found while verifying #404.

## The bug

```python
shape = self.shape[: -len(self.base_shape)]
```

When `base_shape == ()`, `len(self.base_shape)` is `0`, so `-len(self.base_shape)` is `-0`, which Python evaluates as plain `0` — so `shape[:-0]` is `shape[:0]`, which is **always empty**, regardless of `self.shape`. `squeeze()` therefore never removes anything for a base-less measurement.

## Where it shows up

#404 flags this as a "known residual" it deliberately doesn't fix: a single-position PRISM scan with an `AnnularDetector` returns shape `(1,)` with a leftover `PositionsAxis`, where the regular multislice oracle returns `()`. It's pre-existing (this file is untouched by #404) and only became visible there because PRISM's `reduce_ensemble`/squeeze call happens post-detection, on the already base-less measurement — the regular multislice driver's squeeze call happens pre-detection, on the `Waves` object (nonzero base shape), so the `-0` bug never triggers on that path. The underlying bug is general, not PRISM-specific — any code path that calls `.squeeze()` on a base-less measurement hits it.

## Suggested fix

```python
shape = self.shape[: len(self.shape) - len(self.base_shape)]
```

or equivalently guard the zero case explicitly. One-line fix; a regression test should cover a base-less (`AnnularDetector`-style) measurement going through `.squeeze()`, in addition to whatever normal-base-shape case is presumably already covered.

---

Original discussion: [#404](https://github.com/abTEM/abTEM/pull/404)

pzeiger:
Closes #407.

## What

Five sites in `abtem/array.py` slice off the base dimensions with a negative length derived from `len(self.base_shape)`: `shape[: -n]`, `chunks[-n :]`, `chunks[: -n]`. Python has no negative zero for integers, so whenever an `ArrayObject` has no base dimensions (`base_shape == ()` — `MeasurementsEnsemble`, reachable from `Images.to_measurement_ensemble()`), `n` is `0` and each slice collapses to the wrong end: `[:0]` is always empty rather than everything, `[0:]` is always everything rather than nothing. Each site fails a different way:

- **`squeeze()`**: `shape[: -0]` is empty, so a length-1 ensemble axis never disappears.
- **`_has_base_chunks`**: `chunks[-0 :]` is every chunk, so a chunked ensemble is always reported as having base chunks, even though there are none.
- **`no_base_chunks()`**: trusts that wrong report and rechunks the *ensemble* axes into one block — a memory-footprint change on a large ensemble, not a cosmetic one.
- **`_partition_args()`**: `chunks[: -0]` is empty, and dask raises immediately on the shape mismatch.
- **`ArrayObject.apply_transform`'s blockwise callback** (`_apply_transform`) takes the same `n` as `base_ndims` and slices ensemble axes off with it, so applying any `ArrayObjectTransform` to a base-less object either crashes with an axes-metadata/array-ndim mismatch or silently drops every ensemble axis, depending on which side of the `-0` the caller reads. Found by sweeping the tree for every occurrence of the same slicing shape rather than trusting the four already known — #407 names only `squeeze()`.

The same mechanism also lives in `abtem/core/fft.py`'s `fft_crop`/`fft_interpolate`, at the three sites that slice off batch dimensions with `-len(new_shape)`. Left alone at first on the grounds that every in-tree caller passes a concrete, non-empty `new_shape` — true, but two public entry points forward a *caller*-supplied `gpts` straight into these functions with no length check: `Waves.downsample(gpts=...)` and `WavesDetector(gpts=...)`. `WavesDetector`'s guard is `if self._gpts is not None`, so an empty tuple — falsy but not `None` — passes it and reaches `fft_crop` with `new_shape == ()`, which crashed three frames down with an opaque `TypeError` naming neither `gpts` nor the caller. Fixed the same way, same commit.

## Blast radius

Swept `abtem/` for every negative-length slice of this shape. Two lookalikes in `array.py`'s own file are not affected and are left alone:

- `measurements.py:6217` sits in `MomentumResolvedSpectrum.show`, whose `_base_dims` is always `2`.
- `measurements.py:1270` (`_BaseMeasurement2D.interpolate_line`) has the same reason — `_base_dims` is always `2` there too.
- `prism/utils.py:17` subtracts an extra `1`, so it is never exactly `0`.

## Fix

Explicit arithmetic at each site instead of a negative index: `n = len(shape) - len(base_shape); shape[:n]` (and the chunk-tuple equivalents), so a base-less object correctly gets `n == len(shape)` rather than a would-be-negative `-0`. Same pattern in `fft.py`, where `n = len(array.shape) - len(new_shape)` and `new_shape == ()` means every dimension is a batch dimension — the correct generalisation of what these functions already do for a shorter `new_shape`, not a fix confined to the one case that happened to crash.

`no_base_chunks()`'s own line had the same bug, but fixing `_has_base_chunks` alone already makes its guard clause (`if not self._has_base_chunks: return self`) return early for every base-less object, so that line is unreachable there in practice. Fixed anyway — a later change to `_has_base_chunks` alone should not be able to resurrect it — and pinned with a test that bypasses the guard directly, since the public-method test alone cannot exercise it.

`WavesDetector(gpts=())` now returns the same result as the documented `gpts=None` default (no resampling) rather than crashing partway through one — confirmed byte-identical. `Waves.downsample(gpts=())` still raises, now with a plain `IndexError` at the point `downsample` tries to read a new grid out of `gpts`, rather than the same opaque `TypeError` inside `fft_crop`. `downsample` has no no-op interpretation of an empty `gpts` the way `WavesDetector`'s optional resampling does, so rejecting it is correct; the change is only that it now fails at the point that actually explains why.

## Tests

`TestBaseLessArrayObject` (`test/test_array.py`) builds a base-less `MeasurementsEnsemble` through the public API and exercises all five `array.py` sites. `TestFftCropInterpolateEmptyNewShape` (`test/test_fft.py`) pins `fft_crop`/`fft_interpolate` directly, plus the `WavesDetector(gpts=())` case end to end against the `gpts=None` oracle.

Reverting all five `array.py` fixes together fails all six of that class's tests; reverting any one fails exactly one. Reverting any of the three `fft.py` fixes fails the tests that exercise it (`fft_crop`'s underlies all three of its class's tests; `fft_interpolate`'s own two only fail the two tests that go through `fft_interpolate`). `test_array.py` (204), `test_fft.py` (42), `test_measure.py`, `test_ensemble.py`, `test_waves.py`, `test_prism.py` and `test_multislice_input.py` all pass otherwise.


abtem/array.py
```diff
@@ -1475,7 +1475,11 @@ def squeeze(self, axis: Optional[tuple[int, ...]] = None) -> Self:
         else:
             axis = normalize_axes(axis, self.shape)
 
-        shape = self.shape[: -len(self.base_shape)]
+        # Not `self.shape[: -len(self.base_shape)]`: Python has no negative
+        # zero, so for a base-less object (base_shape == ()) that slice is
+        # `[:0]`, always empty, rather than `[:len(self.shape)]`.
+        n_ensemble_dims = len(self.shape) - len(self.base_shape)
```
TomaSusi, line 1481: Nice, thorough fix — traced all 8 sites (5 in `array.py`, 3 in `fft.py`) and confirmed each one is correctly converted from the `-len(...)` "negative zero" footgun to the safe `n = len(a) - len(b)` form, with no site missed anywhere else in the repo.

Minor simplification suggestion: the same 2-line pattern (`n = len(a) - len(b); a[:n]`/`a[n:]`) is repeated at all 8 sites with near-duplicated comments. A small shared helper, e.g.

```python
def _split_ensemble(seq, base_len):
    n = len(seq) - base_len
    return seq[:n], seq[n:]
```

would remove ~40 lines of repeated explanation and make a future 9th call site written the old way impossible instead of just discouraged.

pzeiger:
Thanks for tracing all 8 sites — that count is right, and the conversion at each one holds up on a fresh check.

The shared-helper suggestion doesn't hold up as well under a closer look, though, so it isn't taken.

**The two-line arithmetic (`n = len(a) - len(b)`, `a[:n]`) is shared, but what follows it isn't.** Every site needs exactly one of the two halves a `_split_ensemble` helper would return, never both — `squeeze()`, `_apply_transform`, `_partition_args` want `[:n]`; `_has_base_chunks` and `fft_interpolate`'s two sites want `[n:]`. Returning a pair every caller then discards half of doesn't save code at the call site, it adds an unpacking line. And four of the eight layer something past the plain slice: `no_base_chunks()` appends `(-1,) * len(base_shape)`, `fft_crop` concatenates the caller's own `new_shape`, both `fft_interpolate` sites feed the result straight into `np.prod(...)`. A helper that returns `(a[:n], a[n:])` doesn't shorten any of those — the concatenation or reduction still has to be written by hand right after it.

**The "~40 lines of repeated explanation" doesn't match what's in the diff.** Only 4 of the 8 sites carry the full explanation (`squeeze`, `_has_base_chunks`, `fft_crop`, `fft_interpolate`'s first site); the other 4 already say "see the comment in X" rather than repeating it. So the actual repeated content is the two-line arithmetic idiom, not the prose — and collapsing eight two-line, slightly-different idioms into one three-line generic helper trades a small amount of duplication for an extra indirection that doesn't fit any one site exactly. Three similar lines here beat a premature abstraction.

TomaSusi (approved):
Fair — the rebuttal holds. Returning a pair when every site only ever uses one half doesn't save anything at the call site, and I overstated how much of the comment text is actually repeated versus already deduplicated via "see the comment in X". No change needed, thanks for checking it in detail.

PR #423, opened 2026-09-17
Fix prism eels reduction chunking

pzeiger:
## What

`prism_transition_potential_scan`'s per-site reduction crops a bounding box spanning **every** scan position up front (`minimum_crop(pixel_positions, output_window_gpts)`), then runs one `tensordot` contracting all positions against it — once per scattering site. Peak memory is `n_positions * (scan_span + window)**2`, growing with the scan's spatial extent rather than with the output window, so a production-sized field of view demands a single allocation that crosses any fixed device memory: 118,796,904,960 bytes (110.6 GiB) on a 40 GB A100, for the hBN run in `abtem_issues/prism_eels_reduction_allocates_whole_scan.md` (672×720 grid, `interpolation=3`, 98×101 scan positions, one K-edge with 4 transitions).

The reduction is now chunked over spatially contiguous blocks of scan rows: `minimum_crop`, the periodic-wrap embedding, the `tensordot` and the per-position crop all run per block instead of once, with each block's contribution written into its own slice of the pre-allocated measurement. A strided block scheme would not help here — the bounding box is set by a block's own spatial spread, not its position count — so blocks have to stay spatially contiguous. Positions are already flattened in scan-raster order, so a contiguous range of the flat position axis is a contiguous range of whole rows of the scan's leading axis for any scan dimensionality, including the 0-d single-point case (one row, one column — one block, identical to today's behaviour).

Each block's own `Waves` object needs ensemble-axis metadata restricted to that block's row range, not the whole scan's: `CustomScan`'s `PositionsAxis` (an `OrdinalAxis` subclass) carries an explicit per-position `values` tuple, which `Waves` validates against the array on construction, so the full-scan tuple does not fit a partial block. `AxisMetadata.__getitem__` already does this restriction — the same mechanism dask's own ensemble partitioning uses per block — so slicing through it handles `PositionsAxis` and `GridScan`'s linear `ScanAxis` (which carries no such tuple and is unaffected by the slice) uniformly.

Row count per block is sized from a memory budget rather than fixed. `estimate_scan_batch_size` already estimates how many probe wavefunctions at a given grid size fit a live memory budget elsewhere in this driver; converting that count (sized for one position at `output_window_gpts`, assuming no bounding-box growth) into a row count gives a first guess, divided by the number of transitions sharing the same budget. That guess is then checked against the box a block of that size actually produces — translation-invariant for a regular raster scan, so the first candidate block is representative — and halved until it fits the same budget the guess was converted from. This check runs even when the guess already covers the whole scan (capped at the row count): a batch spanning every row is exactly where the no-growth assumption is most wrong, since the bounding box is then at its largest, and a large `output_window_gpts` (no PRISM downsampling, `interpolation=1`) makes the naive guess clear the row count routinely rather than rarely.

## Numbers

| | rows/block | blocks | peak reduction allocation |
|---|---|---|---|
| before | — (one block) | 1 | 110.6 GiB (measured OOM, see issue file) |
| after, CPU-sized budget (128 MB `dask.chunk-size` default) | 1 | 98 | ≈1.2 GiB |
| after, GPU-sized budget (representative of a 40 GB A100, not measured) | 5 | 20 | ≈3.1 GiB |

The "after" rows recompute this file's own geometry (672×720 grid, `interpolation=3`, 98×101 scan, `n_T=4` transitions) with the same arithmetic style the issue file uses; the GPU row additionally assumes a representative PRISM beam count, since the original crash log does not record the run's exact value. That count does change the resulting block size and block count non-trivially — re-running the GPU row's arithmetic at beam counts from 50 to 2000 moves the block count from 20 to 98 — but the shrink-and-verify pass targets the *byte* budget directly, so the peak allocation itself stays within roughly 2.4–3.1 GiB across that same range, until the block size hits its one-row floor (at very large beam counts the floor can no longer shrink further and the byte figure can exceed the budget; not reached at the file's grid size in this check). Both rows use `estimate_scan_batch_size`'s existing device/dtype-aware budget, so the block count adapts automatically to whatever memory is actually available rather than being fixed at either end.

## Measured on real GPU hardware

The numbers above are analytic, on the failing job's own geometry. Separately, on a Perlmutter A100, a synthetic Si₂ test geometry (128×128 potential grid, `interpolation=1`, one exit plane) at small/medium/large scan size confirms the sizing is not just memory-safe but need-proportional: the auto-sizer chooses **1 row-batch for the small scan (8×8 = 64 positions, no chunking at all), 32 for medium (32×32 = 1,024 positions, one row per batch) and 64 for large (64×64 = 4,096 positions, one row per batch)** — read directly from the settled batch size every real reduction call shares, not inferred from timing. Correctness (chunked vs a forced single whole-scan block) matched to float32 rounding at small and medium; large has no single-block baseline to compare against, since that is exactly the regime the original defect OOM'd in.

Wall-clock and peak-memory numbers from that same run are not reported here: the first measurement omitted a warm-up call, so first-call costs (CUDA context, cuFFT plan and cuBLAS handle creation) landed on whichever size ran first, most visibly at the small scale, where they dwarfed the tiny amount of real work — this project's own "warm up before timing anything" convention exists for exactly this failure mode. A warmed-up re-run is pending; this section will carry those numbers once it lands.

## Blast radius

Both `SMatrix.transition_potential_scan` call paths — `double_channel=False` and `double_channel=True` — route through the same `_reduce_and_record` closure and get the chunking without separate changes; both are covered in the tests below. `prism_transition_potential_scan_beam_basis` (the sibling beam-basis reduction) does not call `minimum_crop` or this function and is untouched. The lazy PRISM path calls the same eager driver once per dask block, so a lazy run with one scan-position chunk gets the same per-call chunking as an eager one; lazy chunking over the scan itself is unaffected and orthogonal to this change.

## Tests

`TestPrismEelsReductionChunking` in `test/test_core_loss_fixes.py`:

- `test_chunked_reduction_matches_a_single_whole_scan_batch` — chunked output matches a forced single whole-scan block bit-for-bit (to float32 rounding), across `double_channel` and three forced memory budgets that resolve to row blocks of 1, 2 and 4 (checked directly by recording `minimum_crop`'s call sizes), on a 7×5 scan where 2 and 4 do not divide the 7 rows evenly (exercises an uneven last block). This cannot fail on unfixed `dev` by construction — it compares two code paths this change introduces, so it is regression coverage for the new chunking logic rather than a guard against the original defect.
- `test_the_reduction_never_crops_around_the_whole_scan` — forces a one-position memory budget and asserts `minimum_crop` never sees the full scan. This is the guard against the original defect: on unfixed `dev` it fails (recorded call size 54 of 54, i.e. one call spanning every position, regardless of the forced budget, since the bounding box is computed once, globally, before the site loop); it passes here.
- `test_matches_reference_with_a_custom_scans_positions_axis` — a `CustomScan` with a deliberately non-uniform, two-cluster position layout, matched against a forced single whole-scan block reference across two forced budgets. `CustomScan` is the case the axis-metadata restriction above exists for; the non-uniform layout checks that correctness does not depend on the sizing heuristic's raster-uniformity assumption, only the chosen block size might.
- `test_matches_reference_with_multiple_exit_planes` — a multi-exit-plane potential in both single- and double-channel mode, matched against the same forced-budget reference pattern. Multiple exit planes add a broadcast slice (single-channel) or a plain int (double-channel) ahead of the scan axes in the measurement's leading indices; composing that with the new row slice is the trickiest indexing this change adds, and no other test here uses more than one exit plane.
- `test_minimum_crop_does_not_scale_with_the_number_of_sites` — `minimum_crop`'s result for a row batch depends only on the batch's own positions, never on the site or exit plane; asserts the call count doesn't change between a 2-site and an 8-site run of the same scan. Fails on unfixed `dev` (14 vs 56 calls, exactly `n_sites × n_batches`).
- `test_the_reduction_still_shrinks_when_the_naive_guess_already_covers_the_whole_scan` — a scan physically wider than its ~8 Å window (unlike this class's other tests, which are drawn in the same few Å as the window and never exercise real bounding-box growth), with a budget picked to land the naive guess exactly at the row count. Checks the *last* recorded `minimum_crop` call rather than the max, since the verify-and-shrink pass itself legitimately probes the full-size candidate before shrinking it — the last call is always from the real per-site reduction, which follows the sizing pass. Fails without the `rows_per_batch > 1` gate (a single unchunked call at the full scan size).


abtem/inelastic/core_loss.py
```diff
@@ -1794,86 +1853,117 @@ def _reduce_and_record(scattered_window, site_xy, exit_idx):
...
 
-        position_waves_shape = (
-            waves_at_positions.shape[:-3]
-            + scan_shape
-            + waves_at_positions.shape[-2:]
-        )
-        waves_at_positions = waves_at_positions.reshape(
-            position_waves_shape
-        )
+            reduce_crop_corner, reduce_size, reduce_corners = minimum_crop(
```
TomaSusi: Traced the row-batching logic carefully (coefficient/position slicing, `PositionsAxis`/`ScanAxis` restriction, measurement write-back indices) — it's correct, and the tests (uneven batches, non-uniform `CustomScan`, multiple exit planes, double-channel) give good coverage.

One efficiency note: `minimum_crop(pixel_positions[flat_start:flat_end], output_window_gpts)` here depends only on the row batch — not on `site_xy`/`exit_idx`/`scattered_window` — but `_reduce_and_record` is called once per (site, exit plane) pair. For a production scan with many sites and exit planes, this recomputes the same bounding box redundantly, partly working against the memory-savings goal of this PR. Worth hoisting `(reduce_crop_corner, reduce_size, reduce_corners)` per row batch once before the site loop and indexing into it here instead.

pzeiger:
Confirmed and fixed. `minimum_crop`'s result for a row batch never depends on the site or exit plane, so computing it inside `_reduce_and_record` (called once per site × exit plane) recomputed the identical box every time — for the 7-batch, 8-site case in the new regression test, 56 calls where 7 would do.

Hoisted the per-batch `(crop_corner, size, corners)` triples out to a list built once, alongside `row_batches` itself, and `_reduce_and_record` now indexes into it instead of recomputing. No behaviour change — same values, computed at a different point in the call graph; the existing chunking tests all still pass bit-for-bit. Added `test_minimum_crop_does_not_scale_with_the_number_of_sites`, which runs the same scan at two site counts and asserts the call count doesn't move — it fails on the pre-hoist code (14 vs 56 calls) and passes now.

TomaSusi (approved):
Nice catch beyond the hoist — the `rows_per_batch < n_rows` gate skipping the shrink-verification exactly when the naive guess already reaches `n_rows` was silently reproducing the whole-scan-as-one-block defect this PR exists to fix, and it's not a corner case (any scan wider than its output window hits it). Good regression test for it too.

Checked out `722cee1` and ran the full `test_core_loss_fixes.py` suite: 109 passed, 1 pre-existing skip, including both new regression tests in isolation. Approving.

Issue #413, opened 2026-09-16
safe_equality guesses when it cannot compare an attribute: lazy arrays always compare equal, sliced atoms never


pzeiger:
Split out of #412 rather than settled there, because it is a question about what `==` should mean library-wide rather than about any one class. #412 fixes the `Potential` half and deliberately leaves the rest of this alone; the two together are the whole picture.

`safe_equality` reduces every attribute to one verdict and has no third state for "I could not compare this". Three situations reach that gap, and each falls into whichever guess the control flow happens to produce. On `dev` @ `f4238571`:

**Deferred content is guessed EQUAL.** A dask-backed attribute produces a dask value from `==`, which is never the `False` singleton, so it never counts as a difference. Every lazy `ArrayObject` compares equal to every other regardless of content.

**Uncomparable content is guessed UNEQUAL.** `except (KeyError, TypeError, ValueError): return False` turns "this attribute raised" into "these objects differ". `SliceIndexedAtoms._slice_index` is a `list` of index arrays, and `list == list` on numpy elements raises `ValueError`, so that class never compares equal to anything, an identical twin included. It is the cache `Potential` builds for `projection="infinite"`, the default; the `projection="finite"` cache is a `SlicedAtoms`, which has no `_slice_index` and compares correctly.

**Content with no comparison defined is guessed UNEQUAL.** `MagneticField` and `VectorPotential` hold a `QuasiDipoleProjections` subclass as `_integrator` — `QuasiDipoleMagneticFieldProjections` and `QuasiDipoleVectorPotentialProjections` respectively — and the base is not an `EqualityMixin`, so `==` falls back to object identity and neither class compares equal to its own `copy()`.

All three are silent, and none is reachable through library control flow — only `AxisMetadata` uses equality internally, and it is never dask-backed — so the exposure is user code and tests. `test_copy_equals` misses all three: it compares an object with itself, and never draws a `MagneticField`.

## Reproduction

```python
import ase.build, dask.array as da, abtem
from abtem.measurements import Images
from abtem.potentials.iam import Potential
from abtem.slicing import SliceIndexedAtoms

atoms = ase.build.bulk("Si", cubic=True)

with abtem.config.set({"fft": "numpy"}):
    zeros = Images(da.zeros((16, 16), dtype="float32"), sampling=0.1)
    ones = Images(da.ones((16, 16), dtype="float32"), sampling=0.1)
    print("lazy Images(zeros) == lazy Images(ones) :", zeros == ones)
[...]

PR #424, opened 2026-09-17
Make SliceIndexedAtoms and the magnetism integrators comparable

pzeiger:
## What

`safe_equality` has no third state for "could not compare" — only "equal" and "not equal" — and two classes hit that gap in the same way `_sliced_atoms` did for `Potential` before #412.

`SliceIndexedAtoms._slice_index` is a list of integer arrays, one per slice. `list.__eq__` reduces each pairwise `==` (itself an array) to a bool, which numpy refuses for anything but a length-1 array. `safe_equality`'s exception guard turns that `ValueError` into "unequal", so this class never compared equal to anything, including an identical twin. It is the cache `Potential(projection="infinite")` builds — the default — which is why #412 could only fix the `finite`-projection half of `Potential` equality (`abtem_issues/equality_guesses_when_it_cannot_compare.md`, upstream #413). Gives it its own `__eq__`: compare `_slice_index` directly with `np.array_equal` per element, then delegate every other attribute to `safe_equality` unchanged.

`QuasiDipoleProjections` (`MagneticField`/`VectorPotential`'s `_integrator`) was a plain class with no `__eq__` at all, so it fell back to object identity and neither class ever compared equal to its own `copy()`. Inheriting `FieldIntegrator` (already `EqualityMixin, CopyMixin`) fixes that, once `LyonParametrization` — held as `_parametrization` and otherwise stopping the recursion on object identity — becomes an `EqualityMixin` too. Its own `periodic`/`finite` properties are removed in favour of the ones `FieldIntegrator` already provides, now backed by a `super().__init__(periodic=False, finite=True)` call — the same values the removed properties returned, so nothing observable changes there.

## A third defect, not in the issue file

`_tables`, an integral-table cache `QuasiDipoleProjections` populates lazily one element at a time, turned out to be the same shape of problem twice over: an unpopulated cache stops a used integrator comparing equal to a fresh one, and two integrators that cached the identical element compare unequal anyway, since `dict.__eq__` on numpy-array values hits the same `ValueError` `_slice_index` does. Same fix as `_DeviceArrayCache` (`integrals.py`) and `Potential`'s own `_sliced_atoms`: `_eq_exclude = ("_tables",)` — a cache is incidental state, never identity.

The issue file's first symptom — a lazy dask-backed attribute compares equal regardless of content, so two lazy `Images` built from different data compare equal while the eager equivalents correctly differ — is left as-is. It is an open design question about what `==` should mean for a lazy object, not a comparability bug in a specific class, and any fix touches every `EqualityMixin` in the tree rather than the two classes here. `safe_equality` already documents why it skips a deferred value rather than forcing it (running the graph it is comparing, discarding the result, letting a mid-graph error escape `==`).

## Blast radius

`SliceIndexedAtoms` is `Potential`'s default (`projection="infinite"`) sliced-atoms cache; `SlicedAtoms` (`projection="finite"`) has no `_slice_index` and is untouched. `QuasiDipoleProjections` has exactly two concrete subclasses, `QuasiDipoleMagneticFieldProjections` and `QuasiDipoleVectorPotentialProjections` — neither overrides `periodic`/`finite` itself, both already implement the (now genuinely enforced, since inheriting `FieldIntegrator` makes the class an ABC) `_calculate_integral_table` abstract method. Verified end to end: both build correctly, produce bit-identical output across `.copy()`, and report the same `periodic`/`finite` values as before.

## Tests

`TestEqualityDiscriminates` in `test/test_copy.py`:

- `test_slice_indexed_atoms_equals_an_identical_twin` / `test_slice_indexed_atoms_with_different_binning_are_not_equal`
- `test_magnetic_field_equals_its_own_copy` / `test_magnetic_field_with_different_atoms_is_not_equal`
- `test_quasi_dipole_integrator_ignores_its_table_cache` — unpopulated on both, populated on one, populated on both

Three of the five fail on unfixed `dev`; the two discrimination tests cannot, by construction — object-identity fallback already says "unequal" for two distinct objects, so there is nothing for the fix to change there. That mirrors the existing `test_potentials_with_identical_atoms_are_still_equal` pattern in the same file.


TomaSusi (approved):
Traced the fix in detail: `list`/`dict` equality on numpy-array elements raising `ValueError` (silently caught by `safe_equality` as "unequal") explains why these classes never compared equal even to their own copy. Verified:

- `SliceIndexedAtoms.__eq__` correctly special-cases the ragged `_slice_index` list via `np.array_equal` per element, deferring the rest to `safe_equality`.
- `QuasiDipoleProjections(FieldIntegrator)` is safe to instantiate — it already implements both abstract methods (`cutoff`, `integrate_on_grid`) — and `_eq_exclude = ("_tables",)` mirrors the existing `_DeviceArrayCache.__eq__` / `Potential._sliced_atoms` pattern elsewhere in the codebase.
- `LyonParametrization(EqualityMixin)` is safe — its `_parameters` is a plain JSON-loaded dict, no numpy arrays, so structural comparison doesn't hit the same pitfall.
- No sibling class (`SlicedAtoms`, the other `FieldIntegrator` subclasses) has the same unfixed defect.

No correctness issues found.

PR #425, opened 2026-09-17
Fix transition potential device memo

pzeiger:
## What

Both core-loss drivers finish setup with `transition_potential.copy_to_device(device)`, unconditionally, once per task. On GPU that is a full host-to-device transfer of the whole payload; on any device it also rebuilds the object through `__init__`, which recomputes `_local_potential` (an `ifft2` + `abs2` + sum over the array) even when the array is already on the target device and the free `copy_to_device` is a no-op (`abtem_issues/transition_potential_uploaded_to_gpu_per_task.md`). Once the transition potential arrives as one graph node shared by every task on a worker (`BaseTransitionPotential._task_local`), both costs are avoidable — every task on a worker uses the identical array.

`copy_to_device` on `TransitionPotentialArray` now memoizes the expensive part per `(shared graph node, device)`, using `_device_cache_key` (the shared per-content-device helper from #392, rather than adding a sixth ad-hoc cache).

## Why the memo can't just cache the returned object

`scatter()` and `generate_scattered_waves()` mutate what they are handed (`self._array = copy_to_device(...)`, `self.grid.match(...)`), so two tasks sharing one instance would race on those mutations exactly as `_task_local` exists to prevent one step earlier. What is cached is the immutable-in-practice array data — the device-resident array and the local potential derived from it — and `copy_to_device` still returns a fresh, independently-mutable wrapper each call, built via `copy.copy` rather than `__init__` so the `_local_potential` recomputation disappears too.

## A latent bug this exposed

The memo dict is set once in `__init__`, on the object every `_task_local` view is shallow-copied from, so sibling per-task views are meant to share it via `copy.copy`'s reference semantics. They didn't, at first: with no `__copy__` defined, `copy.copy` goes through `__reduce_ex__`/`__getstate__` for *both* pickling and in-process copying, and `__getstate__` already zeroed the pre-existing `_local_potential_device_cache` on every copy — correct for pickling, and silently defeating cross-task sharing for either cache. `__copy__` now does a true shallow `__dict__` share, matching what `_task_local`'s own docstring already promised ("everything else stays shared"); `__getstate__` is untouched, so pickling still drops both caches exactly as before.

## Numbers

Measured in calls to the underlying upload/rebuild, not wall-clock, on CPU (`copy_to_device("cpu")`, so this table isolates the call-count claim from any device-timing question, answered separately below). Small, medium, and large simulated task counts against one shared graph node, `_task_local()` then `copy_to_device("cpu")` per task:

| tasks | calls, unfixed (1 per task) | calls, fixed |
|---|---|---|
| 4 | 4 | 1 |
| 64 | 64 | 1 |
| 256 | 256 | 1 |

One call regardless of task count — O(1) per worker instead of O(n_tasks) — across all three sizes tested.

## Measured on real GPU hardware

The call-count table above is device-agnostic. Separately, on a Perlmutter A100, the actual PCIe
transfer and rebuild cost this fix avoids repeating:

| | memoized | unmemoized |
|---|---|---|
| cache miss (one real upload, per device, isolated) | ~4.4–5.3 ms | — |
| cache hit (one call, isolated) | ~0.03–0.05 ms | — |
| batch average per task (64/256 tasks, single- and multi-GPU) | ~0.06–0.17 ms | ~3.0–3.2 ms |

The isolated miss/hit rows are timed per call, one at a time. The batch-average row is a whole
64- or 256-task run's total time divided by task count, so it still carries the 2 one-time misses
(one per device) diluted across many hits — arithmetic on the isolated numbers predicts
2×miss + (n−2)×hit averaged over n tasks, which lands at 0.07–0.18 ms for n=64/256, matching the
measured batch averages. A 4-task batch average is deliberately not reported: at that size the
2 diluting misses dominate enough that a single end-of-batch synchronization (rather than one per
task) reads anomalously high in a way that does not match either the isolated per-call numbers or
the 64-/256-task batches — a benchmark-measurement artifact of that synchronization choice at very
low task counts, not a real per-task cost. The unmemoized cost is constant per task regardless of
batch size, since every call redoes the same rebuild. For any workload with more than a couple of
tasks per device — every real one — the memoized batch-average settles at a durable ~20–50x cut
in per-task cost, not just a first-call saving.

A dedicated 2-GPU correctness check (`bench_device_memo_multigpu.py`, alternating
`with cp.cuda.Device(i):` between calls, matching how abTEM's own multi-GPU dask workers switch
device) confirms the memo stays correct in exactly the scenario `_device_cache_key` exists for —
one process driving several GPUs: the array lands on the device named by the ambient context at
call time, the two devices get two distinct cache entries rather than aliasing onto one, values
round-tripped from each device match the host source, and a repeat call under the same device
context reuses the cached array object rather than re-uploading.

## Blast radius

`TransitionPotentialArray.copy_to_device` is called from exactly two sites, both per-task setup in the core-loss drivers (`multislice.py`, `_prism_eels_common_setup` in `core_loss.py`); grepped for every other `.copy_to_device(` call site in the tree and none touches a transition potential — `finite_difference.py`/`multislice.py`'s `potential_slice`, `s_matrix.py`, and `array.py`'s generic `to_cpu`/`to_gpu` are unrelated classes, unaffected. `to_cpu()`/`to_gpu()` (inherited, calling `self.copy_to_device`) now benefit from the memo too via ordinary virtual dispatch. No other `copy.copy(...)` call site touches a transition potential besides `_task_local`.

Benign, precedented race under the threaded scheduler: two tasks missing the cache simultaneously redundantly compute their own valid `(array, local_potential)` pair rather than corrupting a shared one — the same no-lock shape `_local_potential_on_device` (PR #375) already has for the sibling cache, accepted there.

## Tests

`TestTransitionPotentialDeviceMemo` in `test/test_core_loss_fixes.py`: repeated calls share the uploaded array while returning independent wrappers; mutating one wrapper's `_array` (as `scatter()` does) does not corrupt what a later call pulls from the cache; two `_task_local` siblings of the same node share one upload; memoized output matches an unmemoized rebuild; pickling still drops both device caches, the exact scenario `__getstate__`'s existing comment describes. Three of the five fail on unfixed `dev` (one on the assertion itself, two with `AttributeError` for the `_device_array_cache` this branch adds); the other two — the mutation-safety check and the unmemoized-rebuild comparison — hold on both, since they exercise properties that must be true regardless of whether memoization exists.

Run on a Perlmutter A100: the full `test_core_loss_fixes.py` suite passes, 103/103, including this class. `bench_device_memo_multigpu.py` (not part of the pytest suite; a standalone GPU correctness/timing check, referenced above) adds the multi-GPU coverage no CPU-only test can: distinct devices, distinct cache entries, correct round-tripped values, and cache reuse on repeat, under an actual 2-GPU allocation.


TomaSusi (approved):
Traced the mechanism carefully: `TransitionPotentialArray.copy_to_device` now memoizes the device-resident array + `_local_potential` per device key, and the new `__copy__` override makes `copy.copy` (used by `_task_local`) share `_device_array_cache` by reference instead of going through `__getstate__`, which was silently resetting it on every per-task view and defeating the memo.

Checked in detail:
- Sharing most attributes by reference across sibling task-local views isn't new — `_task_local` already documents "everything else stays shared" and explicitly privatizes the two things that actually get mutated (`_grid`, `_accelerator`). `_device_array_cache` only ever grows by key-insertion, never mutated in place, so sharing it is safe.
- `_local_potential_device_cache` reset logic is unaffected in practice — every real call path goes through the new `copy_to_device`, which unconditionally resets it to `None` on its output.
- `_device_cache_key` (reused from `abtem/integrals.py`) is used consistently with its existing multi-GPU convention.
- No circular import; `Self` matches the existing `>=3.11` requirement and prior usage elsewhere.
- Ran the 5 new tests plus the full `test_core_loss_fixes.py` suite from a scratch worktree off this branch: 107/108 passed (1 pre-existing skip).

No correctness issues found.

Issue #421, opened 2026-09-17
ChargeDensityPotential is not charge-neutral: full Z added against a valence-only density


TomaSusi:
## Summary

`ChargeDensityPotential` adds each atom's **full atomic number** `Z` as a Gaussian-broadened point charge (`add_point_charges_fourier`), but the density it is documented to take is **valence-only** (`ZVAL` electrons per atom). The cell therefore carries a net **+Nc per atom**, where `Nc = Z - ZVAL` is the core-electron count. Nothing cancels it: the Poisson solve simply drops `k = 0`, so the excess is smeared into a uniform neutralising background.

A net charge contributes `q / (eps0 k^2)`, which diverges at low `k` — exactly where imaging contrast lives. The effect is therefore concentrated in the low-order structure factors that dominate HRTEM and diffraction.

This was found while validating `VASPPotential` (#395), which does not have the problem: it subtracts `POTCAR`'s radial core density, so `Z - Nc - ZVAL = 0` by construction.

## Evidence

The error scales with the core fraction, as the diagnosis predicts. Measured against converged `GPAWPotential` references at matching `gpts`/slicing:

| system | `sum(Nc)/sum(Z)` | correlation | RMS diff (% of peak) |
|---|---|---|---|
| hBN | 0.333 | 0.9898 | 3.61% |
| SrTiO3 | 0.548 | 0.8516 | 9.26% |
| MoS2 | 0.757 | 0.8968 | 10.86% |

Per low-order reflection it is much starker. Ratios to the GPAW reference for SrTiO3:

| reflection | k [1/Å] | current |
|---|---|---|
| (-1,0) | 0.253 | **74.7** |
| (-3,0) | 0.760 | **24.9** |
| (-2,-1) | 0.567 | **20.8** |
| (-1,-1) | 0.358 | 4.37 |
| (-2,0) | 0.507 | 2.56 |

The worst of these are the near-forbidden perovskite reflections, whose true amplitudes are small — but 74.7x takes the (-1,0) from 0.85 to roughly 64, comparable to the *strong* reflections (24-31). In other words the current construction injects a spurious feature as strong as a real Bragg peak, at a position where the structure factor should nearly cancel.

## Proposed fix

Subtract the core electrons that the `+Z` point charge implicitly assumes, so the cell is neutral. This needs no GPAW and no VASP:

[...]

PR #426, opened 2026-09-18
Deduplicate test suite fixtures and fix slow-test categorization

TomaSusi:
## Summary
- Consolidates duplicated setup across ~26 test files into shared `test/utils.py` helpers: `devices`/`lazy_params` parametrize decorators (standardizing the `[gpu, "cpu"]` order/spelling suite-wide), `to_host_array`, `si_cubic_atoms`/`si_diamond_atoms`, and a device-aware `synthetic_transition_potential` — replacing ~90+ copy-pasted parametrize sites and several independently-reimplemented helpers.
- Fixes bugs found while deduplicating:
  - A dead duplicate test definition in `test_measure.py` (one copy silently shadowed the other, never ran).
  - `test_waves.py`'s `exit_plane_waves` fixture ignored `device`, so its "GPU" parametrized cases silently ran the CPU build twice.
  - `test_prism.py::test_prism_scan_match_probe_scan` had its real assertions commented out; restored a genuine equality check.
  - `test_core_loss_fixes.py` duplicated precision tests already covered in `test_core_loss.py`; removed the duplicates.
- Audits and fixes test timing:
  - `gpaw_calculator_{no_bonding,bonding}` fixtures in `test_gpaw.py` were function-scoped despite being read-only, re-running a full GPAW DFT calculation per test; now module-scoped (saves ~30s on machines with GPAW installed).
  - Marks several disproportionately expensive, previously-ungated tests as `@pytest.mark.slow`: two deliberately brutal concurrency stress tests in `test_atomic_potential.py`, and three real-GPAW-pipeline integration tests in `test_ionization.py`/`test_potential_parametrization.py`.
  - Splits `TestBlochWavesEnergyEnsemble` onto a dedicated 2-energy list (`BLOCH_ENERGIES`) instead of the 3-energy list shared with the cheap PRISM ensemble tests, cutting that class from ~145s to ~92s. Verified empirically that 2 energies still yields non-degenerate per-energy active-beam sets, so no coverage is lost.
  - Removes an incomplete `test_charge_density_potential` stub from `test_gpaw.py` (no assertions).
- `test_lobato_gpaw_match`'s occasional flakiness (GPAW's own atomic SCF solver failing to converge for certain elements) was found during this pass but left untouched — it's being addressed in a separate open PR.

## GPU-VRAM-safe parallel testing + CI
Prompted by real GPU-workstation timing runs (see PR comments below) surfacing a separate issue: `pytest -n auto` sizes its worker pool from CPU core count, with no awareness that a `device` parametrization means real VRAM. Enough xdist workers landing a GPU-parametrized test at once exhausts the card, surfacing as an OOM deep inside a kernel launch rather than a clean failure.

- `test/conftest.py`'s `pytest_collection_modifyitems` now tags every GPU-touching test — any parametrized value (direct or indirect, whatever the parameter is named) equal to the resolved `gpu` device string, plus anything carrying the `multigpu` marker — with a shared `xdist_group("gpu")`. Combined with `--dist=loadgroup` (added to `pyproject.toml`'s pytest `addopts`), xdist schedules all of them onto a single worker, so at most one GPU-touching test runs at a time regardless of `-n`, while CPU-only tests still fan out across every worker. Verified: 663/2515 collected tests get grouped correctly, zero false positives on CPU-only tests, full suite still passes.
- Multi-GPU tests (`test_multigpu.py`, real `dask_cuda`/`LocalCUDACluster` usage) share the *same* group rather than getting their own — this is deliberate: putting them in a separate group would let them run concurrently with regular single-GPU tests on a different worker, reintroducing the exact VRAM-contention problem this fix exists to prevent.
- CI (`.github/workflows/tests.yml`): `-n auto` is now the default for the test workflow (safe there since GitHub-hosted runners have no GPU, so the grouping above is inert). `--runslow` now runs for PRs into `main` (using the existing release-gate OS matrix) and for every `push`/`schedule`/`workflow_dispatch` trigger, so a slow-test regression on `dev` surfaces within minutes of a merge rather than only in the weekly cron.
- Adds an `all-tests-passed` aggregator job so branch protection has one stable check to require (`"All tests passed"`), rather than needing to name every OS x Python matrix cell individually and re-edit that list whenever the matrix changes.

## Test plan
- [x] Full suite collects the same tests (module count changes intentionally reflect removed duplicate/redundant tests).
- [x] Full suite passes: `pytest test/ -n auto` — 1670 passed, 845 skipped, 0 failed.
- [x] `pytest test/ --runslow` passes for all newly-`slow`-marked and newly-refactored tests.
- [x] `test_gpaw.py` and `test_energy_ensemble.py` independently verified against a real GPAW install (module-scoped fixtures still produce correct, passing results; `TestBlochWavesEnergyEnsemble` timing improvement measured directly).
- [x] `pytest_collection_modifyitems`'s GPU-detection logic verified by introspecting collected items directly: correctly groups all `[gpu]`/indirect-`gpu`-parametrized tests and every `test_multigpu.py` test, with no false positives.
- [x] CI green on this branch across the full Python matrix, including the new `all-tests-passed` job.


TomaSusi:
@pzeiger Could you run a multi-GPU pass of the test suite on your workstation? This complements the timing analysis behind this PR by exercising the GPU- and multi-GPU-gated tests that are otherwise silently skipped on CPU-only machines.

```bash
git fetch origin
git checkout claude/test-suite-review-refactor-c37a95
git pull
```

If you have `dask-cuda` installed and ≥2 GPUs available:

```bash
python -m pytest test/ --runslow -m multigpu --durations=0 -q > /tmp/durations_multigpu.log 2>&1
```

Also useful, the full `--runslow` run (GPU + CPU paths, sorted slowest-first):

```bash
python -m pytest test/ --runslow --durations=0 -q > /tmp/durations_gpu.log 2>&1
```

Both can take a while in serial (our CPU-only run took ~12 minutes; GPU paths add real device-side work). Whenever you have `/tmp/durations_multigpu.log` and/or `/tmp/durations_gpu.log`, share them here or with Toma — we're specifically looking for GPU-path costs that don't show up on a CPU-only machine (slow `[gpu]`-parametrized cases, host↔device transfer overhead, etc.) and any GPU-specific slow-test miscategorization.

**Update:** since this comment, the branch gained a fix for a separate issue Toma hit on his own GPU workstation: `pytest -n auto` was spawning one xdist worker per CPU core with no awareness that a `device` parametrization means real VRAM, and enough of them landing a GPU test at once was exhausting the card. `test/conftest.py` now routes every GPU-touching test — single-device *and* your multi-GPU (`test_multigpu.py`/`dask_cuda`) tests alike — onto one shared `xdist_group("gpu")`, so at most one of them runs at a time regardless of `-n`, while CPU-only tests still parallelize fully.

This is safe for your multi-GPU run, and if anything protects it further: your `LocalCUDACluster`-based tests now can't get scheduled concurrently with an unrelated single-GPU test on another worker (which would have been its own VRAM-contention hazard). So `-n auto` is now safe to add to either command above if you want the CPU-bound portion of the run to go faster — the GPU/multi-GPU tests will still execute one at a time either way, so it won't change what you're measuring on the GPU side, just the wall-clock for everything else in the same run.

TomaSusi:
Heads-up on a merge overlap with [abTEM/abTEM#414](https://github.com/abTEM/abTEM/pull/414): both this PR and #414 modify the same `gpu = pytest.param(...)` definition in `test/utils.py`.

- This PR keeps it gated on `_gpu_count() < 1` (CUDA-only skip check).
- #414 replaces that with `_accelerator_device() is None`, so `gpu` also resolves to `"mps"` on Apple Silicon.

These are compatible in intent, not a real conflict — whichever PR merges second will need to manually keep #414's `_accelerator_device()`-based resolution when resolving the textual conflict on that line. The `devices`/`lazy_params` decorators this PR adds just reference the `gpu` variable, so they'll automatically pick up MPS support once #414's version of `gpu` is in place, with no further changes needed on this side.

Verified this by running this PR's branch on a CUDA GPU workstation (2508 passed, 7 skipped, 0 failed) and comparing against a durations log from #414's sync branch on Apple Silicon (2403 passed, 84 skipped, 0 failed) — same underlying `gpu` sentinel mechanism, different resolved accelerator.

pzeiger:
Thanks Toma. Re-reviewed the head `fd904e38`: all five points of the first round are in as proposed, and the four new commits were read, tested individually, and then attacked with inputs their tests do not cover. Verdict: ready apart from two regressions, both small and both fixed with tests on `parity-projection-fixes-2` (three commits on top of your head, [compare](https://github.com/abTEM/abTEM/compare/claude/parity_projection_phonons...parity-projection-fixes-2)), plus two pre-existing defects found on the way and fixed there as well.

**The four new commits**

- `758be138`, minimum-image check via `find_mic`: right diagnosis (the rank-3 requirement skipped the wrap for ASE's 2D builders), but the replacement keys the wrap on `atoms.pbc`, and hand-built ASE `Atoms` default to `pbc=False`. A bulk cell with an ordinary displacement wrapped across a boundary is now rejected as a 9.95 Å displacement (reproduced with `Atoms("H2", cell=[10, 10, 10])`, no `pbc`), where the previous version accepted it. abTEM treats the potential as periodic along every cell vector regardless of `pbc`, so the wrap directions should come from the cell: wrap along every non-zero cell vector, skip the zero one. That keeps the rank-2 case and restores the bulk `pbc=False` case; the fix branch does exactly that with a test for each.
- `2fd401c2`, `block_direct=True` resolving to a 1 mrad radius: correct, a real bug, and it was in the code before this PR as well. Both call sites fixed and tested.
- `4ac6f6a5`, eight cleanups: seven are fine. The new `ensemble[:, k]` indexing is broken for its own use case: `__getitem__` copies the parent's `ensemble_axes_metadata` into the slice, so the slice's `EnergyLossAxis` keeps all energies while its snapshots keep one, and building a `Potential` from it raises an ordinal-axis size mismatch. The pre-existing non-parity `ensemble[k]` has the same defect, and `ensemble[i, j]` crashes on a bare `Atoms` object. All three fixed in `72aa303f` (the constructor rebuilds the axis from the sliced energies; one test covers the three forms, each followed by a multislice). Two minor notes besides: (a) in the parity `__getitem__`, `item == slice(None)` on a numpy index array evaluates elementwise and raises a `ValueError` instead of the documented `NotImplementedError`; an `isinstance(item, slice)` guard fixes it (included in `f971e480` with a test). (b) `_ensemble_mean_forced` is a dataclass field, so `axis_to_dict` serializes it into zarr like `_ensemble_mean`; old files load fine (the field defaults), but files written by this version will not load in an older abTEM, whose `axis_from_dict` passes the unknown key to the constructor. Same forward-compatibility trade-off as `_ensemble_mean` already makes, so just noting it.
- `fd904e38`, scoped warning test: fine.

One defect in the earlier fixes, found by the same pass and fixed in the third commit of that branch: `unfold_loss_gain` promised to accept a `MomentumResolvedSpectrum` but raised on one, since the spectrum's energy is a base axis; it now unfolds along that axis, and unfolding before or after `momentum_resolved_spectrum` gives the same result.

**Checks run on `fd904e38`**: the seven touched test files (energy-resolved parity, phonon loss, parity multislice, axes grid, spectral detectors, array, waves) pass, 382 tests, plus the adversarial cases above; also verified that the private `_validated` flag does propagate through dask chunk reconstruction, that a `Probe.scan` with a `CustomScan` over a parity potential flows through `phonon_loss_diffraction_patterns` with its positions axis intact, and that a parity result survives a zarr round trip.

**Follow-up, separate from this PR.** The one-phonon channel of a bin-restricted snapshot lacks the Debye–Waller factor of the modes outside the bin (about 12 % in intensity at the outer spots for graphene), which parity projection cannot supply. A second sign average over a rest displacement field recovers it: `EnergyResolvedAtomsEnsemble(..., rest_snapshots=...)` propagates every snapshot at `R_eq + s·u_bin + t·u_rest` for `s` in (+1, −1, 0) and `t` in (+1, −1): the parity axis gains a `"static"` member (the rest-displaced structure without the bin displacement) and a new `PhononRestParityAxis` follows it. `phonon_loss_diffraction_patterns` averages the exit waves over the rest axis before the projection and subtracts the static member from the even part per realization. The part even in `u_rest` carries the damping; the part odd in `u_rest`, one bin phonon plus one rest phonon, is mis-binned and cancels exactly; the static reference keeps the two-rest-phonon fluctuations (order `u_rest⁴`, larger than the bin's own two-phonon signal) out of the multi channel. Verified on graphene against `exp(-Q² σ²)` at the one-phonon peaks, and the multi/one ratio stays that of the bin-only run. It is on `parity-rest-displacements`, merged with your current head ([compare](https://github.com/abTEM/abTEM/compare/claude/parity_projection_phonons...parity-rest-displacements)), and can go in here or as its own PR after this one merges, whichever you prefer; six runs per snapshot instead of two, opt-in, nothing changes without the new argument.

pzeiger:
## Timing runs and a few fixes along the way

Ran the full-suite and `multigpu` timing sweeps requested above, both with `-n auto` added (per your later comment that this is safe and doesn't change what's measured on the GPU side, since the grouping mechanism keeps GPU-path tests to one at a time regardless of `-n`) rather than the plain serial commands originally given. Full duration logs are attached (`durations_gpu.log`, `durations_multigpu.log`); both reflect the branch with all three fixes below applied. `pytest test/ -q -n auto` on Perlmutter (4×A100): **2515 passed, 26 skipped, 0 failed, 0 errors in 463s.**

While going through the branch to run those, found and fixed three small issues along the way, each pushed directly rather than left as a suggestion: a wrong `dashboard_address` default (`80a9a80c`), a missing `tryfirst` on a pytest hook (`4c8c780d`), and a GPU-test detection check that only looked at one of two places a GPU-touching test declares itself (`70c562a2`). Each is small, self-contained, and has a single clearly correct answer rather than being a design judgment call, and each is confirmed working — the first two by a subsequent run (one full-suite, one on real GPU hardware), the third against every previously-missed instance directly (see below). Pushing directly avoided a round trip through a second review cycle on a branch already under active review; happy to have any commit amended, squashed, or reverted if a different approach is preferred.

## FIXED: the `xdist_group("gpu")` marker never actually took effect, for any test

This was more severe than it first looked: the whole grouping mechanism this PR adds was a no-op, not just incomplete for some test shapes.

`conftest.py` added the marker dynamically, inside `pytest_collection_modifyitems`:

```python
def pytest_collection_modifyitems(config, items):
    ...
    if is_gpu or "multigpu" in item.keywords:
        item.add_marker(pytest.mark.xdist_group("gpu"))
```

pytest-xdist has its own `pytest_collection_modifyitems` hook (`xdist/remote.py`, `WorkerInteractor`) that reads each item's `xdist_group` marker and appends a `@groupname` suffix to its nodeid — that suffix is what `LoadGroupScheduling._split_scope()` (`xdist/scheduler/loadgroup.py`) actually groups on. Hook call order across plugins follows registration order unless a hook opts into `tryfirst`/`trylast`; xdist's own copy of this hook ran before a plain conftest.py hook got a chance to add the marker, so by the time xdist read `item.iter_markers("xdist_group")`, it wasn't there yet. The marker got added, but one hook call too late to matter — nodeids never gained the `@gpu` suffix, and `LoadGroupScheduling` never saw a group to schedule.

Verified with a minimal, GPU-free local reproduction (six trivial tests, `--dist=loadgroup -n 4`, a conftest that adds `xdist_group` the same way abTEM's does): the six tests scattered across all four workers, exactly like abTEM's real `test_multigpu.py` did on Perlmutter. A statically-declared `pytestmark = [pytest.mark.xdist_group("gpu")]` on the same tests correctly grouped all six onto one worker. Adding `@pytest.hookimpl(tryfirst=True)` above the dynamic hook function fixed it too — all six tests correctly grouped, with the `@gpu` nodeid suffix present.

This one mechanism failure plausibly explains everything else GPU-shaped seen while testing this PR: `test_multigpu.py`'s `cuda_client` fixture (`scope="module"`, meant to build one dask-cuda cluster and cache it for all six tests in the module) showed all six tests paying a ~4.3-4.5s "setup" cost instead of one real setup plus five free cache hits — because with grouping inert, the six tests could land on up to six different xdist workers, each an independent process with its own module-level cache.

**Fixed in `4c8c780d`**: `@pytest.hookimpl(tryfirst=True)` added immediately above `pytest_collection_modifyitems` in `conftest.py`. Checked for a competing `tryfirst` on the same hook name across every plugin actually installed in this environment (`xdist`, `hypothesis`, pytest core) — none exists, so `tryfirst=True` alone is sufficient to guarantee this hook now runs before xdist's; there's no ordering fragility against the currently-installed plugin set. Also checked that nothing else in the test suite adds `gpu`/`multigpu`/`slow` markers dynamically from a different hook that this one could now be racing against by running earlier — the only marker mutations happen inside this same function.

**Confirmed on real GPU hardware, Perlmutter (4×A100), `test/test_multigpu.py -m multigpu -n auto -v` at `4c8c780d`:** all 6 tests now carry the `@gpu` nodeid suffix and all landed on the same worker (`[gw0]`), out of 64 workers created for the run — previously they scattered across 6 different workers. Total time also dropped from 50.30s to 35.38s, consistent with `cuda_client` now being built once and reused across the module instead of rebuilt independently by however many workers happened to land a test. The fix is verified both against the exact mechanism (a faithful, minimal, GPU-free reproduction) and now empirically on the real suite.

**One real, if intended, consequence worth flagging explicitly rather than discovering at merge time:** this fix makes GPU-test serialization *actually happen* for the first time — every GPU-parametrized test (774 `@gpu`-suffixed entries in the attached `durations_gpu.log`, plus the `multigpu` cases) now genuinely queues onto a single worker under `-n auto`, rather than being free to fan out across all of them. That's the mechanism working as designed, not a regression, but it does mean the full-suite wall-clock (463s, attached) reflects that serialization cost, not just the CPU-bound `-n auto` speedup — the two aren't separable in a single number.

## FIXED: the GPU-detection logic itself also missed real GPU tests

Independent of the hook-ordering bug above — fixed in the same pass since both live in the same function — the *is-this-test-GPU-touching* check only caught one specific shape:

```python
callspec = getattr(item, "callspec", None)
is_gpu = callspec is not None and any(
    isinstance(v, str) and v == _GPU_DEVICE
    for v in callspec.params.values()
)
if is_gpu or "multigpu" in item.keywords:
    item.add_marker(pytest.mark.xdist_group("gpu"))
```

This missed two real shapes already in the suite:

- **Any bare `@requires_gpu`-decorated test with no parametrize at all, or one whose parametrize doesn't include the device string.** `item.callspec` is `None` for a bare test, so `is_gpu` short-circuits to `False` before the check even runs; a test parametrized on something else entirely (e.g. `dtype`) has a callspec, but none of its values is `"gpu"`. Confirmed instances: `test_fft.py::test_oversized_plan_bypasses_cache` (bare), `test_potential_chunking.py::TestComplexExponential::test_gpu_matches_cpu` (parametrized only on `dtype`/`expected_cdtype`), and four bare tests in `test_multigpu_logic.py` (`test_gpu_no_client_forces_synchronous`, `test_multigpu_config_starts_cluster`, `test_single_gpu_multigpu_config_no_cluster`, `test_explicit_scheduler_skips_cluster`).
- **Any parametrized test whose GPU-selecting value isn't the literal string `"gpu"`.** `test_ionization.py`'s `xp_params = ["numpy", pytest.param("cupy", marks=requires_gpu)]` parametrizes on a backend-module name, not the device sentinel — `"cupy" == _GPU_DEVICE` is `False`, so this case was untagged despite running real cupy code. `test_prism.py::test_s_matrix_store_on_host` is parametrized only on `lazy`, with `@requires_gpu` on top and `device="gpu"` used internally — invisible to the callspec check.

All of these ran un-grouped: free to land on any xdist worker concurrently with each other, or with an `xdist_group("gpu")`-tagged test on a different worker — the exact VRAM-oversubscription failure this mechanism exists to prevent.

**Fixed in `70c562a2`**: the hook now also checks, via `item.iter_markers()`, whether the item carries a `skipif` mark whose `reason=` matches `requires_gpu`'s or `requires_multigpu`'s (imported from `test/utils.py` rather than hardcoded, so the check can't drift from whatever text those two helpers actually use). This catches a marked test regardless of how it's applied — bare decorator, mixed with an unrelated parametrize, or passed via `pytest.param(..., marks=requires_gpu)` — independent of whether `"gpu"` shows up anywhere in its callspec. **Verified against every confirmed instance above** (`test_fft.py::test_oversized_plan_bypasses_cache`, `TestComplexExponential::test_gpu_matches_cpu` both cases, all four `test_multigpu_logic.py` tests, `test_ionization.py`'s cupy-parametrized cases, `test_prism.py::test_s_matrix_store_on_host` both cases — 11 cases total): all now carry the `@gpu` nodeid suffix and group onto a single worker under `--dist=loadgroup`, confirmed both on this session's own GPU hardware and, separately, on Perlmutter (4×A100, `-n auto`, 64 workers created, all 11 landed on `[gw1]`). Confirmed the CPU-only sibling cases (`test_cpu_matches_reference`, `test_fast_roll_matches_numpy_roll[numpy]`) are correctly left untagged and still scatter across workers — no over-tagging.

## One disclosed tradeoff worth a second look

`TestBlochWavesEnergyEnsemble`'s energy list was cut from 3 values to 2 (`BLOCH_ENERGIES = [100e3, 200e3, 300e3]` → `[100e3, 200e3]`), while the sibling `PRISM_ENERGIES` list (renamed `TEST_ENERGIES`) correctly kept all 3. This is disclosed in the PR body ("cutting that class from ~145s to ~92s... verified empirically that 2 energies still yields non-degenerate per-energy active-beam sets, so no coverage is lost") rather than hidden in the mechanical diff — but the empirical verification behind that claim isn't shown in the PR itself, so there's no way to check it from the diff alone. Not a defect, just worth asking for the evidence before taking the "no coverage lost" claim at face value.

## Everything else checked out

Verified directly, not just from the PR's own description:

- **The four claimed bug-fixes are real and correctly fixed.** The `test_measure.py` duplicate (`test_diffraction_patterns_integrated_center_of_mass`, lines 623 and 636 on `dev`) is a byte-identical copy — no coverage lost by removing one. `test_waves.py`'s `exit_plane_waves` fixture genuinely never threaded `device` into the `Potential`/`Probe` it built, so its `[gpu]` cases silently ran the CPU path twice; the fix (indirect parametrize, `device = getattr(request, "param", "cpu")`) is textbook-correct. `test_prism.py::test_prism_scan_match_probe_scan`'s restored assertion is a genuine `EqualityMixin`-based numerical equality check between the PRISM and direct-multislice paths, not a shape-only or trivial one — ran it directly, 10/10 CPU cases pass. `test_core_loss_fixes.py`'s two removed precision tests are genuinely duplicated by `test_core_loss.py`'s equivalents (matching `skipif`-on-GPAW gates on both files, so nothing silently loses coverage when GPAW is absent).
- **`test_gpaw.py`'s fixture scope change (function → module) is genuinely read-only.** Traced every consumer, including the one that calls `.write()` on the shared calculator (GPAW's own `write_gpw()` only logs and serializes) and the one that builds `FrozenPhonons` from it (`abtem/inelastic/phonons.py`'s `randomize()` copies the atoms before displacing). No mutation reaches a later test in the same module.
- **The mechanical dedup itself introduces no regressions**, spot-checked across the largest diffs (`test_array.py`, `test_potentials.py`, `test_prism_upsample.py`, `test_diffraction_pattern_cache.py`, `test_ionization.py`, `test_energy_ensemble.py`): no test body consumes its parametrized `device`/`lazy` value positionally, so the `[gpu, "cpu"]` vs `["cpu", gpu]` order unification can't flip which device a test body actually exercises; parametrize case counts match before and after; a genuinely GPU-only bespoke parametrize (`test_potentials.py::test_finite_projection_gpu_matches_cpu_near_atom_core`) was correctly left alone rather than silently widened to the always-both `devices` decorator; shared-helper extractions (`to_host_array`, `synthetic_transition_potential`, `si_cubic_atoms`/`si_diamond_atoms`) all match argument names, defaults, and bodies against what they replace.
- **The infrastructure pieces around the grouping mechanism are sound**: `_gpu_count()` isn't memoized inconsistently across `gpu`/`requires_gpu`/`requires_multigpu`; no other `xdist_group` marker exists anywhere else in the tree to conflict with; `pyproject.toml`'s new `addopts = "--dist=loadgroup"` doesn't clobber a prior value (there wasn't one) and is a documented no-op without `-n`; CI's new `all-tests-passed` aggregator correctly reflects the fanned-out matrix job's aggregate result, so it can't silently pass while a matrix cell fails.


[durations_gpu.log](https://github.com/user-attachments/files/32399283/durations_gpu.log)
[durations_multigpu.log](https://github.com/user-attachments/files/32399288/durations_multigpu.log)

TomaSusi:
Re: the `BLOCH_ENERGIES` 3→2 cut — fair ask, here's the evidence behind that claim (re-run just now against current `dev`+this branch to confirm it still holds, not just quoted from memory):

```python
import ase
from abtem.bloch.dynamical import BlochWaves

def srtio3_atoms():
    return ase.Atoms(
        symbols="SrTiO3",
        scaled_positions=[
            [0.0, 0.0, 0.0], [0.5, 0.5, 0.5], [0.5, 0.0, 0.5],
            [0.5, 0.5, 0.0], [0.0, 0.5, 0.5],
        ],
        cell=[3.9127, 3.9127, 3.9127],
        pbc=True,
    )

for energies in ([100e3, 200e3, 300e3], [100e3, 200e3]):
    atoms = srtio3_atoms()
    bw = BlochWaves(atoms, energy=energies, sg_max=0.1, g_max=8.0)
    dp = bw.calculate_diffraction_patterns(20.0).compute()
    n_union = int(bw._hkl_mask.sum())
    inactive_counts = []
    for i, sub in enumerate(bw._energy_hkl_masks):
        inactive = ~sub
        inactive_counts.append(int(inactive.sum()))
        assert (dp.array[i, inactive] != 0).sum() == 0
    print(f"energies={energies}: n_union={n_union} inactive_per_energy={inactive_counts}")
```

Output:

```
energies=[100000.0, 200000.0, 300000.0]: n_union=5317 inactive_per_energy=[2944, 2868, 2872]
energies=[100000.0, 200000.0]: n_union=4205 inactive_per_energy=[1832, 1756]
```

So at 2 energies, each still has ~1800 beams inactive out of a ~4200-beam union — nowhere near degenerate (e.g. all-active or all-inactive, which would make `test_inactive_beams_are_zero` vacuous). The property the test checks — that beams inactive at a given energy are correctly zeroed — stays meaningfully exercised with the shorter list; dropping to 2 only removes redundant coverage of "does this generalize to N>2", not the energy-dependent-masking behavior itself. That's the basis for "no coverage lost" in the PR description.

This should really live in a code comment next to `BLOCH_ENERGIES` rather than only in a review reply — will follow up with that.

TomaSusi:
## Adversarial review of the dedup, and one fix pushed

Reviewed this on the premise a test refactor can go wrong in a way that still looks green: by running *less* than before. Measured rather than read — merging current `dev` in locally, the collected-test delta is **4 removed, 0 added**.

Three of the four are a real coverage loss, and I've pushed a fix (`bf3ac9d7`). The rest of what I attacked held up, and one part of it is a clear improvement — details below so the negative results are on record too.

### Fixed: `TestPrecisionConfig` lost the only tests that could catch half its own bug

The dedup removed `test_built_array_honours_precision` (both params) and `test_single_and_double_precision_agree`, leaving only `test_no_hardcoded_dtypes_remain` — a regex over `core_loss.py`'s source.

The class docstring describes a **two-part** defect: allocated `complex64`, *and then* "the closing division by a float64 numpy scalar promoted the whole array back to complex128 under NEP 50". The grep catches part one. It structurally cannot catch part two, because a promotion leaves no dtype literal to match.

Demonstrated rather than asserted — reverting the in-place division at `core_loss.py:936`:

```python
-        array /= np.prod(self.sampling)
+        array = array / np.prod(self.sampling)
```

```
test_built_array_honours_precision : FAILED — assert dtype('complex128') == <class 'numpy.complex64'>
test_no_hardcoded_dtypes_remain    : 1 passed
```

So the class was left documenting a regression it could no longer detect, and the failure mode is silent: every transition potential at double the intended memory, with a green suite.

These also weren't dead weight in practice — GPAW 26.7.0 is installed on this machine and all four passed on `dev` in 9.6 s. Restored in `bf3ac9d7`, with a comment at the restoration site recording why the grep cannot stand in for them, so the next dedup pass doesn't reach the same conclusion.

**Not marked slow**, deliberately: 1.8–4.3 s each, below the ~8.7 s of `test_subshell_transitions_real_gpaw_pipeline` that this branch does mark slow — and CI runs `uv sync --group test`, which doesn't include GPAW (the `gpaw` extra is `hankel`+`sympy`, not the package), so they skip in CI either way. Marking them would only remove them from the workstation runs where they are the actual guard. Full suite after: **1744 passed, 862 skipped, 2 xfailed**.

### Not fixed, correctly removed

`test_gpaw.py::test_charge_density_potential` was a one-line stub that constructed a `GPAWPotential` and asserted nothing — deleting it is right. Worth noting only because, with #421/#422 active on `ChargeDensityPotential` charge-neutrality, a real test there would be worth more than the stub was; that's new work, not something this PR should carry.

### What I attacked and could not break

Recording these because "no finding" is only informative if you know what was actually probed:

- **Silent GPU coverage loss** — gpu-parametrized test IDs are *exactly* preserved, 714 → 714.
- **Loosened tolerances** — every `rtol`/`atol` is byte-identical; the only change on those lines is `_array(...)` → `to_host_array(...)`. One assertion looked dropped (`rtol=0.12`, upsample windowed) but had simply moved.
- **Shrunken constants** — only `BLOCH_ENERGIES` 3→2, already justified above with direct evidence, and the justification has since landed as a code comment next to the constant.
- **`--dist=loadgroup` in `addopts`** — verified genuinely a no-op without `-n`, as the comment claims.
- **The GPU grouping fix** — all six tests @pzeiger identified as previously-missed are now tagged. The 32 `gpu`-named tests that remain untagged are the mock-based ones in `test_multigpu_logic.py` and friends, which carry no GPU requirement; correctly excluded.

### CI gets *more* coverage, not less

Worth stating plainly since "slow-gating" sounds like a reduction: `dev`'s workflow was a bare `pytest test`, so its 287 slow tests **never ran in CI at all**. After this PR they run on every push to `dev` and on PRs into `main`. A PR into `dev` now runs 2322 tests against `dev`'s 2331 — the 4 deletions plus 5 newly-slow — in exchange for slow tests actually executing post-merge. Clear net gain.

### Two things left for you to decide

**1. The grouping detects `requires_gpu` by matching the skip *reason string*.** `_GPU_SKIP_REASONS` compares `mark.kwargs.get("reason")` against the literals `"no gpu"` / `"requires >=2 GPUs and dask-cuda"`. Editing a reason string silently drops a test from the group; the failure mode is VRAM exhaustion under `-n auto` on a GPU box — silent, and invisible on CPU-only machines.

I checked #414 specifically, since it touches exactly this: it adds `reason="no gpu or mps"`, but only on the `gpu` *param* (which is detected by value, not by reason) and leaves `requires_gpu`'s own reason identical — so it does **not** break this today. Left alone rather than changed, since the mechanism was just verified on real hardware and I can't re-verify on a GPU. A dedicated `pytest.mark.gpu` applied by `requires_gpu`, with the reason check kept as a fallback, would remove the coupling without weakening anything.

**2. Two concurrency tests became slow-only:** `TestIntegratorCaches::test_concurrent_access_serves_correct_values` and `TestScatteringFactorCacheKey::test_concurrent_access_does_not_race`. Races are nondeterministic, so running them less often directly lowers the odds of catching an intermittent failure — arguably the category you least want demoted from the pre-merge set. They do still run post-merge on `dev`, so it's a delay rather than a loss; flagging it as a judgment call rather than a defect.

---

TomaSusi:
One more thing surfaced going through `durations_gpu.log`: `test_realspace_multislice.py`'s CPU-only tests look catastrophically inflated relative to a plain workstation baseline, well beyond anything else in the log.

| test | serial CPU baseline (this branch, modest workstation) | this log (`-n auto`, Perlmutter) | inflation |
|---|---|---|---|
| `TestAlgorithmComparison::test_higher_orders_differ` | 4.34s | 290.24s | 67x |
| `TestRealSpaceMultislice::test_realspace_expansion_scope[full-cpu]` | 2.39s | 283.67s | 119x |
| `TestRealSpaceMultislice::test_realspace_max_terms` | 1.57s | 90.69s | 58x |

For comparison, other CPU-bound tests in the same log (`test_gpaw.py`, `test_ionization.py`, `test_atomic_potential.py`) only run 1.1-1.6x slower than that same baseline — ordinary node/hardware variance, not this. It's specific to `test_realspace_multislice.py`.

Checked whether abTEM's own FFT backend could explain it: `Images`/multislice FFTs go through `pyfftw.FFTW(..., threads=config.get("fftw.threads"))`, and `fftw.threads` defaults to `1` in `abtem/core/abtem.yaml` — each FFT call is already single-threaded by design, so this isn't internal FFT-thread oversubscription.

That points instead at process-level oversubscription: `pytest -n auto` sizes its worker pool directly from the node's core count with no cap, and Perlmutter's CPU nodes have a lot of cores. If a large number of xdist worker processes land this file's memory-bandwidth-heavy real-space stencil/FFT work concurrently, cache and memory-bandwidth contention alone (independent of any single process's own thread count) could produce a blowup like this. My earlier comment that "`-n auto` is now safe" only accounted for the GPU/VRAM angle this PR's `xdist_group("gpu")` mechanism addresses — it didn't consider CPU memory-bandwidth contention at HPC scale, which is a separate failure mode this PR does nothing to guard against.

Two things that would confirm or rule this out: what `-n` value did this run actually use (i.e. what did `-n auto` resolve to on that node), and was the node exclusively allocated to this job or shared with other users' work at the time? If it's oversubscription, a capped `-n` (rather than `auto`) for HPC-scale runs would be the fix; if the node was shared/contended, this may just be noise from that run and not indicative of anything in the code or this PR.

TomaSusi:
Verified `45257bb6`. Nothing further from my side on this branch.

The shape of the change is the right one, and better than what I suggested: the dedicated `gpu` marker is *preferred*, with the reason-string match kept as a fallback for anything not going through `requires_gpu`/`requires_multigpu`. That makes the change purely additive — group membership can only grow, never shrink — so it cannot reintroduce the silent VRAM-exhaustion failure mode by accident.

Confirmed it is behaviour-preserving rather than merely intended to be: collected the `xdist_group`-tagged set at `bf3ac9d7` and at `45257bb6` and diffed them.

```
bf3ac9d7: 699 tagged
45257bb6: 699 tagged
lost from the group: (none)
newly added:         (none)
```

Identical membership, which is exactly what a robustness refactor should produce. The `_GpuRequirement` class rather than attributes bolted onto a function is also the right call for keeping `.marks`/`.mark` typed.

Nothing outstanding from my side — both the restored precision tests and the marker change are settled.

---

pzeiger:
The `test_realspace_multislice.py` slowdown under `pytest -n auto` traces to a real defect in abTEM itself, not to this branch's diff — it reproduces identically on `dev`.

Every `@njit(parallel=True)` kernel (the CPU Laplacian stencil in `finite_difference.py`, the magnetism kernels in `magnetism/pauli.py`, the partitioned S-matrix kernel in `prism/_partitioned_s_matrix.py`) uses Numba's own thread pool, which defaults to the full visible CPU count and reads neither `OMP_NUM_THREADS` nor `OPENBLAS_NUM_THREADS`. Several such processes sharing one node each independently try to claim every core for themselves — exactly what `pytest -n auto` produces. Measured on a Perlmutter GPU node with `OMP_NUM_THREADS=1`/`OPENBLAS_NUM_THREADS=1` already set: the affected `[cpu]`-parametrized tests took 100-400s each versus 2-8s for their `[gpu]`-parametrized counterparts (the GPU path uses a separate CUDA kernel and never touches Numba, so it was never affected).

Fix and full details: #441

pzeiger (approved):

PR #428, opened 2026-09-18
Fix FieldIntegrator.integrate_on_grid's abstract signature and docstring

pzeiger:
## What

`FieldIntegrator.integrate_on_grid`'s abstract signature names its first argument `positions`, documented as a 2D array of xy-positions, and types `a`/`b` as `numpy.ndarray`. Every concrete implementation — `GaussianProjectionIntegrals`, `ScatteringFactorProjectionIntegrals`, `QuadratureProjectionIntegrals` — takes an `ase.Atoms` for the first argument instead, and the only caller (`iam.py`) passes one positionally along with scalar floats for `a`/`b` (`sliced_atoms.slice_limits[slice_idx][0]`/`[1]`). `QuadratureProjectionIntegrals` alone had already corrected `a`/`b` to `float`; the abstract base and the other two concrete implementations still declared `numpy.ndarray`. Anyone implementing a custom integrator against the documented interface, which is the reason the abstract class exists, writes it for the wrong argument types and finds out at runtime.

Renamed the abstract parameter to `atoms`, corrected `a`/`b` to `float` everywhere, and rewrote the docstring to match — one consistent signature across the abstract base and all three concrete implementations. There are no keyword callers of the first argument anywhere in the tree, so this is not a behavioural change.

## Blast radius

Grepped every `.integrate_on_grid(` call site in the tree: `iam.py` (the only production caller, positional) and two test sites (`test/test_atomic_potential.py:615,1259`, both positional). None pass the first argument by keyword, so renaming it breaks nothing. Type annotations are not enforced at runtime, so correcting `a`/`b` is a documentation-only change with no behavioural effect either.

## Tests

`TestFieldIntegratorSignature` in `test/test_atomic_potential.py` compares the abstract method's full parameter signature — names and annotations — against each concrete implementation's, so the two cannot drift apart again on either axis. Comparing names alone would have missed the `a`/`b` type mismatch, since `QuadratureProjectionIntegrals` already had the right name there with the wrong annotation elsewhere; comparing full parameter objects (`inspect.signature(...).parameters.values()`) catches both. Fails on unfixed `dev` (`positions` vs `atoms`, and `numpy.ndarray` vs `float` on `a`/`b`, both confirmed) and passes on this branch.


TomaSusi (approved):
**Approving** — the change is correct and safe. `Atoms` is already imported at module level (`abtem/integrals.py:13`), all three concrete implementations already name the parameter `atoms`, and the only production caller passes it positionally (`abtem/potentials/iam.py:1012`). `test/test_atomic_potential.py`: 79 passed, 4 skipped on this branch.

Two follow-ups, both in the docstring this PR is already editing.

### 1. The same defect survives in `a` and `b`

The sole caller passes **scalar** slice boundaries:

```python
new_array = self._integrator.integrate_on_grid(
    atoms,
    a=sliced_atoms.slice_limits[slice_idx][0],
    b=sliced_atoms.slice_limits[slice_idx][1],
    ...
)
```

but the abstract signature still declares `a: np.ndarray` / `b: np.ndarray` and documents them as "...along z for each position". `QuadratureProjectionIntegrals.integrate_on_grid` annotates `a: float, b: float`, directly contradicting the base. That is exactly the "written against the documented interface, finds out at runtime" failure this PR exists to fix, left in place two lines below the fix. `TestFieldIntegratorSignature` only compares parameter *names*, so it cannot catch it.

### 2. Pre-existing typo inside the edited paragraph

```
Lower integration limit of the pr
ojection integrals along z for each position [Å].
```

Worth the drive-by while you are in here.

Neither blocks the merge.

---

pzeiger:
Both follow-ups fixed in `5b22c7dc`.

**1. `a`/`b` annotations.** Confirmed: the abstract base and two of the three concrete implementations (`GaussianProjectionIntegrals`, `ScatteringFactorProjectionIntegrals`) still typed `a`/`b` as `numpy.ndarray` while the only caller (`iam.py`) always passes scalar floats (`sliced_atoms.slice_limits[slice_idx][0]`/`[1]`), and `QuadratureProjectionIntegrals` alone had it right. All three now agree with the caller and with each other — `float` throughout, matching the corrected docstring.

`TestFieldIntegratorSignature` compared parameter *names* only, so it couldn't catch this. Strengthened it to compare full parameter objects (names and annotations together, via `inspect.signature(...).parameters.values()`) across the abstract base and all three concrete classes. Verified directly: reverting the annotation fix makes the strengthened test fail (`'np.ndarray' != 'float'`); it passes again with the fix restored.

**2. The typo.** Already gone — fixing it fell out of rewriting that docstring paragraph for the `a`/`b` type change, since the paragraph doesn't wrap across that line boundary anymore.

`test/test_atomic_potential.py -k "not gpu"`: 82 passed, 1 skipped (up from 79/1 pre-fix, from the added parametrized cases).

TomaSusi (approved):
Verified `5b22c7dc` — both follow-ups are fully addressed.

`a`/`b` are now `float` on the abstract base and all three concrete implementations, matching the scalar slice limits the only caller actually passes. The `pr\nojection` line-break typo is gone, and `TestFieldIntegratorSignature` now compares full parameter objects — names *and* annotations — so this class of drift can't recur silently. `test/test_atomic_potential.py`: 82 passed, 4 skipped on `5b22c7dc`.

Approval stands.

---

PR #429, opened 2026-09-18
Memoize the PRISM-EELS transition potential's pure-delayed graph node

pzeiger:
## What

`SMatrix.transition_potential_scan`'s lazy branch wraps the shared transition potential with `dask.delayed(transition_potentials, pure=True)` once per call, so `map_blocks` can hand every ensemble block the same graph node instead of a copy embedded per task. `pure=True` tokenizes the object via a content hash — dask's `_normalize_pickle` fallback for objects with no `__dask_tokenize__` — which costs ~0.4 ms/MB of the underlying array, dominated by hashing the array's out-of-band buffers rather than the pickle round-trip itself (protocol 5 passes the payload zero-copy). A single scan pays this once and does not notice; a frozen-phonon or energy sweep that calls this method many times against the same live `transition_potentials` object re-tokenizes the identical payload on every call.

`BaseTransitionPotential._as_pure_delayed()` memoizes the `Delayed` wrapper as an attribute of the object it wraps, so a caller reusing the same object across many graph builds pays the content hash once.

## Why identity, not content, and the invariant this relies on

The two candidate designs trade off differently:

- **A cached content hash** (`__dask_tokenize__` computing and storing a hash on first use) pays the same cost the first time — the hash *is* the cost — and a memoized hash is a silent-wrong-answer risk on a mutable array: if the payload is ever mutated in place, a stale hash would let two different objects share a graph key, and dask would reuse the wrong node.
- **An identity-keyed memo of the `Delayed` object itself**, what this PR does, sidesteps the id()-reuse problem a bare `id()`-based token would have (ids are reused after garbage collection, and using one to build a token breaks determinism across processes): there is exactly one memoized `Delayed` per live object, and it is dropped before the object is ever pickled (see below), so a stale copy can never outlive the object it was built from.

That does not make the memo immune to staleness, though: the cached `Delayed`'s dask *key* is still a content hash, frozen at the first call, while the node itself keeps a live reference to `self`. Mutating the wrapped object's payload in place after that call leaves the memoized key describing stale content even though the node would compute against whatever `self` currently holds — the same hazard a cached content hash would have, just moved one level up rather than eliminated. Nothing under `abtem/` mutates a transition potential's payload in place after handing it to `transition_potential_scan` today, so the risk is latent rather than live, and `_as_pure_delayed`'s docstring now states this invariant explicitly rather than claiming the memo cannot go stale.

## A pickling trap this needed to avoid

The memoized `Delayed` wraps `self`. Letting it ride through pickle verbatim would embed a self-referential graph node inside the object's own pickle — at best doubling the payload for no benefit, at worst reproducing the exact cost this PR removes. `BaseTransitionPotential.__getstate__` resets it to `None`; `TransitionPotentialArray.__getstate__` (added in #425, which does not call `super()`) needed the same reset repeated rather than inherited. `TransitionPotentialArray._eq_exclude` also gained the new attribute: comparing two `Delayed` objects via `==` builds another lazy value rather than a bool, which is the "deferred content is guessed equal" trap `safe_equality` was fixed elsewhere to avoid — this cache must be excluded from comparison, not merely reset.

`copy.copy` honours `__getstate__` by default, so `_task_local()` views normally get the memo blanked along with everything else `__getstate__` resets — except `TransitionPotentialArray`, whose own `__copy__` shares `__dict__` by reference instead (added in #425, to fix an unrelated per-task-view cache-sharing bug), so its `_task_local()` views share the memoized `Delayed` with their source rather than getting it blanked. That is inert today: `_as_pure_delayed` is only ever called on the object handed directly to a scan method, never on a `_task_local`/`copy_to_device` view, and pickling any view still blanks the memo via `__getstate__` before it reaches a worker. `_task_local`'s docstring now documents this exception explicitly.

## Blast radius

`dask.delayed(..., pure=True)` has exactly one call site in the tree (`abtem/prism/s_matrix.py`), the one this PR changes; the corresponding multislice mechanism (`shared_constant_arg`, used via `_GRAPH_NODE_KWARGS` in `multislice.py`) wraps a *function call* rather than a bare value and is unaffected. `BaseTransitionPotential.__getstate__` is new on the shared base, inherited by both concrete subclasses (`TransitionPotentialArray`, and the lighter-weight `TransitionPotential`, which holds no array and was already cheap to tokenize but benefits from the same correctness properties). `TransitionPotential` has no `__copy__` of its own, so its `_task_local()` copies now route through this `__getstate__` where they previously used a plain shallow copy; the only observable difference is that the new attribute is reset on the copy, which is inert there since nothing calls `_as_pure_delayed()` on a per-task view.

## Tests

`TestDelayedTransitionPotentialMemo` in `test/test_transition_potential_transport.py`: repeated calls return the same `Delayed` object; distinct objects get distinct nodes; a memoized call does not invoke `dask.delayed` again (checked directly via monkeypatch); a PRISM scan run twice against the same object gives the same result; pickling drops the memoized node and a fresh call still works afterward; two content-equal objects compare equal regardless of which one has a populated memo; a bare `TransitionPotential`'s `_task_local()` copy still shares everything the mechanism doesn't touch, while an array `TransitionPotentialArray`'s `_task_local()` copy shares the memo itself rather than getting it blanked (pinning down the `__copy__` exception above); and the memoized key is shown going stale under an in-place mutation of the payload (pinning down the documented invariant, not a defect to fix). Eight of the nine fail on unfixed `dev` with `AttributeError` for the method this branch adds; the result-only test cannot fail on unfixed `dev` by construction (correctness was never broken) and passes on both, the same shape as other tests in this family that guard the new mechanism rather than a defect.


TomaSusi (changes requested):
**Requesting changes** — the mechanism works (19/19 pass in `test/test_transition_potential_transport.py`), but the design rationale in the description is factually wrong, and it is the part a future reader will lean on.

### The memo does go stale under mutation

From "Why identity, not content":

> It cannot go stale under mutation, since it does not summarize content at all

It does summarize content. `dask.delayed(x, pure=True)` bakes a content hash into the **key** at call time, and memoizing the `Delayed` freezes that key. On this branch:

```python
k1     = tp._as_pure_delayed().key
tp.array[0, 0, 0] = 123456.0          # in-place mutation of the payload
k_memo = tp._as_pure_delayed().key
k_fresh = dask.delayed(tp, pure=True).key
```

```
before mutation : TransitionPotentialArray-fd2384a4dc934f068ab11e79313824d5
memoized after  : TransitionPotentialArray-fd2384a4dc934f068ab11e79313824d5   <- stale
fresh after     : TransitionPotentialArray-4ddcf8e77d90f6d6505eabc98604e8f9
```

So the PR trades one staleness risk for the same staleness risk, having argued it eliminated it — the exact hazard the description rules the cached-content-hash alternative out for.

The exposure is narrow but real, and it is the `distributed` case: graph A submits key1 carrying the pre-mutation payload; after an in-place mutation, graph B reuses key1, the scheduler sees a key it already holds, never ships the new literal, and silently computes against stale data. Without the memo the key changes and the result is correct. Local and threaded schedulers are unaffected, since there the graph literal is the live object.

I do not think this blocks the optimization itself — nothing in `abtem/` mutates the payload buffer in place today, and `_task_local()` plus `TransitionPotentialArray.__copy__` keep `scatter`'s `self._array` rebind off the shared object. But the invariant has to be *stated* rather than denied. Either:

- document it (`_as_pure_delayed` is valid only while the payload is not mutated in place), or
- invalidate the memo wherever `_array` is rebound.

As written, the description would mislead whoever relies on it next.

### Smaller: "dropped on every copy" is not true for the array subclass

`_as_pure_delayed`'s docstring says the memo "is dropped before this object is ever pickled (see `__getstate__`)" — true for pickling. But `TransitionPotentialArray` defines its own `__copy__` doing a raw `new.__dict__.update(self.__dict__)`, which bypasses `__getstate__`, so a `_task_local()` view of an array TP **shares** a `Delayed` wrapping its parent. Inert in practice — the view is discarded, and `__getstate__` still blanks it before the view is pickled into a task — but `_task_local`'s own docstring now reads wrong for that subclass:

> Note that `copy.copy` honours `__getstate__`, so a subclass that blanks an attribute there gets it blanked in this view as well.

`test_task_local_copy_of_a_bare_transition_potential_still_shares_state` asserts the blanking only for the bare class, which is the one where it actually holds.

### Checked and fine

`_eq_exclude` on the bare `TransitionPotential`: `BaseTransitionPotential` is not in `EqualityMixin`'s MRO, so it falls back to identity equality and needs no exclusion. No gap there — adding the attribute to `TransitionPotentialArray._eq_exclude` alone is correct.

### Missing issue file

The comment in `_as_pure_delayed` cites `delayed_transition_potential_rehashes_payload.md`, which is not in the repo — the tree has three `.md` files total, and no existing comment under `abtem/` references a `.md` file at all, so this and the matching one in #430 would be the first, pointing at something a reader cannot open. Either commit the issue files or inline the numbers they carry.

---

pzeiger:
All three points addressed in `3fc66025`.

**The staleness claim.** Confirmed by direct reproduction: tokenize, mutate `tp.array` in place, tokenize again — the memoized key matches the pre-mutation content and disagrees with a fresh `dask.delayed(tp, pure=True)` call. `_as_pure_delayed`'s docstring no longer claims the memo "cannot go stale under mutation"; it states the real invariant instead — identity-keying avoids the `id()`-reuse and pickling hazards, but the cached `Delayed`'s dask *key* is still a content hash frozen at first call, so it goes stale under the same in-place-mutation hazard a cached content hash would have, just one level up. Nothing under `abtem/` mutates a transition potential's payload in place after handing it to `transition_potential_scan` today, so this is documented as latent risk, not a live one — not something this PR needs to close off, per the two options given, but something the docstring shouldn't deny.

**The `__copy__`/`_task_local` interaction.** Confirmed: `TransitionPotentialArray.__copy__` shares `__dict__` by reference rather than routing through `__getstate__`, so a `_task_local()` view of an array transition potential shares `_delayed_pure_node` with its source instead of getting it blanked — unlike the bare `TransitionPotential`, which has no `__copy__` override and does get it blanked. `_task_local`'s docstring now states this exception explicitly rather than the unqualified "copy.copy honours `__getstate__`" claim. Agreed this is inert today: `_as_pure_delayed` is only ever called on the object handed directly to `transition_potential_scan`, never on a `_task_local`/`copy_to_device` view, and pickling any view still blanks the memo via `__getstate__` before it reaches a worker.

**Missing issue file.** Removed the dangling `.md` reference in this PR and in #430 — it didn't resolve to anything in this repository. The reasoning and numbers it would have pointed to were already stated inline in the same comment, so nothing is lost.

Two tests added to `TestDelayedTransitionPotentialMemo`: one reproduces the memoized key going stale under in-place mutation directly, the other confirms an array transition potential's `_task_local()` view shares the memo object rather than getting it blanked. `test/test_transition_potential_transport.py`: 21 passed (up from 19 pre-fix).

TomaSusi (approved):
Verified `3fc66025` — this clears my requested changes.

`_as_pure_delayed`'s docstring now states the real invariant rather than denying it: the cached `Delayed`'s key is frozen from the content tokenized at the first call while the node still wraps a live `self`, so it carries the same in-place-mutation hazard a cached content hash would, moved one level up rather than eliminated — latent today because nothing under `abtem/` mutates a transition potential's payload after handing it to `transition_potential_scan`. That's exactly the right resolution: documenting the invariant rather than closing off a risk that isn't live.

`_task_local`'s docstring now carries the `TransitionPotentialArray.__copy__` exception explicitly instead of the unqualified "`copy.copy` honours `__getstate__`" claim, and the two new tests pin down both behaviours — the stale key under mutation, and the array subclass's view sharing the memo by reference. `test/test_transition_potential_transport.py`: 21 passed.

Thanks also for the note on the `.md` references — that abTEM's issue tracking lives in a separate private repo explains why they couldn't resolve here, and dropping them is the right call given the numbers were already inline.

---

PR #430, opened 2026-09-18
Bucket the core-loss scatter batch size to a small repeating set

pzeiger:
## What

`TransitionPotentialArray.generate_scattered_waves` chunks the filtered site list, and `scatter`'s `ifft2` batches over `len(sites_chunk)` — the surviving-site count after threshold filtering (the transitions axis is separate and un-folded, so a chunk of 11 sites with 2 transitions batches as `(11, 2, gy, gx)`, not `(22, gy, gx)`), which varies per slice and per scan region. Every distinct count is a new cuFFT plan shape: the set of distinct plans grows with the scan, pressuring any fixed plan-cache entry limit (#378 raised it to a configurable 64, which a larger scan can re-cross), and every first occurrence of a shape pays full planning cost regardless of cache size.

Each real site chunk is now padded up to the next multiple of 8 (`_SCATTER_BATCH_BUCKET`, via the new `abtem.core.chunks._ceil_to_multiple`) before being handed to `scatter`, so the `ifft2` batch shape depends on the bucket rather than the data-dependent site count. Padding uses a repeat of the chunk's own last real site — any real position works, since the padded results sit at the end of the flattened `(site, transition)` batch axis and are sliced away before anything downstream sees them. The chunk `generate_scattered_waves` yields alongside the result is always the real, unpadded site count; nothing about the public shape changes.

Padding is capped at the real per-chunk site ceiling `max_batch`/the VRAM budget allows (`site_ceiling`, recovered by probing `validate_chunks` with a sentinel axis length far larger than any real chunk size, so it reflects the ceiling rather than however many sites the current call happens to have). `validate_chunks` already sizes each chunk against that budget (`max_elements=limit`); padding past it would multiply a small, memory-constrained chunk's footprint by up to `_SCATTER_BATCH_BUCKET` — exactly the allocation the budget exists to bound. The common case (a generous or default `max_batch`) is unaffected, since that ceiling is never approached; a tight explicit `max_batch` now pads only up to what it already allowed.

## Why 8, and why not `_nearest_power_of_two`

- **A fixed-stride bucket beats power-of-two bucketing.** Next multiple of 8 gives ~10.3% average padding slack with 32 distinct shapes for a 256-site `max_batch` — comfortably inside the 64-entry plan cache — against 34–38% for power-of-two bucketing on the same measurement. That 10% figure is for a chunk comfortably above the bucket size; the `site_ceiling` cap above is what keeps a small, budget-limited chunk from paying a much larger relative overhead.
- **`_nearest_power_of_two` (`core/chunks.py`), named as in-house precedent, rounds *down*** when the upper power would overshoot by more than 25% (by design — its only caller sizes a batch against a VRAM budget, where dropping is safe). Reusing it here would silently drop sites, turning a performance issue into a silent-wrong-answer. `_ceil_to_multiple` is a new, separate helper that only ever rounds up.

## Blast radius

The padding and its budget cap live entirely inside `generate_scattered_waves`; `scatter()` itself is unchanged and still used exactly as before by its other caller (`TransitionPotential.scatter`, an unchunked single call over the whole site set, not implicated by this issue). `_ceil_to_multiple` is a new function with no other callers. `site_ceiling` costs nothing extra when `max_batch` is an explicit int (`validate_chunks` already tiles it verbatim regardless of the VRAM budget, so the ceiling is just `max_batch`); the `"auto"` case needs an extra `validate_chunks` call, probing with a sentinel axis length instead of the real site count — memoized in `_get_site_ceiling`, keyed on `(max_batch, waves.shape, dtype, device)`, since `generate_scattered_waves` runs once per potential slice against the same waves shape throughout one scan. Unmemoized, this cost 3–6 ms per call measured inside that real per-slice loop (an isolated call to `validate_chunks` alone understated it badly), against 0.3 ms/call for the int path that never probes — an order-of-magnitude regression on the `"auto"` path that the sentinel size (`_SCATTER_BATCH_CEILING_PROBE_SITES = 2**24`) barely affects either way; memoized, `"auto"` costs the same as the int path.

At the shipped production default (`transition_potential_multislice_and_detect`'s `scatter_max_batch=1`), `site_ceiling` is always `1` and padding never engages — every chunk is already forced to exactly one site before padding is ever considered, so there is no shape variability left to remove at that default. Expected, not a gap; a before/after benchmark of this PR needs a larger `scatter_max_batch` to measure anything.

## Tests

`TestScatterBatchBucketing` in `test/test_core_loss_fixes.py`: the actual `ifft2` batch shapes (monkeypatched to record, not inferred from the yielded chunk size) collapse from six distinct values, the site counts themselves (`{3, 5, 7, 11, 17, 23}`, confirmed by recording the same shapes against unpatched `dev`), onto three, `{8, 16, 24}`; a parametrized sweep across and exactly on bucket boundaries (1 through 25, including 8/16/24) matches a direct, unchunked, unpadded `scatter()` call bit-for-bit; a multi-chunk scan (37 sites, `max_batch=6`) reassembles to the same result; a small explicit `max_batch=2` (below the bucket) never pads past it, confirmed against the actual recorded `ifft2` batch shapes and still bit-exact against a direct call; the ceiling probe runs at most once across five repeated calls at the same waves shape; the cache correctly distinguishes two different waves shapes; and an explicit int `max_batch` never touches the cache. `TestCeilToMultiple` checks the new helper never rounds down, contrasting it directly against `_nearest_power_of_two`'s documented behaviour. The shape-collapse, budget-cap, and memoization tests each fail against their respective unfixed code (reproduced all three); the remaining correctness tests cannot fail on unfixed `dev` by construction, since correctness was never broken — they guard the new padding/slicing logic itself, the same shape as other tests in this family that guard a new mechanism rather than a defect.


TomaSusi (commented):
**Comment, not a block** — 24/24 pass and the padding/slicing logic is sound. I verified the three things it rests on: `scatter` reshapes site-major (`(len(sites), len(self), ...)`, then flattens), so `[: n_real * len(self)]` drops exactly the padding; `validate_sites` ends in `np.array(sites, dtype=get_dtype())`, forcing host numpy, so the `np.repeat`/`np.concatenate` can never hit a CuPy array; and the `threshold` interaction is safe, since `generate_scattered_waves` pre-filters and resets `threshold=None`, so `scatter`'s own `filter_sites` cannot drop a padded site.

Two things to fix.

### 1. The test's baseline claim is wrong

`test_ifft2_batch_shapes_are_bucketed` records `x.shape[0]`, which is the **site** count, not `n_sites * len(self)`. So this comment is not describing what the test measures:

> Unbucketed, this reproduces the issue file's own numbers exactly: `{6, 10, 14, 22, 34, 46}`

Running that same recorder against `dev`:

```
shape[0] on dev: [3, 5, 7, 11, 17, 23]
full shapes: [(3, 2, 32, 32), (5, 2, 32, 32), (7, 2, 32, 32), (11, 2, 32, 32), (17, 2, 32, 32), (23, 2, 32, 32)]
```

The assertion itself (`{8, 16, 24}`) is right and the six-to-three collapse is real, so the test is valid — but the comment and the PR description both misdescribe what is measured, and they are internally inconsistent: if the recorded values really were the ×2 products, the bucketed set would be `{16, 32, 48}`, not `{8, 16, 24}`. Worth correcting, since it is the test's own statement of what it reproduces.

### 2. Padding escapes the memory budget

`validate_chunks(..., max_elements=limit)` sizes each chunk against a VRAM target, and the padding is applied *after* that, adding up to 7 extra sites' worth of `(len(self), gy, gx)` complex to the `ifft2` intermediate. At the 256-site `max_batch` the description benchmarks, that is the quoted ~10% and entirely fine. On a memory-tight run where auto-chunking lands at 1–3 sites per chunk, it is a 3–8× blowup of exactly the allocation the budget exists to bound — on the same GPUs this PR is meant to help.

Cheap fix either way: round the validated chunk size *down* to a multiple of `_SCATTER_BATCH_BUCKET` so only the remainder chunk ever pads (which also removes almost all of the padding waste), or subtract one bucket's worth from `limit` before calling `validate_chunks`.

### Missing issue file

`_ceil_to_multiple`'s docstring and the `_SCATTER_BATCH_BUCKET` comment cite `scatter_batch_sizes_unbounded_fft_shapes.md`, which is not in the repo — the tree has three `.md` files total, and no existing comment under `abtem/` references a `.md` file at all, so this and the matching one in #429 would be the first, pointing at something a reader cannot open. Either commit the issue files or inline the numbers they carry.

---

pzeiger:
Both points addressed, in `9fa1445c` and `05621feb`.

**1. The test's own numbers.** Confirmed by recording actual `ifft2` call shapes against unpatched `dev`: `ifft2`'s batch axis is the site axis alone — the transitions axis (length `n_transitions=2` in the test) is separate and un-folded, e.g. `(11, 2, 32, 32)` for 11 sites, not `(22, 32, 32)`. So `x.shape[0]` really is just the site count, and the unbucketed values are `{3, 5, 7, 11, 17, 23}`, not `{6, 10, 14, 22, 34, 46}`. The comment and PR description both had this wrong; both are corrected now. The bucketed assertion (`{8, 16, 24}`) was already right — only the description of what it's contrasted against was off, and (as flagged) the two were internally inconsistent: doubled values bucketed to 8 would collapse onto `{8, 16, 24, 40, 48}`, not `{8, 16, 24}`.

**2. Padding past the memory budget.** Fixed rather than left as a known gap. Padding is now capped at `site_ceiling` — the real per-chunk site count `max_batch`/the VRAM budget allows, independent of how many sites the current call happens to have. `validate_chunks` already sizes each chunk against that budget (`max_elements=limit`); capping padding at the same ceiling means a small, memory-constrained chunk (1–3 sites) never pads past what it was already allowed, while the common case (a generous or default `max_batch`) is unaffected, since that ceiling is never approached. New test: an explicit `max_batch=2` (below the bucket) with 9 real sites — confirmed the actual recorded `ifft2` batch shapes never exceed 2, and the result still matches a direct, unpadded call bit-for-bit.

`site_ceiling` costs nothing extra for an explicit int `max_batch` (`validate_chunks` already tiles it verbatim regardless of the VRAM budget, so the ceiling is just `max_batch`); the `"auto"` case recovers it by probing `validate_chunks` with a sentinel axis length (`2**24`, already a multi-hundred-GB chunk at any real `gpts`/dtype) rather than the real site count, so a call whose site count already fits in one chunk — the common case — doesn't get its own count mistaken for the ceiling.

**Missing issue file.** Removed the dangling `.md` reference here too — it didn't resolve to anything in this repository.

`test/test_core_loss_fixes.py -k "ScatterBatchBucketing or CeilToMultiple"`: 25 passed (up from 24 pre-fix).

TomaSusi (commented):
Verified `9fa1445c`/`05621feb`. Both original points are fixed, and the cap is better than what I suggested — it *fails safe*: when `site_ceiling < n_real` the `if n_padded > n_real` guard skips padding entirely, so truncation is structurally impossible rather than merely ruled out by the invariant in the comment. Confirmed bit-exact against a direct unpadded `scatter()` at `max_batch` 1/2/3/6, with the recorded `ifft2` batch never exceeding the cap. `test/test_core_loss_fixes.py test/test_site_filtering.py`: 141 passed, 1 skipped.

One new problem, though, introduced by the fix.

### The ceiling probe is a ~10x regression on the `"auto"` path

`generate_scattered_waves` is called **once per potential slice** (`abtem/multislice.py:1175`), and the probe runs on every one of those calls:

```
                  dev        branch
max_batch='auto'  0.59 ms →  6.38 ms per call   (~10.8x)
max_batch=64      0.56 ms →  0.76 ms per call
```

Instrumenting `validate_chunks` inside a real call shows all of it is the probe — `shape[0]=16777216` costs **5.39 ms**, against 0.07 ms for the real chunking call beside it.

The cost figures in `05621feb`'s message don't reproduce here. Measured warm, and `validate_chunks` has no memoization, so every call pays full freight:

| sentinel | cost | chunk tuple |
|---|---|---|
| `2**24` (shipped) | 5.46 ms | 1074 entries |
| `10**9` (original) | 6.49 ms | 64000 entries |

That is ~1.2x, not 4x, and not "<1 ms". Worth noting *why* the optimization underdelivered: the tuple shrank 59x while the cost barely moved, so tuple construction was never the bottleneck — something else inside `validate_chunks` scales with the sentinel axis length. A cheaper sentinel won't fix this.

**Scope.** Narrower than the number suggests: `transition_potential_multislice_and_detect` defaults `scatter_max_batch=1`, an int, so the production path takes the `isinstance(max_batch, int)` shortcut and never probes. It bites callers passing `scatter_max_batch="auto"` explicitly (reachable through `**multislice_func_kwargs` on the scan methods) and anyone calling `generate_scattered_waves` directly — whose own default *is* `"auto"`.

**Suggested fix.** The probe's result is a pure function of `(max_batch, waves.shape, dtype, device, limit)`, none of which vary from slice to slice within a scan. Memoizing it — or hoisting it out of the per-slice call into the driver — removes the regression outright, and is a better lever than tuning the sentinel further.

### Separately: the bucketing is a no-op at the shipped default

With `scatter_max_batch=1`, `site_ceiling=1`, so `n_padded = min(_ceil_to_multiple(1, 8), 1) = 1` and padding never engages. That is correct behaviour — the cap is doing its job, and #378's plan-cache pressure only arises at larger batches anyway — but it does mean any before/after benchmark of this PR has to raise `scatter_max_batch`, or it measures nothing at all. Flagging it so the default isn't mistaken for a representative configuration.

Not blocking from my side; the correctness work here is solid.

---

pzeiger:
Reproduced the regression before touching anything, independently: `generate_scattered_waves` in a real per-slice driver loop, `max_batch="auto"` costs 3.71 ms/call against 0.30 ms/call for an explicit int here — 12.5x, the same order of magnitude as the 10.8x measured on your side. The isolated `validate_chunks` microbenchmark in `05621feb`'s message was the wrong instrument: it measured the function alone and missed whatever inside `validate_chunks` actually scales with the sentinel axis length. Confirmed that directly — same real-loop harness, `9fa1445c`'s `10**9` sentinel against `05621feb`'s `2**24`: 4.80 ms/call vs 3.71 ms/call, 1.29x, matching your ~1.2x and nowhere near the 4x the isolated benchmark suggested.

Fixed in `91f53364` by memoizing, per your suggestion, rather than tuning the sentinel further. `_get_site_ceiling` caches the probe's result on the object, keyed by `(max_batch, waves.shape, dtype, device)`, all fixed for the lifetime of one multislice run reusing the same `transition_potential`. An explicit int `max_batch` still returns immediately without touching the cache. Memoized `"auto"` now costs 0.34 ms/call against 0.31 ms/call for the int path — statistically the same, regression closed. Three new tests: the probe runs at most once across five repeated calls at the same waves shape, the cache correctly distinguishes two different waves shapes (and produces a physically sensible smaller ceiling for a larger grid at the same budget), and the int path never touches the cache.

On the shipped default: confirmed `scatter_max_batch=1` in `transition_potential_multislice_and_detect`, and that `site_ceiling=1` there means every chunk is already forced to exactly one site by `validate_chunks` itself before padding is ever considered — there is no shape variability left for bucketing to remove at that default, so it doing nothing there is expected, not a gap. Noted for anyone benchmarking this against `dev`.

`test/test_core_loss_fixes.py`: 142 passed.

TomaSusi (approved):
Verified `91f53364` — the regression is closed. Approving.

Measured on the new head, same driver-loop benchmark as before:

```
                  dev     previous    now
max_batch='auto'  0.59 ->   6.38   ->  0.87 ms/call
max_batch=64      0.56 ->   0.76   ->  0.80 ms/call
```

and the probe now runs **once across 50 per-slice calls** rather than on every one, with the cache landing as `{('auto', (32, 32), complex64, 'cpu'): 15625}`. The `"auto"` path is now statistically indistinguishable from the int path, which is the right outcome.

Two things about the cache key that I checked rather than assumed:

- `limit` isn't part of the key, but it doesn't need to be: the int branch returns before touching the cache, so the memoized branch only ever sees `limit == max_batch`, which *is* in the key. Sound as written — worth keeping the early return and the key in sync if either changes.
- `self.device` being in the key is what makes it safe to carry `_site_ceiling_cache` through `__getstate__` (unlike the two device-array caches): a pickle built on one device can't produce a stale hit on a worker with another. The reasoning in the commit message for not resetting it holds.

`test/test_core_loss_fixes.py test/test_site_filtering.py`: 144 passed, 1 skipped.

---

TomaSusi:
Heads-up: I pushed a merge of `dev` into this branch to clear the conflict (`91f53364..1ffb69d9`, a fast-forward — your head was untouched). Flagging the resolution since it lands in a comment block that now spans two PRs' reasoning.

**Cause.** #429 merged to `dev` as `52e63e06`, and both PRs add an entry to `TransitionPotentialArray._eq_exclude`: `_delayed_pure_node` from #429, `_site_ceiling_cache` from this branch. That was the only conflict hunk — `__getstate__`, `__init__` and `test_core_loss_fixes.py` all auto-merged.

**Resolution.** Kept both entries and both comment fragments:

```python
    _eq_exclude = (
        "_local_potential_device_cache",
        "_device_array_cache",
        "_delayed_pure_node",
        "_site_ceiling_cache",
    )
```

Taking either side wholesale would have silently dropped the other's exclusion — and the symptom would have been a `TransitionPotentialArray` that stops comparing equal to an identical one the moment the dropped cache is populated, which no existing test would necessarily catch in isolation.

The merged comment still reads correctly without rewording: "``__getstate__`` already drops all three" continues to refer to the three attributes named before it, and the `_site_ceiling_cache` sentence immediately after carves it out as the deliberate exception (small ints, safe to pickle, so `__getstate__` leaves it alone). Worth a second look if you'd rather phrase that differently now that four attributes are listed but only three are reset.

**Verified on the merged result** (locally, before pushing):

- Full suite: 1750 passed, 857 skipped, 2 xfailed — up from 1691/1699 on the two branches separately, so both PRs' tests run together.
- Both mechanisms coexist on one live object: memo populated, ceiling cached as `{('auto', (32, 32), complex64, 'cpu'): 15625}`.
- `tp == tp2` still holds with *both* caches populated — the assertion that fails if either exclusion is lost.
- The `"auto"` path is still 0.82 ms/call, so the `_get_site_ceiling` memoization survived the merge intact.

My approval still stands on the merged head. CI is running.

---

PR #431, opened 2026-09-18
Document that projection='finite' is not bit-reproducible on GPU

pzeiger:
## What

Building the same `Potential` twice with `projection="finite"` on GPU gives different arrays — four consecutive builds of an identical, fully specified potential produce four distinct byte patterns, ~1e-7 relative deviation, float32 rounding. Root cause: the finite projection's radial interpolation kernel (`abtem/core/_cuda.py`) accumulates overlapping atoms' contributions into shared grid cells with an atomic add, which guarantees no lost update but not an order, and float32 addition is not associative. `projection="infinite"` (the default) never superposes per-atom disks — it multiplies scattering factors in reciprocal space — so it has no atomic accumulation and stays deterministic on both devices; CPU is deterministic either way, since its own accumulation order is fixed. Nothing told a user which combination to expect reproducibility from.

Documents the actual reproducibility contract on `Potential`'s `projection` parameter: which combination of device and projection setting is deterministic, the size of the deviation, and the tolerance a test comparing GPU finite-projection output should use instead of exact equality.

The non-determinism itself. Every real fix trades something real against abTEM's non-default finite-projection path: deterministic accumulation (sort or partition contributions so summation order is fixed) costs kernel speed and complexity; accumulating in float64 and casting at the end reduces but does not eliminate order sensitivity, and how much it helps for realistic overlap counts needs measuring rather than assuming; a `config` flag selecting a deterministic kernel needs a decision on what the default should be. Picking one is a design decision, not a bug fix, and is left open. This documents the contract that exists today so a user chooses correctly in the meantime — build on CPU, or use the default `infinite` projection, where bit-for-bit reproducibility matters.

## Tests

None. This is a documentation-only change to an existing docstring; the underlying physical claims (which device/projection combinations are deterministic, and the ~1e-7 deviation size) were already established with a working reproduction in the issue file this PR partially addresses, not invented here. The issue file itself stays open, narrowed to the still-unfixed non-determinism, per this project's own convention for a partial fix.


TomaSusi (approved):
**Approving.** Verified the claims rather than taking them on trust:

- the atomic accumulation is real — `atomicAdd(&array[k * cols + m], val);` at `abtem/core/_cuda.py:191`, in the radial interpolation kernel;
- `"If 'finite' the 3D potential is numerically integrated..."` appears exactly once in the tree (`abtem/potentials/iam.py:1234`), so there is no duplicate `projection` block to drift out of sync with this one;
- there is no deprecation path on `projection`, so documenting the contract that exists today is the right thing to be doing.

Scoping the fix out, and saying explicitly which trade-off each candidate fix carries rather than picking one silently, is the right call for a non-default path.

---

PR #434, opened 2026-09-19
Fix FlexibleAnnularDetector's cross-call latch and the accelerator-en…

pzeiger:
## Summary

`FlexibleAnnularDetector._match_waves` latches onto the first `Waves` it ever sees (`if self.outer is None: self._outer = min(waves.cutoff_angles)`), so which member of a multi-energy `Probe` ensemble arrives first silently decides the angular axis for every member. Eager and lazy split the ensemble at different points, so they latch onto different members and disagree in shape.

Separately, `Accelerator.match`/`check_match` read `other.accelerator.energy` directly, which is `None` for an *indexed* member of an energy ensemble by design — its real energy lives in `metadata["energy"]` instead, the same place `Waves._valid_energy` already knows to look. Reading only the raw accelerator energy means `match` treats a perfectly well-defined 200 keV member as "no energy defined" and overwrites it with whatever the other side happens to hold (e.g. a `TransitionPotential`'s baked-in 100 keV), silently. `check_match`'s existing energy-mismatch guard (added in `ac4149b6` for exactly this class of problem) never fires, because by the time it runs the mismatch has already been erased.

## The fix

- **`abtem/core/energy.py`**: `_resolve_other_energy()` resolves through `metadata`/`ensemble_axes_metadata` before falling back to the raw accelerator energy, the same resolution order `Waves._valid_energy` already uses. Used in both `Accelerator.match` and `Accelerator.check_match`. This is a mechanism-level fix: every `.accelerator.match(`/`.accelerator.check_match(` call site benefits uniformly (`multislice.py`, `transfer.py`, `waves.py`, `core_loss.py`'s several call sites including the PRISM-EELS driver, `potentials/iam.py`, `s_matrix.py`) — none needed a separate change.
  - An **unbuilt, rebuildable `TransitionPotential`** now gets its per-task accelerator set to each ensemble member's real energy before `.build()` runs, so it rebuilds correctly and matches standalone single-energy runs, both energy orders, eager and lazy.
  - An **already-built `TransitionPotentialArray`** can't be made correct this way — its form factors are baked into the array at one energy. `check_match` now correctly recognizes the mismatch and raises, instead of silently relabeling the output:
    ```
    RuntimeError: Inconsistent energies: 100000.0 eV vs 200000.0 eV. A fixed energy
    (e.g. a built TransitionPotentialArray, or a CTF with an explicit energy) cannot
    be applied to wave functions at a different energy...
    ```
- **`abtem/detectors.py`**: un-latches `_match_waves` — `_outer_is_explicit` tracks "the user gave an outer angle" separately from "a value is currently matched", so a detector reused across two different single-energy scans re-matches instead of keeping the first scan's value. For a genuine multi-member energy ensemble with no explicit `outer`, it now raises rather than picking any one convention silently, since a single radial axis can't represent two different cutoff angles:
  ```
  RuntimeError: FlexibleAnnularDetector cannot auto-size its outer angle for a
  multi-energy ensemble: each energy has its own cutoff angle (aperture), so no
  single radial axis fits every member. Pass an explicit outer= ..., or run one
  energy at a time and combine the results yourself.
  ```
- **`abtem/multislice.py`**: the eager energy-ensemble branch now matches every detector against the full, un-split ensemble waves before splitting per member — mirroring what the lazy path's shape-resolution step already does — so eager hits the same clear raise as lazy, in either energy order, instead of silently latching on whichever member happens to be split off first.

An explicit `outer=` still bypasses the detector-latch guard as before, and then correctly hits the energy-mismatch guard instead if the transition potential is a built, fixed-energy array — the two guards are independent and neither can be used to silence the other.

## Tests

`test/test_core_loss_fixes.py`: three new test classes covering both energy orders and both eager/lazy paths where applicable, plus regression guards for single-energy scans, a length-1 energy list, and a detector reused across separate single-energy calls. Confirmed the fix-specific assertions fail on unfixed `dev` and pass on this branch; the regression guards pass on both, as they should. Single-energy and length-1-energy-list configurations are bit-for-bit identical before and after, both eager and lazy — nothing that was already correct changed.

## What this doesn't fix

PRISM-EELS (checked directly, not assumed): an energy-ensemble `SMatrix` already raises `EnergyUndefinedError` before reaching any of the code this PR touches, identically before and after — that combination is unsupported today, independent of this change.
A third, independent defect surfaced while building the standalone-vs-ensemble correctness tests above: an energy-ensemble scan against a scan with more than one position gives per-position values that don't match the corresponding standalone runs — confirmed present on this branch too, unaffected by either fix here, tracked separately as #435.

## Cross-check against open PRs

#384 and #365 are both still open and don't touch this mechanism: #384's detector changes are confined to `PixelatedDetector`/`WavesDetector`, nowhere near `_AbstractRadialDetector`/`_match_waves`/`FlexibleAnnularDetector`. #365 adds a new `IonizationDetector` that overrides `_calculate_new_array` to sum over all angles unconditionally and never calls `_match_waves`, so it's unaffected by the detector-latch fix; its new energy-ensemble `SMatrix` test class exercises the PRISM-EELS `EnergyUndefinedError` path noted above, not the fixed mechanism.


TomaSusi (approved):
**Approving**, with one wording fix worth making before merge.

This changes `Accelerator.match`/`check_match` for every call site in the tree, so I ran the whole suite rather than the touched files: **1706 passed, 854 skipped, 2 xfailed**. Things I checked specifically, since they are where this could have gone wrong quietly:

- **`resolve_energy` can't silently pick a member.** Its `len(axis.values) == 1` guard means a full, un-indexed multi-energy ensemble resolves to `None` and falls back to the previous behaviour. Had it returned `values[0]` instead, `match` would have relabelled a whole ensemble with its first member's energy — the exact failure this PR exists to prevent, reintroduced one layer down. It doesn't.
- **`_outer_is_explicit` is initialized on every construction path.** `AnnularDetector`, `FlexibleAnnularDetector` and `SegmentedDetector` all route through `_AbstractRadialDetector.__init__` with `outer=`, both `outer` setters maintain it, and `_match_waves`'s own auto-assignment at the end correctly writes `_outer` *without* setting the flag — which is what makes re-matching work. No path leaves it unset.
- **The new eager `_match_waves` loop is correctly scoped.** `hasattr(detector, "_match_waves")` matches only `_AbstractRadialDetector` subclasses; no other detector defines it, so nothing else is touched.
- **Checked and dismissed:** I expected `_resolve_other_energy`'s new `getattr(other, "metadata", ...)` to be able to trip a grid check for an object with an undefined grid, turning a previously-fine `match` into a `GridUndefinedError`. Tested `Probe`/`CTF` with neither energy nor grid defined, on `dev` and on this branch — identical behaviour. Not an issue.

### The nit: "(aperture)" is the wrong word, in a message users will act on

Both the new `RuntimeError` and `_match_waves`'s docstring say:

> each energy has its own cutoff angle (aperture)

`cutoff_angles` is the *antialias* cutoff — `antialias_cutoff_gpts // 2 * angular_sampling` (`abtem/waves.py:277`) — which is a property of the grid scaled by wavelength, and is independent of `semiangle_cutoff`:

| | `semiangle_cutoff=20` | `semiangle_cutoff=40` |
|---|---|---|
| 60 keV | 66.9 mrad | 66.9 mrad |
| 200 keV | 34.5 mrad | 34.5 mrad |

Identical across apertures, different across energies. The *behaviour* is right — the cutoff genuinely does differ per member, so refusing to auto-size a single radial axis is correct — but a user who reads "(aperture)" will go and change `semiangle_cutoff`, which will not move the number they were told is the problem. Something like "each energy has its own antialias cutoff angle (it scales with wavelength at fixed grid)" points them at the real cause, and at the actual remedies the message already gives.

Nothing else blocking. The `check_match` message is otherwise a big improvement over the bare `"Inconsistent energies"` — naming both values and explaining the built-`TransitionPotentialArray` case is exactly what that error needed.

---

pzeiger:
Fixed in `ca7256bf`.

Confirmed the claim directly rather than taking it on trust: `cutoff_angles` at 60 keV is identical for `semiangle_cutoff=20` and `semiangle_cutoff=40` (127.7 mrad both), and at 200 keV identical again (65.8 mrad both) — different across energy, unchanged across aperture, exactly as described. `semiangle_cutoff` never enters `cutoff_angles`' computation (`antialias_cutoff_gpts` scaled by `angular_sampling`), so "(aperture)" was pointing a reader at a parameter that does nothing here.

Reworded both the `RuntimeError` and `_match_waves`'s docstring to name the antialias cutoff and what it actually depends on (wavelength at a fixed grid) instead of the aperture. `test/test_core_loss_fixes.py`'s existing check only matches on `"cannot auto-size its outer angle"`, unaffected by the wording change; 129 passed, 2 xfailed.

TomaSusi (commented):
Verified `ca7256bf` — the wording nit is resolved, and more thoroughly than I asked for. The error message now says "each energy has its own antialias cutoff angle (it scales with wavelength at fixed grid)", and `_match_waves`'s docstring names `waves.cutoff_angles` directly and states outright that `semiangle_cutoff`/the aperture does not enter it at all. That second half is the part that will actually stop someone going down the wrong path.

My approval from earlier stands; nothing outstanding from my side.

---

PR #436, opened 2026-09-19
Resolve xp from the real array, not a possibly-stale device label

pzeiger:
## What

A detector's `to_cpu=True` (the default) moves a measurement's array to `numpy` without updating the object's own `.device` property, so `.device` can say `"gpu"` while `.array` is already a plain `ndarray`. `ArrayObject._stack` and `.to_hyperspy` both picked their array module from that label instead of the array's real type:

```python
xp = get_array_module(self.device)  # or array_objects[0].device
...
array = xp.stack(...)  # or xp.moveaxis(...)
```

`get_array_module("gpu")` resolves to `cupy` regardless of what `.array` actually is, so handing it a `numpy.ndarray` raises `TypeError: Argument 'a' has incorrect type (expected cupy._core.core._ndarray_base, got numpy.ndarray)`. A single measurement never notices — nothing forces `xp` to agree with `.array` until something needs the module to match. Combining two or more (`_stack`, reached via `abtem.stack`/`stack_array_object`) or exporting one (`.to_hyperspy`) both do, and both crash on GPU, eager (`lazy=False`) execution. The lazy path is unaffected: `_stack` branches to `da.stack` whenever any input is lazy, and `da.stack` doesn't need to know the concrete backing type ahead of time.

`ArrayObject.device` is not simply derived from `.array`: it returns the explicitly-stored `._device` when one is set, and only falls back to deriving it from `.array` otherwise. That stored override is itself deliberate — e.g. `squeeze` preserves it across a copy, and the ensemble-reduction path sets `output._device = "gpu"` on a now-CPU-resident result specifically so `_compute()` still picks the synchronous scheduler for what was GPU work, per its own comment. `.device` diverging from `.array`'s real type is therefore an intended part of its contract, not an oversight to patch there — `_stack` and `to_hyperspy` were wrong to assume the two always agree, not the label.

Both now resolve `xp` from the array itself — `get_array_module(self.array)` / `get_array_module(array_objects[0].array)` — matching the pattern every other `get_array_module` call in this file already uses (`concatenate`, `swapaxes`, `moveaxis`, `squeeze`, and `.device`'s own fallback branch); `_stack` and `to_hyperspy` were the only two outliers. `get_array_module` already handles a dask array via its `_meta`, so this is safe for `_stack`'s lazy branch too, even though that branch never reads `xp`.

## Blast radius

Every other `get_array_module(X.device)` call site in the tree (31, across `waves.py`, `potentials/iam.py`, `prism/s_matrix.py`, `bloch/dynamical.py`, `multislice.py`, `detectors.py`, `tilt.py`, `mcf.py`, `transform.py`, `antialias.py`, `inelastic/`) resolves it on a `Waves`/`Potential`/`SMatrix`/dynamical-diffraction object. None of those pass through a detector's `to_cpu` transfer, so their `.device` stays accurate on this path — a different mechanism, not an instance of this one.

Two call sites delegate to the two methods fixed here rather than duplicating them, so they inherit the fix automatically: `PotentialArray.to_hyperspy` (via `Images.to_hyperspy`) and `Waves._stack_over_transitions` (via `stack_array_object`, which is `abtem.array.stack` under an alias, itself dispatching to `arrays[0]._stack`). `IndexedDiffractionPatterns._stack` is a genuine override of the base classmethod, but it never calls `get_array_module` — it stacks plain, always-host `numpy` dicts directly — so it was never exposed to this mechanism.

## Tests

`TestStackAndHyperspyTrustTheRealArrayType` in `test/test_array.py`: a precondition test pins down the setup (`.array` is `numpy.ndarray`, `.device` is `"gpu"`, produced by a plain elastic `AnnularDetector` scan — no core-loss machinery needed, this is a general property of any eagerly-detected GPU measurement); one test confirms `stack()` no longer crashes and the result matches a direct `numpy.stack` of the same arrays; one confirms `to_hyperspy()` no longer crashes and the result matches the array transposed the same way the method's own `transpose=True` default does. The `to_hyperspy` test stubs `abtem.array.hs`'s two signal classes rather than gating on hyperspy being installed (`test_hyperspy.py`'s existing tests do that, and are skipped in this environment) — hyperspy not being installed everywhere this suite runs is exactly the condition under which this regression would go unnoticed by a test that only runs when it is. All three fail on unfixed `dev` (the two crash tests with the reported `TypeError`; the precondition test passes on both, since it isn't testing the fix).


TomaSusi (approved):
**Approving.** The fix is right and minimal: resolving `xp` from `.array` rather than the `.device` label is unambiguously correct, and the reasoning for fixing the two consumers rather than the label itself is sound — `.device` diverging from `.array` is a deliberate part of its contract (the ensemble-reduction path sets `_device = "gpu"` on a host-resident result precisely so `_compute()` picks the synchronous scheduler), so the consumers were the ones making the bad assumption.

Verified the two load-bearing claims rather than taking them on trust:

- `get_array_module` does handle a dask array via `_meta` (`abtem/core/backend.py:343`), so resolving from the real array is safe in `_stack`'s lazy branch too.
- These two really were the only `get_array_module(<...>.device)` outliers in `abtem/array.py` — every other `get_array_module` call in the file already reads an array.

Full suite on this branch: 1691 passed, 857 skipped.

### One note: the new tests never run in CI

All three are `@requires_gpu`, and `.github/workflows/tests.yml` runs only `ubuntu-latest`/`windows-latest`/`macos-latest` — no GPU runners — so they skip on every PR and only ever execute on a workstation with cupy.

That is slightly at odds with the reasoning given for stubbing hyperspy:

> A version of this test gated behind hyperspy's presence (test_hyperspy.py's own skipif) would silently skip in exactly the environments where this regression would go unnoticed.

The same argument applies one level up to `@requires_gpu`. I don't think it's avoidable for these tests as written — reproducing the bug needs `get_array_module("gpu")` to actually resolve to cupy — so this isn't a change request, just worth being explicit that CI cannot catch a regression here and the guard depends on someone running the suite on a GPU box. If you wanted a CPU-runnable complement, asserting that `_stack`/`to_hyperspy` pass `.array` (not `.device`) to `get_array_module` would be a weaker but universally-executing guard.

---

pzeiger:
Added the CPU-runnable complement, in `61094cfc`.

`test_get_array_module_receives_the_array_not_the_device_label` skips the GPU/cupy machinery entirely: a plain CPU `Images` object with `._device` set to an arbitrary marker string that disagrees with `.array`, and a monkeypatched `get_array_module` recording what it's called with. Asserts `_stack` and `to_hyperspy` both call it with the real array, never the device label. It doesn't reproduce the crash — that still needs a GPU, as you said — but it does directly check the fix's actual invariant, and runs on every CI job rather than only on a workstation with cupy. Verified it fails on unfixed `dev` (records the device string instead of the array) and passes with the fix. `test/test_array.py`: 419 passed.

TomaSusi (commented):
Verified `61094cfc` — this is a good answer to the CI-coverage note, better than the weaker version I suggested.

`test_get_array_module_receives_the_array_not_the_device_label` asserts the fix's real invariant directly: `get_array_module` is called with `m.array` (an identity check, `c is m.array`, not just an equal value) and never with a string. Using a marker `_device` rather than a genuine GPU mismatch is the right call — the invariant under test is which argument gets passed, which doesn't care what the two actually contain, only that they disagree. It runs here on CPU and passes, and it would fail on unfixed `dev`.

The three `@requires_gpu` tests still can't run in CI, but that is now a coverage overlap rather than a gap.

My approval from earlier stands; nothing outstanding from my side.

---

PR #437, opened 2026-09-19
Register dask.sizeof for abTEM's object-array-wrapped payloads

pzeiger:
## Summary

`dask.sizeof.sizeof` has no registration for any of abTEM's classes, and abTEM ships large payloads to dask wrapped in a small object-dtype `ndarray` (`_wrap_with_array` in `abtem/core/ensemble.py`, 32 call sites across 14 modules — potentials, transition potentials, waves, scans, detectors, frozen phonons, S-matrices, ...). Neither half is accounted for: a wrapped payload's `sizeof` is a handful of bytes for the wrapper's own pointer, regardless of what it carries, and `sys.getsizeof` on the unwrapped object itself is a similarly fixed few dozen bytes. Distributed's spill, rebalance and transfer-cost decisions all read from `sizeof`, so a worker holding hundreds of MB of abTEM payload looks empty to them — it never participates in spill-to-disk, the dashboard reports it as "unmanaged" memory, and moving it between workers is costed as free.

## The fix

`abtem/core/dask_sizeof.py`, registered at abTEM's own import time (last line of `abtem/__init__.py`, after every class it registers already exists):

1. **Explicit registrations** for the classes that hold a payload directly: `ArrayObject` (covers every subclass — `Waves`, `PotentialArray`, `TransitionPotentialArray`, `Images`, `SMatrixArray`, ...), `CompressedSMatrixArray` (not an `ArrayObject`; sums its four truncated-SVD factors directly), and `ase.Atoms` (carried the same way by `DummyFrozenPhonons`, and just as unregistered upstream).
2. **The part that actually matters**: an override of dask's own `np.ndarray` handler so an object-dtype array reports the summed size of its elements, recursing back through `sizeof` — which is what makes a *wrapped* payload (the shape abTEM actually hands to dask everywhere) report correctly, not just a bare unwrapped instance. Every non-object dtype delegates unchanged to dask's own captured implementation, including its 0-in-strides broadcast-view special case, so ordinary array accounting is untouched. Registering (1) alone would fix nothing real, since abTEM never hands dask a bare `ArrayObject` — only the wrapped array.

This is scope-generic by construction: since the override is keyed on dtype (`object`) rather than on abTEM's own types, every one of the 32 `_wrap_with_array` call sites benefits automatically, including the many carrier types not explicitly registered in (1) (`Potential`, `LineScan`/`GridScan`/`CustomScan`, detectors, transforms, distributions). Those now get their own correct generic size once unwrapped, rather than the wrapper's constant few bytes — a real improvement even without a dedicated registration, though a class with its own large internal array would need one to be sized exactly rather than through the generic `sys.getsizeof` fallback.

Both registrations happen at import time because `dask.sizeof.sizeof` is a `Dispatch` whose lookup table is a plain dict keyed by exact type: a later `.register()` call for a type overwrites its own entry immediately, but a *subclass* weighed before its parent class is registered gets its own stale cache entry that a later parent registration cannot retroactively fix. This can't happen for abTEM's own classes (`TransitionPotentialArray` etc. cannot exist before `import abtem` has already run this registration), which is why only the numpy override needs the explicit "capture the real handler before installing ours" step — `np.ndarray` is a foreign type that other code could plausibly have already weighed.

## Verified numbers (this branch, issue's own reproduction)

| | before | after |
|---|---|---|
| `sizeof(tp)` (16 MB `TransitionPotentialArray`) | 48 | 16777264 |
| `sizeof(_wrap_with_array(tp, ndims=0))` | 8 | 16777272 |
| `sizeof(_wrap_with_array(tp, ndims=1))` | 8 | 16777272 |
| a 4-element object array of `float64[1000]` arrays | 32 | 32032 |

`managed_bytes` reported by a real `distributed.LocalCluster` worker for the same wrapped payload (`scheduler_info()["workers"][addr]["metrics"]["managed_bytes"]`, never polled via `client.run`) goes from 8 to 16777272 for the identical computation — confirming the fix reaches the actual accounting distributed's spill logic reads, not just a standalone `sizeof()` call.

## Real-workload validation

Timed a production-shaped multi-GPU core-loss EELS scan (Perlmutter, 4×A100, `dask.multi-gpu`) on this branch against `dev`, same configuration, several repeats each. No measurable wall-clock regression: this branch's core-loss timings (263.8-283.6 s, 4 runs) fully overlap `dev`'s (277.0-281.0 s, 2 runs) — the run-to-run spread within each branch (traced separately to host-side page-cache warming across consecutive container invocations, unrelated to this change) is larger than any difference between them.

## Tests

`test/test_dask_sizeof.py`, 8 tests, each running in a fresh subprocess rather than in-process — `dask.sizeof.sizeof`'s dispatch cache is a single process-wide table, so an earlier test in the same pytest session could plausibly have already weighed one of these exact types before this file's own tests run, silently invalidating an in-process assertion for reasons unrelated to the registration's own correctness. Covers: `TransitionPotentialArray` direct and wrapped (both `ndims`), a generic (non-abTEM) object array of numpy arrays, `ase.Atoms`, ordinary non-object arrays unchanged (including the broadcast-view case), ordinary arrays unchanged even when dask's own lazy numpy registration is forced to fire before `import abtem`, and that an object array of arbitrary user objects abTEM has no registration for is sized by its elements' own generic `sizeof` rather than inflated by reaching into them. Confirmed each test that exercises the fix fails on unfixed `dev` (6 of 8; the other 2 assert non-regression of already-correct behaviour and correctly pass either way).

## Scope

Deliberately limited to the accounting fix. Not touched: spill thresholds, dashboard presentation, or anything about which workflows should change their settings as a result — this changes a number distributed already reads, not the policy built on top of it.


TomaSusi (commented):
The design is well targeted — the object-dtype `ndarray` override is correctly identified as the part that does the real work, since abTEM never hands dask a bare `ArrayObject`, only the wrapper. Registering (1) without (2) would indeed have fixed nothing.

Verified the numbers independently rather than trusting the table:

```
plain ndarray (4 MB)      4000000     (identical before and after `import abtem`)
broadcast view (0-stride)       80     (dask's own special case preserved)
eager Waves (2 MB array)   2097200
lazy  Waves                    136     <- correctly ~nothing
wrapped eager              2097208
wrapped lazy                   144
```

The lazy case is the one I most expected to be wrong — a naive `_array`-based implementation would report the full logical `nbytes` for a dask-backed payload and make every lazy object look enormous — and it isn't, because dask has no `da.Array` registration and the fallback is small. Worth knowing that's load-bearing and silent.

Also confirmed the 10-sample scaling is a verbatim mirror of dask's own `sizeof_python_collection`, not something invented here. Full suite on this branch: 1699 passed, 854 skipped. `test/test_dask_sizeof.py`: 8 passed.

Two nits — **holding off on approving until you've had a look at these**, neither is a correctness problem in normal use.

### 1. `importlib.reload` of this module recurses infinitely

```python
import abtem, importlib, numpy as np
import abtem.core.dask_sizeof as m
importlib.reload(m)
np.zeros(10)  # -> RecursionError
```

On reload, `sizeof.dispatch(np.ndarray)` returns the override this module installed the *first* time, so `_dask_sizeof_numpy_ndarray` becomes `_sizeof_ndarray` itself and every non-object array delegates to itself forever.

Manual `importlib.reload` is exotic, but `%autoreload 2` is not, and abTEM users spend a lot of time in notebooks — an autoreload triggered by editing anything under `abtem/core/` would hit this. A sentinel closes it in one line:

```python
_captured = sizeof.dispatch(np.ndarray)
if getattr(_captured, "_abtem_sizeof_override", False):
    _captured = _captured.__wrapped_dask_impl__   # or just keep the original module-level ref
_dask_sizeof_numpy_ndarray = _captured
```

Flagging it mainly because the module is otherwise meticulous about precisely this class of import-order/dispatch-caching hazard — its docstring spends a paragraph on the memoization trap — so this feels like an omission rather than a considered trade-off.

### 2. The sampled estimate is unstable for a heterogeneous object array

A 100-element object array carrying one 8 MB payload and 99 tiny ones, weighed 40 times in a row on the same object:

```
distinct values: [1600, 80001520]
true payload   : 8000000
```

Either 5000x under or 10x over, depending on whether the 10-element sample happens to include the big one — and nondeterministic across calls on an unchanged object.

This is inherited from dask, and it doesn't bite abTEM's actual usage: `_wrap_with_array` gives `n == 1` (exact, no sampling), and a real ensemble's elements are homogeneous in size, where a scaled sample is fine. So this may well be acceptable as-is. But the whole point of the PR is that distributed's spill decisions can trust this number, and above 10 elements it becomes an unstable estimate rather than a measurement. Since object arrays here are small by construction, raising the exact-sum threshold (say to a few hundred elements) would be close to free and would make the common cases deterministic. Your call whether that's worth diverging from dask's constant.

### Minor

The PR body says "32 call sites across 14 modules"; the module docstring says "roughly fifty call sites". Worth making these agree.

---

pzeiger:
Thanks for verifying the numbers independently, and for catching the reload case.

**Infinite recursion on module reload**: fixed in `0517703c` — `_sizeof_ndarray` is now marked with an attribute, checked against whatever `sizeof.dispatch(np.ndarray)` returns before re-capturing it: if that's this module's own wrapper (the marker is set) and a previously-captured delegate is already bound in the module's namespace, the previous delegate is reused instead of re-capturing the wrapper. Verified fixed across multiple consecutive reloads, with object-array sizing still correct afterward. Added `test_module_reload_does_not_recurse` to `test/test_dask_sizeof.py`.

**Sampled estimate instability**: left unchanged, deliberately. Verified `_wrap_with_array` always produces a single-element object array (`n == 1`, exact, no sampling involved), and every other object-dtype array in the tree holds either dask graph blocks or lightweight axis metadata — homogeneous in size, not the large, heterogeneous payloads that make the sampled estimate unstable. Raising the exact-sum threshold would trade the uniformly-fast `sizeof()` calls that both a scheduler under load and abTEM's actual usage benefit from, for robustness against a case the tree does not currently construct.

**32 vs "roughly fifty" call sites**: fixed in the same commit — the module docstring now says 32, matching the PR body's count and a fresh grep of every `_wrap_with_array`/`shared_constant_arg` call site.

TomaSusi (approved):
**Approving** — `0517703c` closes both nits.

**Reload recursion.** Re-ran my original reproduction across five consecutive reloads rather than one:

```
baseline      : plain=4000000  objarray=24024
after reload 1: plain=4000000  objarray=24024
...
after reload 5: plain=4000000  objarray=24024
broadcast view still special-cased: 80
```

No recursion, delegation to the upstream handler still correct, object-array summing intact, and dask's 0-in-strides case still reached. The `_previously_captured` approach also holds up on the subtle bit: after a reload, dask's registry holds the *previous* `_sizeof_ndarray` object, and the marker is on that object, so the check fires on the right thing rather than on the freshly-defined function. `test/test_dask_sizeof.py`: 9 passed. Full suite: 1700 passed, 854 skipped.

**Sampling.** Agreed, and the case is stronger than the reasoning gives it credit for. It isn't just that `_wrap_with_array` happens to produce `n == 1` today — it does so structurally:

```python
wrapped = np.zeros((1,) * ndims, dtype=object)
itemset(wrapped, 0, x)
```

That is exactly one element for any `ndims` (checked 0 through 3, all `size == 1`). So the `n > num_samples` branch is not merely unexercised by the current tree, it is unreachable through the transport convention this module exists to serve. Keeping dask's constant rather than diverging is the right call, and worth noting in case anyone revisits: the way this could become live is a *new* wrapper that packs multiple payloads into one object array, not new call sites of the existing one.

Docstring count now matches the PR body. Nothing outstanding.

---

PR #438, opened 2026-09-19
Parallelize sum_run_length_encoded over batch and pixels, not just bins

pzeiger:
## Summary

`sum_run_length_encoded` (`abtem/core/_cuda.py`) is the CUDA kernel behind `DiffractionPatterns._radial_binning`, which every radial/polar detector (`FlexibleAnnularDetector`, `SegmentedDetector`, `AnnularDetector` outside its single-bin fast path) runs once per site batch. It launched one thread per bin and had that single thread serially loop over both the batch axis and every pixel in the bin's segment. With `n_bins` fixed by detector angular resolution (typically ~10-1000) and unrelated to problem size, this occupied a single CUDA thread block regardless of how large the batch or the diffraction pattern actually were — on an A100, one Streaming Multiprocessor out of 108, the other 107 idle for the kernel's entire execution.

This was found while investigating low GPU utilization on a production core-loss EELS scan: `nvidia-smi` showed ~99% GPU-Util alongside only 100-180 W of a 400 W power cap, and a full time breakdown placed 87.3% of the scan's wall-clock inside this one kernel call.

The kernel now launches one block per `(bin, batch)` pair — `n_bins * n_batch` blocks instead of `ceil(n_bins / 256)` — and each block sums its bin's pixel segment with a strided read plus a block-level shared-memory tree reduction across threads, instead of one thread walking the segment serially. Both axes that actually scale with problem size (batch count and pixels-per-bin) now get real parallelism; only `n_bins` — a small, roughly fixed parameter — does not need its own grid dimension.

## Performance

Measured directly, same process, same array, before and after, on two independent GPUs — an A100 (the hardware the original production scan ran on) and a consumer iGPU used for local development:

| shape (n_batch, n_selected, n_bins) | A100 before | A100 after | A100 speedup | iGPU before | iGPU after | iGPU speedup |
|---|---|---|---|---|---|---|
| (10, 5,000, 32) | 0.075 ms | 0.015 ms | 5.1x | 0.117 ms | 0.007 ms | 16.5x |
| (80, 400,000, 97) — mirrors a real `FlexibleAnnularDetector` call | 40.325 ms | 0.101 ms | 399.5x | 81.050 ms | 0.808 ms | 100.3x |
| (1,000, 400,000, 97) — 12.5x the batch, same per-call diffraction-pattern size | 504.041 ms | 1.139 ms | 442.6x | 1060.080 ms | 9.736 ms | 108.9x |

The A100 sees a larger speedup at the two batch-heavy shapes than the iGPU does, consistent with the fix's mechanism: it trades a single occupied thread block for `n_bins * n_batch` blocks, and a GPU with more Streaming Multiprocessors to spread those blocks across gets more benefit from the same change.

On the production EELS scan this was found on, `nvidia-smi` now reads ~90% utilization and ~300 W of the A100's 400 W cap, up from the ~99% utilization / 100-180 W split described above — consistent with the GPU doing genuine parallel work rather than being kept nominally "busy" by a single serialized thread block. Re-running that scan's own internal time breakdown, same 6733 calls into this kernel: 2.36 s total (2.2% of wall-clock, 0.35 ms/call average), down from the 715.70 s (87.3%, 106.30 ms/call average) that motivated this fix — a 303x reduction in this step, on the exact workload that surfaced the problem.

## Correctness

The reduction now sums the same terms in a different order (block-parallel rather than one serial thread), so exact bit-for-bit equality with the pre-fix kernel is not expected. Output is verified against the CPU oracle (`_sum_run_length_encoded`) across 8 shapes x 2 dtypes (float32/float64), with tolerance scaled to each shape's own accumulated magnitude rather than a bare default `atol` — including degenerate cases (zero batch rows, zero bins, a single giant bin, more bins than selected pixels). All match to within the data's own floating-point accumulation error. Independently, the benchmark run on the A100 above also asserts old-vs-new agreement for each of its three shapes before trusting its own timing, and all three passed on that hardware too.

## Other call sites of this mechanism

`sum_run_length_encoded` has exactly one caller in the codebase: `DiffractionPatterns._radial_binning` (`abtem/measurements.py`). No other site needed changing.

## Tests

`test/test_radial_bin_sum_kernel.py` adds:
- A correctness sweep against the CPU oracle across 8 shapes and both dtypes.
- Zero-sized-input handling (zero batch rows, zero bins).
- A timing-bound regression test for the actual defect (batch-axis serialization): confirmed to fail against the pre-fix kernel (303.7 ms measured) and pass against the fix, with the bound set an order of magnitude looser than the post-fix measurement and well inside the pre-fix one so it doesn't pin to this session's specific hardware.

The full existing suite also passes against this branch on an A100: 2246 passed, 319 skipped, 0 failed, 0 errors in 462.81s.

## Known residual

The within-bin pixel sum is parallelized via a fixed-size (256-thread) shared-memory reduction rather than also scaling the reduction's own width to the segment size; this is a large improvement over the fully serial pre-fix design but leaves some headroom for a bin with an unusually large pixel count relative to others in the same call. The two changes (one block per `(bin, batch)` pair, and the within-block reduction) were not benchmarked in isolation from each other, so the table above reflects their combined effect, not an attribution to either one individually.


TomaSusi (commented):
I don't have a GPU on this machine, so rather than take the kernel on trust I transcribed its exact indexing and reduction — `blockIdx.x % n_bins` / `blockIdx.x / n_bins`, the strided read `j = start + threadIdx.x; j += blockDim.x`, and the `offset >>= 1` shared-memory tree — into a simulator and checked it against a direct per-bin oracle:

```
n_batch=   1 n_selected=       1 n_bins=   1   max rel dev=0.00e+00
n_batch=  10 n_selected=       5 n_bins=  10   max rel dev=0.00e+00   (n_selected < n_bins, many empty bins)
n_batch=   2 n_selected=    5000 n_bins=   1   max rel dev=4.33e-16   (single giant bin)
n_batch=   8 n_selected=    4096 n_bins= 256   max rel dev=1.61e-16
n_batch=   3 n_selected=    1000 n_bins=  97   max rel dev=1.43e-16
worst across all shapes: 4.33e-16
```

The decomposition is correct, including the degenerate shapes. Three further things I checked:

- **The `i >= n_batch` early return cannot deadlock.** Every thread in a block shares `blockIdx.x` and therefore `i`, so the return is block-uniform and no thread is left waiting at the later `__syncthreads()`. (It is also dead code at the current launch config, since `grid.x == n_bins * n_batch` — fine as defence.)
- **Grid size can't silently overflow.** `n_bins * n_batch` is computed as a Python int and cupy rejects a grid above 2^31, so an extreme batch fails loudly rather than wrapping `blockIdx.x` into the `int i`.
- **The `n_batch == 0` guard is a real addition**, not just tidiness — without it the new launch config would request a zero-block grid where the old one always had at least one.

Full suite on this branch: 1691 passed, 855 skipped (the kernel file isn't imported without cupy).

One nit — **holding off on approving until you've had a look**, since it is a trap rather than a live bug.

### The power-of-two requirement on `blockDim.x` is documented but not enforced

The launcher comment is clear that `threads` must stay a power of two, because the reduction only sums pairs at each halving stride. But nothing checks it, and the failure is silent — simulating `blockDim = 100` on a 100-element single bin:

```
blockDim=256 (pow2) : -8.25405098643877    (true sum: -8.254050986438774)
blockDim=100 (!pow2): -3.0717216596466628  <- silently wrong, no error
```

`threads = 256` is a hardcoded literal three lines below the comment, so this only bites a future edit — but it bites it silently, in a kernel whose output feeds every radial/polar detector, and the correctness sweep in `test_radial_bin_sum_kernel.py` would not catch it either since it never varies `threads`. One line makes it impossible:

```python
assert threads & (threads - 1) == 0, "the kernel's tree reduction requires a power-of-two block size"
```

### Smaller, no action needed

`test_sum_run_length_encoded_parallelizes_over_batch` is a wall-clock bound, which is the kind of test that ages badly on shared or thermally-throttled hardware. At ~17x headroom over your reported post-fix measurement, and GPU-only so it never runs in CI, I think the risk is acceptable — noting it only so it's a known trade-off rather than a surprise later.

The measured results and the mechanism story behind them are convincing, and I appreciate that the "known residual" section says plainly which of the two changes were not benchmarked in isolation.

---

pzeiger:
Thanks for the careful check.

**Power-of-two block size**: fixed in `b102f462` — added `assert threads & (threads - 1) == 0` right after the `threads = 256` literal, with the message pointing at the tree-reduction requirement the comment above it already described. Reproduced the silent-corruption case before fixing it: a 100-element single-bin reduction under a non-power-of-two block size returns 3.51 against a true sum of 8.11, matching what the review's simulation found. No dedicated test added for the assertion itself — `threads` is a hardcoded local literal, not something any caller or existing test parametrizes, so the assertion is the whole mitigation; the review's own "one line makes it impossible" already treats it that way.

**Timing-bound test fragility**: agreed as stated — GPU-only, never runs in CI, and the ~17x headroom over the measured post-fix time is a deliberate trade-off, not an oversight. No change made.

TomaSusi (approved):
**Approving** — `b102f462` closes the nit.

The assertion sits immediately after the `threads = 256` literal and before `block`/`shared_mem` are derived from it, so nothing can be launched with a block size the tree reduction cannot handle. File parses cleanly; full suite 1691 passed, 855 skipped (the kernel file still isn't imported without cupy).

Good that you reproduced the silent-corruption case independently before fixing it rather than taking my simulated numbers on trust — different data, same shape of result (3.51 against a true 8.11, where my transcription gave -3.07 against -8.25), which is the useful kind of agreement: two independent reproductions of the same mechanism rather than one number copied forward.

Agreed on the timing bound, and on not adding a test for the assertion — `threads` is a hardcoded local with no caller or test that parametrizes it, so the assertion is the mitigation rather than something needing its own guard. Recording it as a stated trade-off is the right outcome.

Nothing outstanding.

---

PR #439, opened 2026-09-20
Stack a GridScan energy-ensemble's per-member results at the right axis

pzeiger:
## Summary

A `GridScan` probe's multi-energy ensemble (`Probe(energy=[...]).transition_potential_scan(...)`) put the energy axis at the wrong position of the returned array. This is not scrambled data — the array's raw values are internally consistent and match the corresponding standalone single-energy runs exactly, once read along the axis the data actually uses — but the measurement's own `ensemble_axes_metadata`/`axes_metadata` unconditionally reports the energy axis leading, while `GridScan`'s data had it trailing. Anything that trusts that metadata to interpret the array — an automated axis-aware slice, a hyperspy/xarray conversion, `array[energy_index]` following the stated convention — gets silently wrong results.

`MultisliceTransform._calculate_new_array` (`abtem/multislice.py`) splits an energy ensemble per member, then reassembles the results with `np.stack(per_energy, axis=energy_axis_idx)`, where `energy_axis_idx` locates the energy axis within the *input* `waves` object's own combined ensemble axes — a different, unrelated axis list from the *output* measurement's own metadata. For `GridScan`, `waves.ensemble_axes_metadata` is `[ScanAxis, ScanAxis, EnergyAxis]`, so `energy_axis_idx=2`; each per-energy member's reduced result is already a clean `(rows, cols)` array with that axis stripped, so stacking at position 2 appends the new axis past the end instead of where the metadata says it belongs.

## Fix

`detectors.py`'s own `_out_ensemble_axes_metadata` already computes the correct answer for the output type: it calls `_scan_axes(waves)` (`abtem/measurements.py`) to find which of `waves`' ensemble axes are the "main" 2D `ScanAxis` pair, then moves exactly those to the end, ahead of `_out_base_shape` folding them into the base/image shape. This PR mirrors that same computation in `_calculate_new_array`, instead of reusing `energy_axis_idx` or a hardcoded axis:

```python
scan_source = _scan_axes(waves)
new_energy_axis = sum(1 for i in range(energy_axis_idx) if i not in scan_source)
```

This gives `0` for `GridScan` (both `ScanAxis` entries excluded, energy is the only remaining axis) and `1` for a `CustomScan` probe (whose single `PositionsAxis` is not a `ScanAxis` subclass at all, so nothing is excluded and energy keeps its original second position) — matching each scan type's real, already-correct metadata exactly.

An earlier attempt hardcoded `axis=0`, which fixed `GridScan` but broke the existing `CustomScan`-based `TestUnbuiltTransitionPotentialEnergyEnsemble` test, whose own assertion (`ensemble[..., i]`, energy trailing) is correct for that scan type. Recorded in the code comment so a future edit doesn't reintroduce it.

Eager and lazy share this exact function (`lazy=False` calls it directly, `lazy=True` invokes it once per dask block), so one fix covers both — confirmed directly, no separate lazy-side change was needed.

## Other call sites of this mechanism

Three other places match `isinstance(ax, EnergyAxis) and len(ax.values) > 1` in `waves.py`:

- `Waves.apply_ctf` (line ~1487): splits per energy and restacks with `abtem.array.stack(members, energy_axis, axis=axis_idx)`. Unaffected — this operates entirely within the `Waves` domain, before detection, where scan axes are never reclassified into base shape (that reclassification is `detectors.py`'s own, detector-side transformation). `axis_idx`, computed from the full pre-split ensemble list, is exactly where the axis belongs in the restacked result, since nothing reorders axes in between.
- Two lazy-rechunking sites (`waves.py` ~2310, ~2678): ensure each energy value gets its own dask chunk before multislice. No stacking involved, unrelated to this mechanism.

`sum_run_length_encoded`-adjacent `energy_axis_idx` uses in `measurements.py` (EELS energy-loss-axis processing, plasmon/phonon-loss) are a different axis type (`EnergyLossAxis`, not `EnergyAxis`) and a different concept (energy-loss spectrum, not beam-energy ensemble) — not instances of this mechanism.

## Tests

`test/test_core_loss_fixes.py`:
- New `TestGridScanEnergyEnsembleAxisOrder`, mirroring the existing `CustomScan`-based `TestUnbuiltTransitionPotentialEnergyEnsemble` now that the `GridScan` case it deliberately avoided is fixed. Confirmed to fail against the pre-fix code (both energy orders, both eager and lazy) and pass against the fix.
- Verified directly against a two-detector scan (not covered by the parametrized test, since the fix's tuple-output branch is otherwise unexercised) — both detectors' outputs correctly reproduce their standalone runs.

Full existing suite run locally: `test_core_loss_fixes.py` (162 passed, 2 xfailed — same xfail count as baseline, both pre-existing and unrelated), `test_waves.py` (134 passed), `test_detect.py` + `test_measure.py` + `test_prism_upsample.py` (359 passed), `test_energy_ensemble.py` + `test_multislice_input.py` + `test_realspace_multislice.py` (118 passed, 7 skipped), `test_scan.py` + `test_ionization.py` + `test_transition_potential_transport.py` (86 passed).

## Known residual

A separate, unrelated defect remains: a lazy (`lazy=True`), multi-member energy ensemble crashes for most scan/energy-count shapes, each with a different error — a `CustomScan` probe raises `ValueError: too many values to unpack (expected 2)`, a `LineScan` probe raises `IndexError: tuple index out of range`, and a `GridScan` probe with 3 or more energies raises `RuntimeError: number of values for ordinal axis (...) does not match size of dimension (...)`. The only lazy shape confirmed to work is `GridScan` with exactly 2 energies, which is what this PR's own test covers. Confirmed by direct testing (including a temporary revert of this PR's own fix) that all three crashes are untouched by this fix and do not touch this fix's own mechanism. The `CustomScan` case is already marked `xfail(strict=True)` in the existing test suite; all three are tracked separately.


TomaSusi (commented):
The fix is correct, and the design choice behind it is the right one: mirroring `_out_ensemble_axes_metadata`'s own computation rather than hardcoding an axis is what keeps this in sync with `detectors.py` by construction instead of by coincidence, exactly as the code comment says. Verified the claims rather than taking them on trust.

Measured against standalone single-energy runs, in every case the PR's mechanism touches:

| case | pre-fix deviation | post-fix |
|---|---|---|
| GridScan eager | 2.367e-06 | **0.000e+00** |
| GridScan lazy | — | **0.000e+00** |
| Two-detector tuple branch (both detectors) | — | **0.000e+00** |

Bit-identical, not merely inside tolerance. The two-detector result independently reproduces the manual check described in the PR, so the tuple branch is confirmed from a second direction.

Three further things I checked:

- **The new test genuinely discriminates.** Reverting `new_energy_axis` back to `energy_axis_idx` gives 4 failed, so `TestGridScanEnergyEnsembleAxisOrder` is not passing vacuously — and its own `assert not np.allclose(reference[...], reference[...])` vacuity guard is the right instinct for this shape of test.
- **The residual is exactly as described.** Ran the three crashing shapes on `dev` and on this branch: byte-identical outcomes (`line` lazy → `IndexError`, `grid` lazy n=3 → `RuntimeError` on the ordinal axis, `grid`/`custom` lazy n=2 → OK). Pre-existing and untouched, as claimed.
- **Full suite on this branch:** 1754 passed, 857 skipped, 2 xfailed.

## LineScan eager is a silent, untested beneficiary

The PR reasons about GridScan and CustomScan. LineScan is the third case, and it is the interesting one: the three scan types exercise **three different values** of the new expression, because `_scan_axes` returns `scan_axes[-2:]`, which for a single `ScanAxis` is a 1-tuple:

- GridScan — 2 axes excluded → energy rank 0
- CustomScan — 0 axes excluded (`PositionsAxis` is not a `ScanAxis`) → energy rank 1
- **LineScan — exactly 1 axis excluded** → the only case exercising the partial-exclusion arithmetic

It was broken before this PR and is fixed by it, with the same signature as GridScan:

```
pre-fix   line: ens[0] vs ref[100k] max abs dev = 2.367e-06
post-fix  line: ens[0] vs ref[100k] max abs dev = 0.000e+00
```

So the fix is *broader* than the PR claims — good news — but the newly-repaired case has no regression guard, and it is the one case whose arithmetic neither tested scan type covers. Adding `"line"` alongside the existing GridScan class looks like one parametrize value plus a scan constructor; worth considering before merge, since it is the case most likely to break under a future edit to `_scan_axes` or to this expression.

## Two smaller notes, no action needed

**Lazy coverage of the fixed path is inherently thin.** Per the residual, lazy GridScan works only at exactly 2 energies — which is precisely the shape the new test's `lazy=True` parametrization uses. That is not a criticism of the test (it cannot do better until the residual is fixed), but it does mean the lazy half of the matrix is pinned to the single shape that happens to work today, and will silently stay that way if the residual is fixed later without revisiting this test.

**A pointer for the tracked residual.** The lazy LineScan `IndexError` bottoms out at `abtem/array.py:201`, in `multi_output_blockwise`'s `drop_axis` check. That is consistent with `_out_ensemble_shape` doing `ensemble_shape[:-2]` unconditionally whenever `_scan_shape(waves)` is non-empty, which over-strips when only one `ScanAxis` exists — the same single-vs-pair asymmetry in `_scan_axes` that makes LineScan interesting above. I did not prove the causal chain, so treat it as a starting point rather than a diagnosis.

Holding off on approving only for the LineScan parametrize question — the fix itself I am satisfied is correct.

---

pzeiger:
Added `LineScan` coverage in `aed9d955`.

Confirmed the analysis: `LineScan` eager was indeed a silent, previously-untested beneficiary of this fix, and it's the case that exercises `_scan_axes` excluding exactly one `ScanAxis` (as opposed to `GridScan`'s two or `CustomScan`'s zero). Renamed `TestGridScanEnergyEnsembleAxisOrder` to `TestScanEnergyEnsembleAxisOrder` and parametrized the scan construction rather than adding a separate class, since `GridScan` and `LineScan` share the same energy-leading axis convention and test body.

`LineScan` + lazy is excluded from the new coverage (`xfail`, `strict=True`) as the separate, pre-existing defect this review's own residual note already pointed at: it raises `IndexError` from `multi_output_blockwise` regardless of this fix. The causal chain the review left open as unproven is confirmed: `_out_ensemble_shape`'s `ensemble_shape[:-2]` (`abtem/detectors.py`) unconditionally strips the last two ensemble axes, which over-strips whenever only one is actually a `ScanAxis`, exactly `LineScan`'s case.

Confirmed the three newly-passing combinations (`grid` eager, `grid` lazy, `line` eager) fail without this fix; the `xfail`ed `line`+lazy combination fails identically with or without it. Full file: 163 passed, 4 xfailed.

TomaSusi (approved):
Verified `aed9d955` — approving.

Parametrizing the existing class rather than adding a second one was the better call; `GridScan` and `LineScan` genuinely do share the energy-leading convention and the test body, so a separate class would have been duplication.

Runs exactly as described:

```
[order0-grid-False] PASSED   [order0-line-False] PASSED   [order0-line-True] XFAIL
[order0-grid-True]  PASSED   [order1-line-False] PASSED   [order1-line-True] XFAIL
...
6 passed, 2 xfailed
```

And it discriminates: reverting `new_energy_axis` back to `energy_axis_idx` gives **6 failed, 2 xfailed** — the two `xfail`s stay `xfail` (they fail for the unrelated `multi_output_blockwise` reason either way), while every newly-covered combination flips. That is the right signature for this fix, and it confirms the `line` eager case is now genuinely guarded rather than incidentally passing.

Thanks for closing out the causal chain I left open — `_out_ensemble_shape`'s unconditional `ensemble_shape[:-2]` over-stripping when only one axis is a `ScanAxis` is the confirmation I couldn't make myself, and `xfail(strict=True)` is the right way to hold that case until it's fixed.

### One thing to watch at merge time, not a change request here

This PR and #440 are each green on their own but **fail when merged together** — 2 failures, in #440's own `test_gridscan_multiblock_reproduces_its_own_standalone_run`, whose assertion encodes the energy-*trailing* convention this PR corrects to energy-*leading*. There is no textual conflict, so nothing will warn at merge. Details and the underlying issue are in my comment on #440; flagging here only so whichever lands second isn't merged on a stale green.

---

PR #440, opened 2026-09-20
Fix coreloss customscan partition

pzeiger:
## Summary

`Probe(energy=[...]).transition_potential_scan(..., lazy=True)` crashed with `ValueError: too many values to unpack (expected 2)` as soon as the waves ensemble spanned more than one dask block and the potential had no ensemble of its own (no `FrozenPhonons`) — for any scan type, not just `CustomScan`. A plain `GridScan` with the same trigger crashes identically.

## Root cause

Once a waves ensemble spans more than one dask block, `_prebuild_reused_potential` (`abtem/waves.py`) eagerly builds an otherwise-unbuilt potential into a `PotentialArray` so it is computed once and shared across blocks rather than rebuilt per task. That path exposed two compounding bugs in `WavesBuilder` (`abtem/waves.py`), both in the machinery that combines a probe's independent sub-ensembles (scan positions, energy, aperture, ...) into one blockwise ensemble array:

1. `_from_partitioned_args_func` wrapped each per-block reconstructed probe via `_wrap_with_array(new_probe)` with no explicit `ndims`, defaulting to `len(new_probe.ensemble_shape)`. A reconstructed sub-ensemble can keep non-trivial structure even once resolved to a single block — a `GridScan` position stays a `(1, 1)` grid, not a squeezed `()`, unlike a resolved scalar energy value — so this didn't reliably reflect how many combined ensemble axes the call was actually filling.
2. `_build_validated`'s `da.map_blocks` call assumed `_calculate_array`'s return shape always matched `block.shape + gpts`. `block.shape` is always `(1,)*ndims` for this object-dtype array regardless of how many logical ensemble values that block's declared chunk actually covers (`ensemble_blocks`'s own `adjust_chunks` can make those differ — a single unsplit block covering an entire axis, for example) — so this broke for the resolved-per-block path once the first bug no longer masked it.

## Fix

1. Thread the real, fixed combined-ensemble axis count through from `_from_partitioned_args` into `_from_partitioned_args_func`, and use it explicitly instead of the reconstructed probe's own (unreliable) `ensemble_shape`.
2. Read the declared chunk shape for the current block from `da.map_blocks`'s own `block_info` instead of `block.shape`, and reshape `_calculate_array`'s result to match. Inserting or resizing size-1 axes via reshape never reorders data, so this is safe, and a genuine shape mismatch still raises rather than silently reinterpreting wrong data.

## Tests

`test/test_core_loss_fixes.py`: `TestUnbuiltTransitionPotentialEnergyEnsemble`'s lazy case, previously marked `xfail(strict=True)` and documented as a pre-existing, unrelated defect, now passes and is un-xfailed. Added a `GridScan` counterpart confirming the fix is scan-independent. Both confirmed to fail without the fix (all 4 lazy parametrizations; eager was always correct).

Verified against a broad sweep given how central `WavesBuilder` is to wave-function construction: `test_core_loss_fixes.py`, `test_waves.py`, `test_scan.py`, `test_detect.py`, `test_measure.py`, `test_prism_upsample.py`, `test_energy_ensemble.py`, `test_multislice_input.py`, `test_realspace_multislice.py`, `test_potentials.py`, `test_potential_chunking.py`, `test_commensurate_potential.py`, `test_ionization.py`, `test_transition_potential_transport.py`, `test_prism.py`, `test_atomic_potential.py`, `test_phonon_loss.py` — all green, roughly 985 tests, no regressions. `flake8`/`mypy` finding counts on the touched file are unchanged from baseline.


TomaSusi (commented):
The `ndims` threading is well-diagnosed, and the `_wrap_with_array(combined, ndims)` correction in `array.py` (replacing a hardcoded `1`) looks right — the previous value only happened to be correct at `ndims == 1`. Full suite on this branch alone: **1765 passed, 858 skipped**.

Two things, one of which only shows up in combination with #439.

## This PR and #439 are each green alone, and fail when merged together

Merging both onto current `dev` produces **no textual conflict** — and then **2 failures**, both in this PR's own new test:

```
FAILED TestUnbuiltTransitionPotentialEnergyEnsemble::test_gridscan_multiblock_reproduces_its_own_standalone_run[order0-False]
FAILED TestUnbuiltTransitionPotentialEnergyEnsemble::test_gridscan_multiblock_reproduces_its_own_standalone_run[order1-False]
```

The cause is the assertion's axis convention:

```python
# GridScan on unfixed dev (this branch, no PR #439) also stacks
# energy trailing -- same axis convention as CustomScan here.
np.testing.assert_allclose(ensemble[..., i], reference[e], ...)
```

The comment is accurate about today's `dev`, but #439 is precisely the change that makes `GridScan` stack energy *leading*, so `ensemble[..., i]` inverts the moment it lands. Since nothing conflicts textually, neither PR's CI will warn — whichever merges second goes in on a stale green.

## The part that matters more: the lazy multiblock path is still genuinely wrong

Worth separating the test breakage from what it exposes. With both PRs applied, I compared the data's real axis order against what the measurement's own `ensemble_axes_metadata` reports, across the four `GridScan` configurations:

| config | metadata reports | data actually is |
|---|---|---|
| eager, default batch | EnergyAxis leading | leading |
| eager, `max_batch=1` | EnergyAxis leading | leading |
| lazy, default batch | EnergyAxis leading | leading |
| **lazy, `max_batch=1`** | EnergyAxis leading | **trailing** |

So the multiblock lazy path — the one this PR's `_prebuild_reused_potential` mechanism is about — still has exactly the metadata/data mismatch #439 exists to fix. #439's change to `_calculate_new_array`'s stacking doesn't reach it.

That makes the failing assertion more than a merge nuisance: as written, this PR's new test would codify the surviving bug as the expected behaviour for the one configuration where it persists. It is invisible on this branch only because, without #439, *every* `GridScan` path is trailing, so the test is self-consistent.

I'd suggest the eager/lazy split here is itself the signal worth acting on — a test asserting one axis convention for eager and the opposite for lazy, at the same `max_batch`, is describing a bug rather than a contract.

## Smaller: the reshape's safety claim is stronger than what reshape guarantees

```python
# Inserting/resizing size-1 axes via reshape never reorders data, so this
# is safe, and a real shape mismatch still raises here instead of silently
# reinterpreting wrong data.
```

The first half holds; the second doesn't. `reshape` raises only when the **total size** differs. A same-total-size but differently-arranged shape succeeds silently:

```
(2,3,2) -> (1,2,3,2)   size-1 insert, order preserved
(2,3,2) -> (3,2,2)     no error, data silently remapped
```

Given this sits on `WavesBuilder`'s central path and the whole point is to avoid silently reinterpreting wrong data, asserting the invariant the comment already states would make it real — something like requiring the non-1 dimensions to match before reshaping, so anything that isn't a size-1 adjustment raises instead of being reinterpreted. Same shape of fix as the power-of-two assertion on #438.

Holding off on approving pending the #439 interaction — the underlying `ndims` fix I'm satisfied is correct.

---

pzeiger:
Thanks for the thorough pass. Taking the three points in turn.

**Reshape safety comment (fixed, `dfb0ade3`).** Agreed on the counter-example: `(2, 3, 2)` reshapes into `(3, 2, 2)` without error, silently remapping real data across axes at equal element count, so the old comment's blanket claim was wrong. Added an explicit assertion that every non-1 dimension of `result.shape` and `expected_shape` already matches, in order, before the reshape runs — matching #438's own precedent for guarding a reshape with an assertion rather than a comment alone. All 164 tests in `test_core_loss_fixes.py` still pass.

**Test assertion needs updating once #439 merges (confirmed, not yet actionable).** You're right that `test_gridscan_multiblock_reproduces_its_own_standalone_run`'s `ensemble[..., i]` (trailing) indexing is specific to GridScan's pre-#439 eager behavior, and #439 changes that. Deliberately not touching this now: the correct fix depends on merge order between #439 and this branch, and — see the next point — the "leading" position #439 declares for GridScan isn't actually reliable yet for every case this test exercises, so a metadata-driven fix here would just be trading one hardcoded assumption for another. Will update this assertion as part of whichever PR lands second.

**Deeper finding: confirmed, root-caused, and now has a working fix — but it's bigger than this PR and needs a routing decision.** Reproduced your 4-row table exactly. Root cause: `multi_output_blockwise`'s reordering (the `ensemble_sources`/`_out_ensemble_source` hook, which lives on the third branch, `fix-coreloss-lazy-ensemble-chunk-order`, not originally this PR) only relabels the *declared chunk-size tuple* it hands to `da.map_blocks` — it never physically transposes the underlying dask array, because `da.blockwise`'s own axis-symbol assignment stays in the waves' natural (unreordered) axis order throughout. For a single dask block this was invisible, because #439's `_calculate_new_array` fix already reorders the data at the Python level before this code runs. For `max_batch=1` (multiple blocks), that per-block reorder never triggers — each block already resolves to exactly one energy value — so the real data stayed in natural order while the declared metadata claimed otherwise. Silent whenever natural and declared chunk structures are structurally identical (your 2×2 GridScan / 2-energy case), and a crash (`Dimension i has N blocks, adjust_chunks specified with M blocks`) whenever they're not (also confirmed with a 3-energy case on the same 2×2 scan).

An unconditional transpose after `map_blocks` alone regresses the single-block case, since it double-reorders data #439's own logic already placed correctly — the reordering responsibility needs to live in exactly one place, not split between #439's Python-level logic and the dask-level chunk bookkeeping. Implemented that: simplified `_calculate_new_array` to always stack at its own natural axis position (no declared-order awareness at all), and moved the one, uniform reorder-to-declared-order step into `ArrayObject.apply_transform`, applied identically to both the eager and lazy result via an actual `.transpose()`. Folded `fix-coreloss-lazy-ensemble-chunk-order`'s commits into this branch and built the fix there; all 176 tests in `test_core_loss_fixes.py` pass, your original 4-row table now reads all-leading, and a broader sweep (2 and 3 energies, symmetric and asymmetric scan sizes, all four eager/lazy × batch combinations) confirms it, including cases your table didn't cover.

The catch: part of it touches `_calculate_new_array` on #439, which is already approved. Pushed the whole thing here on this branch for now, including that piece, so it's reviewable and the numbers above are reproducible — but where it should permanently live is still open: a small follow-up on #439's own branch once #439 merges (letting the two PRs' diffs interact normally at merge time), or staying folded into this PR as-is. Either way the fix itself is ready for review now.

TomaSusi:
Just as a personal note: although it may be good programming practice to make minimal PRs, I would prefer larger ones that cover all interconnected parts of the fixes (like the connection between 339 and 440 mentioned above). In this case maybe stacking the two would make sense?

pzeiger:
Good point @TomaSusi , I am also struggling a bit with this amount of PRs. I have updated my instructions to Claude to create larger PRs for abTEM if issues are related and/or similar files are touched. It should also do it, even if a PR is already approved and later additional problems become apparent. I think that was the issue here and what I wanted to hear your call on, if the approval of the other PR is more important or limiting the amount of PRs and I got my answer :).

On a side note: I am very close to done with all the issues Claude and I had identified, so the PR frequency should drop dramatically from my side soon. =)

pzeiger:
Agreed on consolidating rather than keeping this minimal — retargeting this PR's base onto `fix-coreloss-multiposition-corruption` (#439) to make that explicit.

Turns out this branch already stacks on #439 in practice: `fix-coreloss-lazy-ensemble-chunk-order` (the third branch folded in above) was originally branched off #439's own branch, not off `dev`, so #439's fix has been an ancestor of this branch all along. Merged in #439's latest commit as well, so this branch is now fully caught up — `git merge-base --is-ancestor origin/fix-coreloss-multiposition-corruption HEAD` confirms a clean, linear ancestry with no conflicts.

That merge surfaced one more thing worth mentioning: it pulled in a `strict=True` xfail for a pre-existing `LineScan` + lazy crash (`ensemble_shape[:-2]` over-stripping in `_out_ensemble_shape`), and that case now genuinely passes — the same axis-reordering fix already in this PR replaced that unconditional slicing with the dynamic, `_scan_axes`-based version. Removed the stale marker. `test_core_loss_fixes.py`: 180 passed.

TomaSusi (commented):
All three points addressed, and the architectural fix is better than I expected. One new blocker, created by the fix reaching further than claimed.

## Verified

**Reshape assert (`dfb0ade3`).** Present and correct: non-1 dimensions of `result.shape` and `expected_shape` must already match, in order, before the reshape is allowed to run. The invariant the comment stated is now enforced rather than assumed.

**The multiblock axis mismatch.** Reproduced your result and went wider than my original 4-row table — 16 combinations, 2x2 and 2x3 scans, 2 and 3 energies, eager and lazy, default `max_batch` and `max_batch=1`:

```
scan(2, 2) n_e=2  lazy=False mb=None   shape=(2, 2, 2)  LEADING=True
scan(2, 2) n_e=2  lazy=True  mb=1      shape=(2, 2, 2)  LEADING=True
scan(2, 2) n_e=3  lazy=True  mb=1      shape=(3, 2, 2)  LEADING=True
scan(2, 3) n_e=2  lazy=True  mb=1      shape=(2, 2, 3)  LEADING=True
scan(2, 3) n_e=3  lazy=True  mb=1      shape=(3, 2, 3)  LEADING=True
  ... all 16 combinations: LEADING=True
```

All correct, including the asymmetric-scan and 3-energy cases that previously either crashed or silently mismatched. Suite on this branch alone: **1777 passed, 858 skipped**.

**The `apply_transform` transpose is well-bounded.** This was my main worry about moving the reorder onto such a central path, and `_transpose_to_ensemble_source`'s identity short-circuit answers it: a transform whose `_out_ensemble_source` is not a real permutation returns the array untouched, so only detectors that actually move scan axes are affected. Putting the responsibility in exactly one place rather than splitting it between Python-level stacking and dask-level chunk bookkeeping is the right call, and the reasoning about `da.blockwise` never permuting `out_ind` relative to its own symbols matches what I see in the code.

**Your own `ensemble[..., i]` assertion** turns out to be updated to leading already, with the CustomScan one correctly left trailing — so the failure I reported last round is gone.

## New blocker: this now fixes the case #439 declares permanently broken

Lazy `LineScan` works on this branch:

```
lazy line n=2: OK shape=(2, 2)        (previously: IndexError: tuple index out of range)
```

That is exactly the combination #439 marks `xfail(strict=True)`. Merging both onto current `dev` is textually clean — no conflict — and then fails:

```
2 failed, 1779 passed, 858 skipped
[XPASS(strict)] pre-existing, unrelated defect: LineScan + lazy raises IndexError(...)
```

Both failures are that XPASS, in `TestScanEnergyEnsembleAxisOrder[order0-line-True]` and `[order1-line-True]`.

Worth flagging that this is the **inverse** of what I found last round: when I checked the earlier version of this branch, it did *not* fix lazy LineScan, so #439's strict xfail was safe. The centralized-transpose rework changed that. Since nothing conflicts textually, neither PR's CI will warn.

## On the routing question

I'd keep it here rather than moving the `_calculate_new_array` piece onto #439.

#439 is approved, small, and self-contained, and its `new_energy_axis` computation ships with a test and a comment explaining that specific mechanism. Reopening it to replace that with "always stack at the natural position" would invalidate both the approval and the explanation, for a change that only makes sense in the presence of the centralized transpose this PR adds. Splitting a single coherent fix across two branches to satisfy merge order seems worse than letting the diffs interact normally.

That does put two obligations on this PR, since it is the one superseding:

1. **Remove #439's `xfail(strict=True)`** once #439 lands — it cannot be done from here today, since the marker exists on neither `dev` nor this branch.
2. **State in the description that this supersedes #439's `_calculate_new_array` mechanism.** Without that, the history reads as an approved fix being quietly reverted a day later, and the next person to `git log` that function has no way to tell that the replacement was deliberate.

Suggested order: merge #439, then merge `dev` into this branch, drop the now-obsolete xfail, re-run CI. The current green here is on a tree that does not contain #439, so it should not be treated as evidence for the merged result.

Not approving yet purely on the XPASS interaction — the fix itself I am satisfied is correct, and the sweep above is the evidence.

---

TomaSusi (approved):
**Approving** — both findings closed, and restacking onto #439 was the better answer to the routing question than either option I laid out.

Retargeting the base solves the merge-order problem structurally instead of by coordination: the two PRs' diffs now interact at review time rather than at merge time, so there is no window where one lands on a green that was computed without the other. Verified the stack is real rather than nominal — #439's branch is a true ancestor of this head, and `dev` is an ancestor of the stack.

Verified on `12d13850`:

- **The XPASS blocker is gone.** #439's `xfail(strict=True)` on `line`+lazy is removed, and `TestScanEnergyEnsembleAxisOrder` is now **8 passed, 0 xfailed** — that case is genuinely covered and passing rather than declared broken.
- **Full suite on the stack: 1781 passed, 858 skipped, 0 failed**, no XPASS.
- **My 16-combination sweep: 0 incorrect** — 2x2 and 2x3 scans, 2 and 3 energies, eager and lazy, default `max_batch` and `max_batch=1`. Every one leading, matching metadata.
- **All four residual shapes work**, including `line`+lazy and `grid`+3-energies, which previously raised `IndexError` and `RuntimeError` respectively.
- **No orphaned references** to the superseded `new_energy_axis` mechanism, and `multislice.py` is net -8, consistent with the simplification genuinely replacing #439's logic rather than layering on top of it.

The reshape assertion from the previous round is in place, and `_transpose_to_ensemble_source`'s identity short-circuit is what makes moving the reorder onto `apply_transform`'s central path safe — a transform whose `_out_ensemble_source` is not a real permutation returns the array untouched, so only detectors that actually move scan axes are affected. Concentrating the reordering in one place, rather than splitting it between Python-level stacking and dask-level chunk bookkeeping, is the right structure and is why the `max_batch=1` case now works at all.

Worth noting for whoever merges: since this is stacked, #439 lands first and this retargets to `dev` automatically. No separate coordination needed, which was the point.

---

PR #441, opened 2026-09-21
Cap Numba's parallel thread pool to OMP_NUM_THREADS when set

pzeiger:
## Summary

Every `@njit(parallel=True)` kernel in abTEM — the CPU Laplacian stencil in `finite_difference.py`, the magnetism kernels in `magnetism/pauli.py`, the partitioned S-matrix kernel in `prism/_partitioned_s_matrix.py` — uses Numba's own thread pool, which defaults to the full visible CPU count and reads neither `OMP_NUM_THREADS` nor `OPENBLAS_NUM_THREADS`. Numba's default "workqueue" threading layer (the fallback when neither TBB nor OpenMP is available) is independent of both.

Several such processes sharing one node each independently try to claim every core for themselves. Measured on a Perlmutter GPU node (1x AMD EPYC 7763, 64 physical / 128 logical cores, `OMP_NUM_THREADS=1` and `OPENBLAS_NUM_THREADS=1` already set): `test_realspace_multislice.py` under `pytest -n auto` took 100-400s per `[cpu]`-parametrized test versus 2-8s for the equivalent `[gpu]`-parametrized test (the GPU path dispatches to a separate CUDA kernel and never touches Numba at all, so it was never affected by this). All logical cores stayed at 100% load regardless of the xdist worker count (`-n 8` showed the identical symptom), confirming the oversubscription is per-process rather than a function of how many worker processes are running.

This applies `OMP_NUM_THREADS` to Numba's thread pool too, once, at `abtem.core.backend` import time — the same convention already used everywhere else in scientific Python for BLAS/OpenMP thread limits, rather than introducing a new abTEM-specific setting. With the fix, the same suite dropped from 420s to 43s total, with every previously affected `[cpu]` test down to 2-4s, in line with its GPU-path counterpart.

## Behavior

- `OMP_NUM_THREADS` unset: no change — Numba keeps its existing default (full visible CPU count).
- `OMP_NUM_THREADS` set, `NUMBA_NUM_THREADS` unset: Numba's thread pool is capped to the same value.
- `NUMBA_NUM_THREADS` already set explicitly: left alone, so a deliberate Numba-specific choice is never overridden.
- `OMP_NUM_THREADS` set above Numba's own launch-time thread ceiling (`numba.config.NUMBA_NUM_THREADS`, itself derived from the visible CPU count): clamped to that ceiling rather than raising `ValueError` at import time.

## Testing

- `test/test_backend.py` (new): covers all four cases above directly against `_cap_numba_threads_to_omp_num_threads`, since it runs once as an import-time side effect.
- `test/test_realspace_multislice.py`: unchanged, passes (41 passed, `-k "not gpu"`, 19.9s locally).

## Scope

The `parallel=True` kernels themselves are untouched — this only bounds how many threads they may use, so single-process CPU performance is unaffected when nothing sets `OMP_NUM_THREADS`.


TomaSusi (commented):
The diagnosis is right and well-evidenced — Numba's workqueue layer genuinely ignores both `OMP_NUM_THREADS` and `OPENBLAS_NUM_THREADS`, the per-process (rather than per-worker) oversubscription signature is convincing, and the 420s → 43s measurement is real. Deferring to the existing `OMP_NUM_THREADS` convention instead of inventing an abTEM-specific setting is the right call, as is leaving an explicit `NUMBA_NUM_THREADS` alone.

The mechanism, though, only reaches the thread it runs on.

## `numba.set_num_threads()` is thread-local

```
numba 0.63.1, set_num_threads(2) called on the main thread:
  main thread   get_num_threads() = 2
  worker thread get_num_threads() = 20
```

Confirmed against a real `@njit(parallel=True)` kernel (counting distinct `get_thread_id()` values inside a `prange` loop, not just reading the setting back): 2 worker threads when called from the main thread, 20 from a spawned thread.

On this branch, with `OMP_NUM_THREADS=2`:

```
main thread (what test_backend.py checks): 2
dask threaded-scheduler tasks:            20   <- uncapped
dask synchronous scheduler:                2
```

And in an actual abTEM lazy computation, tasks execute on both `MainThread` and `ThreadPoolExecutor-0_0`, observing `{2, 20}`.

That matters here specifically because `abtem/array.py` forces the synchronous scheduler only for **GPU** computation — so a CPU `lazy=True` run goes through dask's threaded scheduler, which is exactly where the `parallel=True` CPU kernels this PR names (the `finite_difference.py` Laplacian, the `magnetism/pauli.py` kernels) actually execute. The configuration the fix most needs to reach is the one it doesn't.

This does not contradict the measured speedup: pytest executes tests on each worker process's main thread, so the eager `[cpu]` tests that dominated that run were genuinely capped. The gap is narrower than the fix looks, and invisible from the tests, since `test_backend.py` exercises `_cap_numba_threads_to_omp_num_threads` directly on the main thread — the one thread where it does work.

## A verified alternative

Seeding `NUMBA_NUM_THREADS` from `OMP_NUM_THREADS` *before* numba is first imported is process-global rather than thread-local:

```
proposed  -> numba threads observed in abTEM lazy tasks: {2}
this branch -> numba threads observed in abTEM lazy tasks: {2, 20}
```

I checked that this is actually placeable: `numba` is **not** in `sys.modules` before `abtem/__init__` runs, so an `os.environ.setdefault("NUMBA_NUM_THREADS", ...)` early enough in `abtem/__init__.py` takes effect before `backend.py`'s own `import numba`. `setdefault` also preserves the existing "never override an explicit `NUMBA_NUM_THREADS`" behaviour for free.

Worth keeping the `set_num_threads` call as a fallback rather than replacing it: if a caller imported numba before abtem, the env var can no longer take effect, and the current approach still caps the main thread in that case. Two mechanisms, with the limitation of each stated, seems better than either alone — the one thing worth avoiding is the current situation, where the limitation isn't visible anywhere.

## On the tests

`test_backend.py`'s four cases are the right cases for the helper's own logic, and I'd keep them. What's missing is a test of the property that actually matters — that the cap holds where the kernels run. Asserting `get_num_threads()` inside a worker thread (or inside a `dask` threaded-scheduler task) would have caught this, and would keep catching it if the mechanism changes.

Full suite on this branch: **1767 passed, 858 skipped, 2 xfailed** — no regressions. The issue is reach, not correctness, so nothing here is a reason to revert what's already working.

---

pzeiger:
Verified directly and confirmed — `numba.set_num_threads()` only rebinds the calling thread's own active count. After `set_num_threads(2)` on the main thread, a real `@njit(parallel=True)` kernel launched from a spawned worker thread still executed with the full visible core count of threads, not 2 (checked via `get_thread_id()` inside the kernel, not just `get_num_threads()`'s reported value). Since dask's threaded scheduler runs a CPU-side lazy computation on spawned worker threads rather than the caller's own thread, that is exactly the configuration this fix needed to reach and didn't.

`e81a2a95` seeds `NUMBA_NUM_THREADS` from `OMP_NUM_THREADS` itself, before numba's own first import anywhere in the package (`abtem.core._numba_threads`, imported as `abtem/__init__.py`'s own first statement, ahead of `abtem.core.backend` — the module that first imports numba, reached transitively via `abtem.distributions`). Reading the env var at numba's own import time applies process-wide, to every thread its workqueue layer ever launches — confirmed against the same spawned-worker-thread reproduction: with `OMP_NUM_THREADS=2`, both the main thread and a spawned worker thread now execute the kernel across exactly 2 distinct thread ids, versus the worker thread spreading across the full visible core count beforehand.

The previous `set_num_threads()` call stays as a fallback, for a caller that imports numba before abtem, when the env var can no longer take effect.

One more edge case surfaced while implementing this: numba's own import-time handling of `NUMBA_NUM_THREADS` isn't bounds-checked against the visible core count, and an implausibly large value segfaults the process outright the first time a parallel kernel launches, rather than raising a catchable exception the way `set_num_threads()` does above its own launch ceiling. Clamped to `os.cpu_count()`.

`test/test_numba_threads.py` adds the missing property test via a subprocess (env-var seeding only takes effect at process start, so the existing process's already-imported numba can't be used to test it): the cap now reaches a spawned worker thread's kernel execution, not just the thread that imported abtem.

That commit's own test additions then broke CI across all four Python versions:

```
RuntimeError: Cannot set NUMBA_NUM_THREADS to a different value once the threads have been launched (currently have 3, trying to set 4)
```

Root cause: `test/test_numba_threads.py`'s `TestSeedNumbaNumThreadsFromOmp` called `_seed_numba_num_threads_from_omp()` directly against the real process `os.environ`, protected only by `monkeypatch.setenv`/`delenv` on the individual keys it cared about. `monkeypatch.delenv(key, raising=False)` on a key that isn't currently set does not protect a later *raw* write to that same key made by the function under test — confirmed directly, running a single one of those tests alone left `NUMBA_NUM_THREADS="3"` in the real environment after the test process returned.

That leak reaches a real numba compilation elsewhere in the same process: CI runs the whole suite with plain `pytest test`, no `-n`, so every test shares one process. `numba.core.config.reload_config()` — called from `numba/core/compiler.py` on every new JIT compilation — re-reads `NUMBA_NUM_THREADS` and raises exactly this error if it differs from whatever value the parallel thread pool already launched with. Any test file collected after this one alphabetically that triggered its first `@njit(parallel=True)` compilation hit it — matching the actual failing tests (`test_potential_parametrization.py`, `test_realspace_multislice.py`).

Fixed in `2498fddc`: `os.environ` is swapped for a disposable copy for the duration of each test (an autouse fixture), rather than tracking individual keys — guarantees the function's real writes can never reach the actual process environment, regardless of which keys it touches.

Verified by running the entire non-GPU test suite locally in one process, matching CI's exact invocation: 1772 passed, zero `NUMBA_NUM_THREADS` errors. All four CI checks now pass (3m54s-5m13s each).

TomaSusi (approved):
**Approving** — `e81a2a95` closes the thread-locality gap, and the two issues you found along the way were both worth finding.

Verified end-to-end on `2498fddc`, the same reproduction that showed the gap before:

```
OMP_NUM_THREADS=2 -> seeded NUMBA_NUM_THREADS=2, numba ceiling 2
numba threads seen in abTEM lazy tasks: {2}        (previously {2, 20})
thread names: ['MainThread', 'ThreadPoolExecutor-0_0']
```

The dask worker thread now sees the cap, which was the whole point. The implementation is careful in the places it needed to be: `_numba_threads` is `abtem/__init__.py`'s first statement with an `I001` noqa so an isort pass cannot silently reorder the requirement away, the module itself never imports numba, `os.cpu_count() or 1` handles `cpu_count()` returning `None`, and `set_num_threads` stays as the fallback for a caller that imported numba first — which the env var genuinely cannot reach.

I ran the **CI-shaped invocation specifically** (`pytest test`, single process, no `-n`), since that is the shape the env leak needed to manifest: **1777 passed, zero `NUMBA_NUM_THREADS` errors**. Swapping `os.environ` for a disposable copy via an autouse fixture is the right fix — tracking individual keys could never have been sound against a function whose whole job is writing an arbitrary key.

The `NUMBA_NUM_THREADS=99999` segfault is a good catch, and worth stating plainly for the archaeology: numba raising cleanly from `set_num_threads` but dumping core from the env var is not an obvious asymmetry, and clamping to `os.cpu_count()` is the right guard.

### One non-blocking documentation point

The Scope section says single-process CPU performance is unaffected "when nothing sets `OMP_NUM_THREADS`" — accurate, but the converse now has teeth:

```
OMP_NUM_THREADS unset -> numba threads: 20
OMP_NUM_THREADS=1     -> numba threads: 1
```

HPC schedulers routinely set `OMP_NUM_THREADS=1` in the job environment by default. A *single* abTEM process there now runs the `parallel=True` kernels single-threaded, where before it used every visible core. That is the correct reading of the convention and I would not change the behaviour — but it is a real change for single-process runs in exactly the environment this was measured in, where the oversubscription win came from many processes sharing a node. Worth a sentence, and probably in `_numba_threads.py`'s own docstring rather than only the PR description, since that is where someone will look when a single-process run is unexpectedly slow.

Not blocking the merge on it.

---

TomaSusi:
Following up on the non-blocking note in my approval, with measurements rather than the bare assertion — recording them here so the trade-off is on file rather than needing to be re-derived later. **Nothing here changes the approval**; the conclusion is that the behaviour change is close to free on every path anyone routinely runs.

Measured on an Intel Xeon W-2150B (10 physical / 20 logical cores, quad-channel DDR4, 64 GB), numba 0.63.1, numpy 2.3.5, at `2498fddc`.

### 1. The default multislice path never reaches a `parallel=True` kernel

Standard Fourier multislice scan — 3x3x6 bulk Si, `gpts=(512, 512)`, `slice_thickness=2.0`, 8x8 `GridScan`, `AnnularDetector(50, 150)`, eager:

```
NUMBA_NUM_THREADS=1   ->  4.81 s
NUMBA_NUM_THREADS=20  ->  4.85 s
```

Indistinguishable. Ordinary multislice and PRISM are FFT-bound; the cap cannot affect them at all, whatever `OMP_NUM_THREADS` says.

### 2. Only three kernels are exposed, on non-default paths

`grep -rn "parallel=True" abtem/` gives exactly:

- `finite_difference.py:278` — the Laplacian stencil, reached only via `RealSpaceMultislice`, not the default `FourierMultislice`
- `magnetism/pauli.py:8,52` — the Pauli gradient kernels (non-collinear magnetic multislice)
- `prism/_partitioned_s_matrix.py:121` — the partitioned S-matrix kernel

(`integrals.py:980` already carries a comment explaining why it is deliberately *not* `parallel=True`.)

### 3. Those kernels scale poorly anyway, being memory-bandwidth bound

| kernel | 1 thread | 20 threads | speedup |
|---|---|---|---|
| `central_difference_gradient_pbc`, `X` float64 `(64, 512, 512)` | 0.258 s | 0.203 s | 1.27x |
| `_laplace_stencil_cpu_batch`, accuracy 4, complex64 `(32, 512, 512)` | 0.101 s | 0.087 s | 1.16x |

So the worst realistic case — magnetic or real-space work on a node whose scheduler exports `OMP_NUM_THREADS=1` — costs roughly 15-25% on those kernels, not a factor of the core count. Against the ~50x oversubscription harm this PR removes (100-400 s versus 2-8 s per test), the trade is heavily favourable.

### 4. What these numbers do *not* cover

A 10-core, quad-channel Xeon is close to the worst case for demonstrating parallel scaling in a bandwidth-bound kernel. An EPYC 7763 has 8 memory channels and 64 physical cores, so both kernels may scale considerably better there — which would make the single-process cost of an inherited `OMP_NUM_THREADS=1` correspondingly larger than 1.16-1.27x suggests. Anyone running Pauli or real-space multislice work single-process on Perlmutter is the one case worth measuring before assuming this is free; the test-suite measurements in this PR were many processes sharing a node, which is the opposite regime.

### 5. The escape hatch, worth stating explicitly

Because this PR leaves an explicit `NUMBA_NUM_THREADS` alone, a job that wants the old behaviour back overrides it per-job without touching code or unsetting the scheduler's `OMP_NUM_THREADS`:

```bash
export NUMBA_NUM_THREADS=64   # overrides an inherited OMP_NUM_THREADS=1
```

That is the part most worth having in `_numba_threads.py`'s docstring alongside the rest, since it is what someone needs when a single-process run is unexpectedly slow.

---

pzeiger:
Documented in `5ac17e70`, in `_numba_threads.py`'s own module docstring per the non-blocking note — including the escape hatch (`export NUMBA_NUM_THREADS=<N>`).

One addition to the picture in the measurement comment above, since the single-caller numbers there (1.16-1.27x) are from one isolated call, and abTEM's typical execution shape is many concurrent dask worker threads, not one:

Measured on a 24-core box, same `_laplace_stencil_cpu_batch` kernel, 24 concurrent callers (`ThreadPoolExecutor`) each repeatedly invoking it, sweeping `NUMBA_NUM_THREADS` from 1 to 24 for the whole batch:

```
NUMBA_NUM_THREADS=1   -> 96 calls in 2.124 s
NUMBA_NUM_THREADS=2   -> 96 calls in 2.092 s
NUMBA_NUM_THREADS=4   -> 96 calls in 2.130 s
NUMBA_NUM_THREADS=8   -> 96 calls in 2.137 s
NUMBA_NUM_THREADS=24  -> 96 calls in 2.110 s
```

Statistically indistinguishable. Once 24 independent tasks already saturate the node, giving each one its own internal thread pool on top has nothing left to parallelize into — the single-caller cost this PR trades away doesn't survive contact with the concurrency abTEM actually runs under day to day.

Also swept the single-caller case more finely than 1-vs-20, since a memory-bandwidth-bound kernel need not scale monotonically with thread count: on the same box, `_laplace_stencil_cpu_batch` peaks around 8-12 threads (29.5 ms/call) and is *slower* again at the full 24 (30.8 ms/call) than at 12 — the same non-monotonic shape already known from BLAS diagonalization kernels, just reached earlier here. Worth knowing if `NUMBA_NUM_THREADS` is ever tuned deliberately rather than left to inherit `OMP_NUM_THREADS`: more is not better past a fairly small number, for these specific kernels.

TomaSusi:
Reproduced the concurrency result independently, on different hardware — it comes out **stronger** than "statistically indistinguishable", which I think is worth recording since the docstring now leans on that line.

Same `_laplace_stencil_cpu_batch` kernel (accuracy 4, complex64 `(8, 512, 512)`), 20 concurrent callers via `ThreadPoolExecutor`, 4 repeats each (80 calls total), three trials per setting, on the 10-core/20-thread Xeon W-2150B from my earlier comment:

| `NUMBA_NUM_THREADS` | trials (s) | min |
|---|---|---|
| 1 | 0.668, 0.621, 0.610 | **0.610** |
| 4 | 0.765, 0.721, 0.711 | 0.711 |
| 20 | 0.812, 0.721, 0.736 | 0.721 |

On your 24-core box the settings were a wash; here `=1` is consistently **15-18% faster** than either higher setting, with the per-setting ranges not overlapping across three trials — so it is an ordering, not noise.

The mechanism is presumably the same oversubscription this PR exists to fix, just occurring *within* one process rather than across several: 20 concurrent callers each claiming their own N-thread internal pool, contending for 20 cores. Which means the cap is not merely free under abTEM's actual execution shape — it is a modest win there.

Two things follow that might be worth a word in the docstring, since it currently says the cost "disappears entirely" under concurrency:

- "Disappears entirely" is the conservative reading and holds on both boxes. On a machine with fewer cores relative to the concurrency level, capping is actively faster, so nobody is trading throughput away by leaving `OMP_NUM_THREADS` to propagate.
- It reinforces your non-monotonicity point from the other direction: more threads per kernel is not better, and once concurrency is in the picture the optimum moves further *down*, not up. The single-caller peak at 8-12 threads is an upper bound on what deliberate tuning could buy, and only for a genuinely serial, single-call workload.

The single-process/single-call caveat I raised earlier still stands unmeasured on an EPYC 7763 specifically, but this makes it considerably less interesting: the regime where the cap could cost anything is the one abTEM is least often in.

Approval unchanged — `5ac17e70` is docstring-only and reads well, and putting it in `_numba_threads.py` rather than only the PR description was the right call.

---

